PR #29, opened 2014-11-07
Reproduce problem with decrypting properties from @PropertySource

szpak:
Decryption done in EnvironmentDecryptApplicationListener is performed before properties
added with `@PropertSource` are added and therefore they are not decrypted. By the way
a few other tests were added.


spencergibb:
Closing as this is very old and the code this references is now in spring-cloud-commons

Issue #74, opened 2015-01-28
logging.level. properties not being picked up when coming from remote config repo


spencergibb:
They show up in `/env`, but the logging system must be already initialized by then.

spencergibb:
see http://stackoverflow.com/questions/28198647/managing-logging-level-using-configserver

PR #75, opened 2015-01-28
listen for ContextRefreshedEvents to rebind logging

spencergibb:
fixes gh-74

@dsyer is `ContextRefreshedEvent` to broad to listen for here?  Should we publish an internal event (ie: `RemoteRepoRefreshedEvent`) when we first load the environment.


dsyer:
I'm not sure what this change accomplishes. Isn't the config loaded on startup anyway, so rebinding on `ContextRefreshedEvent` is basically redundant?

spencergibb:
@dsyer I'm seeing config is loaded but specifically `logging.level` properties don't seem to work.  Like the logging system is already initialized.

PR #77, opened 2015-02-03
Subversion support for spring-cloud-config-server

eidottermihi:
Many companies are still using Subversion as their main SCM tool, so it would be nice if Spring Cloud Config Server would support it too.

I tried to implement this feature based on an implementation of `EnvironmentRepository` which is backed by SVNKit. Because many parts are very similiar to `JGitEnvironmentRepository` I pulled them up to a new abstract superclass called `AbstractSCMEnvironmentRepository`.


dsyer:
Interesting. Did you complete the contributor's agreement (link in README)?

eidottermihi:
@dsyer Whoops, forgot about that, but completed it right now.

pom.xml
```diff
@@ -48,6 +48,11 @@
 			<artifactId>org.eclipse.jgit</artifactId>
 			<version>3.5.3.201412180710-r</version>
 		</dependency>
+        <dependency>
+            <groupId>org.tmatesoft.svnkit</groupId>
+            <artifactId>svnkit</artifactId>
+            <version>1.8.7</version>
+        </dependency>
```
spencergibb: probably need `<optional>true</optional>` here now that we have multiple impls.  Probably on jgit as well.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ConfigServerConfiguration.java
```diff
@@ -42,7 +42,6 @@ public SpringApplicationEnvironmentRepository EnvironmentRepository() {
 	}
 
 	@Configuration
-	@Profile("!native")
```
spencergibb: Should probably keep a `@Profile("git")` here.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/JGitEnvironmentRepository.java
```diff
@@ -184,36 +103,11 @@ private synchronized Environment loadEnvironment(Git git, String application,
...
-		return locations.toArray(new String[0]);
-	}
-
 	private Ref checkout(Git git, String label) throws GitAPIException {
 		CheckoutCommand checkout = git.checkout();
 		if (shouldTrack(git, label)) {
 			trackBranch(git, checkout, label);
-		}
-		else {
+		} else {
```
spencergibb: Try and keep our [code conventions](https://github.com/spring-cloud/spring-cloud-netflix/#code-conventions-and-housekeeping)

pom.xml
dsyer: I'm OK with keeping Jgit non-optional (anyone who uses svn has to jump through an extra hoop, but got works out of the box).

spencergibb: sure, np

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ConfigServerConfiguration.java
eidottermihi: I wasnt sure about that. My understanding is that if we keep a "git" Profile it must be activated explicitly. 
This way the default is Git, but can be overriden (to "native" or "subversion").

spencergibb: That matches up with @dsyer's comment about optional on jgit.

spencergibb:
@eidottermihi can you rebase against master?  Changes were made to `JGitEnvironmentRepository` that I'm not sure how to resolve.

spencergibb:
@eidottermihi I'm getting failed tests running the builds

```
Failed tests:
  SubversionConfigServerIntegrationTests.contextLoads:39 null

Tests in error:
  SVNKitEnvironmentRepositoryIntegrationTests.update:89 » BeanCreation Error cre...
  SVNKitEnvironmentRepositoryIntegrationTests.vanilla:76 » BeanCreation Error cr...
```

spencergibb:
Odd that the tests fail for me, but not on the build server

spencergibb:
The tests are loading git even though `@ActiveProfiles("subversion")` is set.

Issue #58, opened 2015-01-12
Supporting multiple git repositories?


iceycake:
Instead of a single git repository, does Spring Cloud Config Server plan to support more than 1 git repository?  If running Spring Cloud Config Server as a central config management server, it may have ownership concerns on having hundreds or thousands different application configurations in a single git repo.

warmuuh:
also the whole branch-stuff becomes a bit dirty because if you have e.g. a branch for app1, version X and a branch for app2, version Y, the Y-branch should not contain any artefacts from app1, but master then contains all artefacts? also the branch-naming should then be explicit (i.e. with prefix) to differentiate.

brianlaframboise:
I second this. Our current configuration solution involves one git repository per application so that a) each application's configuration is isolated from all others and b) each git repo can have its own set of permissions (administered in GitHub). It would be great if Spring Cloud Config Server could support a similar setup.

@warmuuh As for the isolation, a possible workaround would be to create the initial config repo with only a basic commit (ex: a README). Then create a branch per-application and commit application-specific config files in the relevant branch.

As an upside, this would isolate application files from each other, preventing an accidental "search, replace all, commit, push" workflow from affecting multiple apps.

However, creating speculative configuration would require creating a branch-from-a-branch (which may get confusing or be error prone for some users). Also, this doesn't solve the git credential problem of all users being able to push changes to any app (even if the configs are isolated in various branches).

dsyer:
I don't think it would be terribly difficult to implement a git-repository-per-application version of `EnvironmentRepository`. If anyone wants to contribute some code, please feel free (see the guidelines on contributing in the readme).

iceycake:
I will give it a shot since this is a deal breaker for the project.

brianlaframboise:
@dsyer it might be a little more complicated than that. Currently the `encrypt.key` value is global to the Spring Cloud Config Server. Supposing @iceycake implements such an `EnvironmentRepository`, the Config Server would still allow any client to decrypt any other project's encrypted values.

I know this behaviour is the current stated design of the Config Server. However, assuming we want the Config Server to aggregate multiple git repositories so that they can be independently managed and audited, it seems fitting to extend that capability to the encryption of the properties inside that git repository.

The solution here might be (as the documentation alludes) to embed the Config Server into a Spring Boot app with Spring Security enabled with custom/advanced authorization rules such that various clients can only access the configurations to which they are entitled.

[...]

PR #82, opened 2015-02-05
Issue#58

iceycake:
This implementation adds multiple git repositories support to Spring Cloud Config Server.  By given the following properties file:

<pre>
info:
  component: Config Server
spring:
  application:
    name: configserver
  jmx:
    default_domain: cloud.config.server
  cloud:
    config:
      server:
        git:
          uri: https://github.com/spring-cloud-samples/config-repo
          repos:
            - patterns: iceycake-*
              uri: https://github.com/iceycake/config-repo
            - patterns: project1-*,*-project1
              uri: https://github.com/spring-cloud-samples/config-repo-1
</pre>

with uri:

<pre>
  /{application}/{profile}/{label}
</pre>

where

* spring.cloud.config.server.git.uri is the default git repository
* spring.cloud.config.server.repos is a mapping between a git repository and a string pattern that matches the {application} string in the endpoint.

For details, please refer to https://github.com/spring-cloud/spring-cloud-config/issues/58


dsyer:
Nice, thanks. Could you add some tests, maybe? And documentation source code is here https://github.com/spring-cloud/spring-cloud-config/tree/master/docs/src/main/asciidoc if you want to add your example from above.

Also, one small suggestion for an improvement: you can use `PatternMatchUtils` to match one from an array of patterns, so it would be useful to make the `name` multi-valued. Maybe we could call it "pattern" as well to make it more obvious?

iceycake:
I have added unit testing for the implementation.  However, I have ran into issue when I try to implement the integration test.  This may not be the right forum to ask and I should open an issue in the Spring Boot project:

<pre>
		context = new SpringApplicationBuilder(TestConfiguration.class).web(false)
				.properties("spring.cloud.config.server.git.uri:" + defaultRepoUri)
				.properties("spring.cloud.config.server.git.repos[0].patterns:" + "*test1*")
				.properties("spring.cloud.config.server.git.repos[0].uri:" + test1RepoUri).run();
		EnvironmentRepository repository = context.getBean(EnvironmentRepository.class);
</pre>

I'm unable to inject the 2 new properties "patterns" and "uri" into MultipleJGitEnvironmentRepository, which is

<pre>
List&lt;Map&lt;String, Object&gt;&gt; repos
</pre>

iceycake:
Nevermind.  I found workaround by using

<pre>properties(Map<String, Object>)</pre>

PR #97, opened 2015-03-01
Suggestion: add a property to Config Server to allow Git to read sub directories

michelzanini:
Hi,

I would like to separate my Git config repo into several sub directories, one for each app so all the files are not mixed together in the root folder.

I know there is a property I can use: *spring.cloud.config.server.git.searchPaths* where I can provide a list of directories to add when loading the files.

However, I would like this to be more automatic instead of having to add another folder to *spring.cloud.config.server.git.searchPaths* every time I have a new folder for an app configuration added.

I created a sample pull request just to show the idea with one-level sub directories.
However, I think we can make this even better by recursively going into all sub directories even after first level.

Anyway, just a suggestion, since I think this can be very useful to reduce and organise configuration.

PS: I am not suggesting to merge this pull request, this is just to present the idea.

Thanks.


dsyer:
I prefer to just allow the search paths to be patterns. See ee4717b.

Issue #112, opened 2015-03-20
Enhancement [spring-cloud-config-server]: Add endpoint for .json similar to .properties, .yml, and .yaml endpoints


kevincmanning:
.yml, .yaml, and .properties require 3rd party libraries to parse whereas a json can be parsed natively by JavaScript interfacing with the config server api.

The outputs coming from {name}/{profiles} is verbose (great for debugging) but does not resolve the property that is to be used. If you want to use JSON, the logic of selecting the correct property is delegated to the client.

Suggestion: Add an endpoint for {name}-{profiles}.json and for /{label}/{name}-{profiles}.json

I have this code already implemented on forked repo and committed (with tests added and small refactor). If you think this is a good idea, may I submit a pull request?

spencergibb:
PRs are always welcome.

kevincmanning:
Cool. :+1:

Here is the pull request: https://github.com/spring-cloud/spring-cloud-config/pull/113

kevincmanning:
Sorry if there was a way I could have attached that pull request to this issue. My github experience is limited.

spencergibb:
by mentioning the issue in the PR they get linked

PR #113, opened 2015-03-20
Enhancement [spring-cloud-config-server]: Add endpoint for .json similar to .properties, .yml, and .yaml endpoints

kevincmanning:
.yml, .yaml, and .properties require 3rd party libraries to parse whereas a json can be parsed natively by JavaScript interfacing with the config server api.

The outputs coming from {name}/{profiles} is verbose (great for debugging) but does not resolve the property that is to be used. If you want to use JSON, the logic of selecting the correct property is delegated to the client.

Suggestion: Add an endpoint for {name}-{profiles}.json and for /{label}/{name}-{profiles}.json

Mentioned in Issue #112

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
@@ -1,11 +1,7 @@
 package org.springframework.cloud.config.server;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
 
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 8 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
```diff
@@ -72,34 +68,42 @@ public Environment labelled(@PathVariable String name, @PathVariable String prof
...
-	private String sortLines(Map<String, Object> properties) throws IOException {
-		List<String> list = new ArrayList<String>();
-		for (Entry<String, Object> entry : properties.entrySet()) {
-			if (entry.getKey().equals("spring.profiles")) {
-				continue;
-			}
-			String line = entry.getKey() + ": " + entry.getValue();
-			list.add(line);
-		}
-		Collections.sort(list);
```
spencergibb: Why remove the sort?

kevincmanning: I changed from LinkedHashMap to TreeMap in convertToProperties method.
TreeMap implements SortedMap

This way, the map is sorted by key upon insertion and the order propagates to both the json and the other endpoints.

spencergibb: Gotcha

dsyer:
The method you removed explicitly removes `spring.profiles` though as well. Can we have that back?

kevincmanning:
That logic is still in place. Method was renamed to getPropertiesString since sortLines no longer seemed appropriate.

However, this did bring to my attention that spring.profiles is present in the .json endpoint since this method is not utilized. I will move this logic to the method generating the properties Map.

After adding another commit, do I need to close and open another pull request in order to get it included?

dsyer:
> do I need to close and open another pull request in order to get it included?

No, just push to the same branch as you did before (and squash the commits if you are feeling helpful).

kevincmanning:
Okay. I have completed this task, ran tests, squashed commits (I think, first time using rebase here as my company only uses merge in their development cycle so for sure double check me on this), and pushed to master branch on my forked repo.

Thanks,
Chris Manning

dsyer:
Thanks. Final task for you: did you complete the contributor's agreement (link in README)?

kevincmanning:
I have now :+1:

PR #114, opened 2015-03-24
Use 'trunk' as default label for Subversion repositories

### LINKED ISSUE #107: make default label 'trunk' for svn on server

royclarkson:
This commit adds a new AbstractEnvironmentRepository that includes
a getDefaultLabel() method for all EnvironmentRepository
implementations. The default implementation returns 'master'.
SvnKitEnvironmentRepository overrides this method to return 'trunk'.
ConfigServerMvcConfiguration and ConfigServerBootstrapConfiguration
will utilize labels configured on the client and server before falling
back to the repository's default label.

Resolves #107

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ConfigServerBootstrapConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ConfigServerBootstrapConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,9 +29,9 @@ import org.springframework.util.StringUtils;
  * enabled with <code>spring.cloud.config.server.bootstrap=true</code>. This would be
  * useful, for example, if the config server were embedded in another app that wanted to
  * be configured from the same repository as all the other clients.
- * 
- * @author Dave Syer
  *
+ * @author Dave Syer
+ * @author Roy Clarkson
  */
 @Configuration
 public class ConfigServerBootstrapConfiguration {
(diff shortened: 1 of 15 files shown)


dsyer:
I'm not sure I understand why `AbstractEnvironmentRepository` (or any `EnvironmentRepository` implementation) needs a default label. Isn't the label always passed into the `findOne()` method explicitly? I'd expect the caller to always be responsible for providing the label.

royclarkson:
Considering the labels, "master" and "trunk" are fairly specific to Git and Subversion respectively, it seems reasonable OOD to make those properties of the `EnvironmentRepository` implementations. I considered another `findOne()` method that did not accept a label (to correspond with the same method in the controller). Then it could use the default label from the repository internally. In the end it felt a bit like clutter to add another method though, but maybe that would be a good idea? An alternative route could be to set this in `ConfigServerConfiguration` when the Profile specific settings/beans are applied.

dsyer:
I see. I just hate using base classes for contracts I guess. Maybe we should add the method to the `EnvironmentRepository` interface? Or else the externalization concerns should go in `ConfigServerProperties`. I think I prefer the latter, but without seeing it I'm not sure.

spencergibb:
Deleting `AbstractEnvironmentRepository` and moving `getDefaultLabel()` to `EnvironmentRepository` seems like a good way to go.

royclarkson:
I did add `getDefaultLabel()` to `EnvironmentRepository` since the interface is what is passed around everywhere. So in terms of a contract, it's not reliant on a base class. I didn't really want to add the base class, but the reason I did was to have a single place to define the default label for all repository implementations. Removing `AbstractEnvironmentRepository` would obviously mean adding `getDefaultLabel()` into each repository class. Otherwise, I'm happy to remove it.

dsyer:
Let's remove it then. I'm sure there is no actual "base class" of common shared defaults since the default label is very backend specific.

royclarkson:
ok, will do. thanks.

royclarkson:
Changes have been pushed. I removed `AbstractEnvironmentRepository` as discussed and updated the affected repository implementations.

spencergibb:
:+1:

PR #121, opened 2015-04-02
Initial pr for introducing encryption key per environment

wojtkiewicz:
We are trying to extend current encryption functionality by adding feature to use different encryption keys per environment. However we think that scope of adding such a feature seems to be too large for a single PR. In this PR we introduced no functional changes but some abstractions that will help us to extend current implementation in the future. Please let us know what do you think and if we are moving in the right direction.

Notable changes:
- TextEncryptorLocator - this class will be responsible for locating appropriate TextEncryptor for given application name and profile, in this version implementation is rather simplistic and mimics old behaviour found in EncryptionController
- EnvironmentEncryptor - encryptor that operates on Environment abstraction, contains extracted logic from EncryptionController.decrypt(Environment) function

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -152,7 +152,7 @@ spring:
 
 If you don't use HTTPS and user credentials, SSH should also work out
 of the box when you store keys in the default directories (`~/.ssh`)
-and the uri is points to an SSH location,
+and the uri points to an SSH location,
 e.g. "git@github.com:configuration/cloud-configuration". The
 repository is accessed using JGit, so any documentation you find on
 that should be applicable.


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
```diff
@@ -15,22 +15,10 @@
  */
 package org.springframework.cloud.config.server;
 
-import java.io.IOException;
```
spencergibb: Please follow our conventions

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/TextEncryptorLocator.java
```diff
@@ -0,0 +1,25 @@
...
+    public TextEncryptorLocator(TextEncryptor encryptor) {
+        this.encryptor = encryptor;
+    }
+
+    // temporary solution to support EncryptionController REST API
+    public void setEncryptor(TextEncryptor encryptor) {
+        this.encryptor = encryptor;
+    }
+
+    public TextEncryptor get() {
```
spencergibb: Not sure I see the benefit of this abstraction yet.  Can you explain a bit what you're thinking here?

rzukow: In the future, TextEncryptorLocator could be used to find an ecryptor dedicated for given application profile (hopefully with some new finders methods). 
In this PR it is not very usefeul yet, but at least it extracts a state (TextEncryptor) from other controllers. 
Currently TextEncryptor is injected and at the same time there is a method to update key. 
This scenario is improved by the abstraction, as updating a key will update a property of singleton bean instead of recreating a property in a EncryptionController. You don't need to depend on EncryptionController if you want to have access to encryptor, just inject the TextEncryptorLocator.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
rzukow: fixed

PR #124, opened 2015-04-10
Introducing TextEncryptorLocator abstraction

wojtkiewicz:
We are trying to extend current encryption functionality by adding feature to use different TextEncryptor per environment. However we think that scope of adding such a feature seems to be too large for a single PR. In this PR we introduced no functional changes but some abstractions that will help us to extend current implementation in the future. Please let us know what do you think and if we are moving in the right direction.

TextEncryptorLocator can now be overridden with custom implementation which enables providing different encryption strategies per application profile.

Notable changes:

- TextEncryptorLocator - this class will be responsible for locating appropriate TextEncryptor for given application name and profile, in this version implementation is rather simplistic and mimics old behaviour found in EncryptionController
- EnvironmentEncryptor - encryptor that operates on Environment abstraction, contains extracted logic from EncryptionController.decrypt(Environment) function

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EncryptionController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EncryptionController.java
@@ -41,6 +41,7 @@ import org.springframework.security.rsa.crypto.KeyStoreKeyFactory;
 import org.springframework.security.rsa.crypto.RsaKeyHolder;
 import org.springframework.security.rsa.crypto.RsaSecretEncryptor;
 import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -62,13 +63,17 @@ public class EncryptionController {
 
 	private final TextEncryptorLocator encryptorLocator;
 
-	public EncryptionController(TextEncryptorLocator encryptorLocator) {
+	private final ConfigServerProperties properties;
+
+	public EncryptionController(TextEncryptorLocator encryptorLocator,
+								ConfigServerProperties configServerProperties) {
 		this.encryptorLocator = encryptorLocator;
+		this.properties = configServerProperties;
 	}
 
 	@RequestMapping(value = "/key", method = RequestMethod.GET)
 	public String getPublicKey() {
-		TextEncryptor encryptor = encryptorLocator.locate();
+		TextEncryptor encryptor = locateDefaultTextEncryptor();
 		if (!(encryptor instanceof RsaKeyHolder)) {
 			throw new KeyNotAvailableException();
 		}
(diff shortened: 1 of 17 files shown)


wojtkiewicz:
This is a followup to https://github.com/spring-cloud/spring-cloud-config/pull/121 with everything squashed together. We introduced few changes discussed on gitter and tried to follow Spring code conventions this time.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EncryptionController.java
```diff
@@ -63,15 +61,16 @@
 
 	private static Log logger = LogFactory.getLog(EncryptionController.class);
 
-	private TextEncryptor encryptor;
+	private final TextEncryptorLocator encryptorLocator;
 
-	@Autowired(required = false)
-	public void setEncryptor(TextEncryptor encryptor) {
-		this.encryptor = encryptor;
+	@Autowired
```
spencergibb: Unneeded here because this constructor is used in an `@Configuration` class.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EncryptionController.java
```diff
@@ -90,8 +89,9 @@ public String getPublicKey() {
 			ByteArrayResource resource = new ByteArrayResource(file.getBytes());
 			KeyPair keyPair = new KeyStoreKeyFactory(resource, password.toCharArray())
 					.getKeyPair(alias);
-			encryptor = new RsaSecretEncryptor(keyPair);
-			body.put("publicKey", ((RsaKeyHolder) encryptor).getPublicKey());
+			RsaSecretEncryptor encryptor = new RsaSecretEncryptor(keyPair);
+            updateEncryptor(encryptor);
```
spencergibb: tabs here

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EncryptionController.java
```diff
@@ -102,23 +102,33 @@ public String getPublicKey() {
...
 		}
 		logger.info("Key changed with literal value");
 		return new ResponseEntity<Map<String, Object>>(body, HttpStatus.CREATED);
-
 	}
 
+    // this is temporary solution to support existing REST API
+    // downcasting is only necessary until we introduce some key management abstraction,
+    // we don't want TextEncryptorLocator to have setEncryptor method
+    private void updateEncryptor(TextEncryptor encryptor) {
```
spencergibb: Maybe move this to `EncryptorFactory`?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
```diff
@@ -23,7 +23,6 @@
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeMap;
-
```
spencergibb: Formatting

spencergibb:
I general we use tabs instead of spaces.  We have our [eclipse-code-formatter.xml](https://github.com/spring-cloud/spring-cloud-build/blob/master/spring-cloud-build/eclipse-code-formatter.xml) checked in that you can use in eclipse or intellij.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EncryptionController.java
wojtkiewicz: Could you elaborate on this? EncryptorFactory has been moved recently to spring-cloud-context module and does not have any state. Instead of changing the state on SingleTextEncryptorLocator I would rather add new key to some key storage or something similar but this seems to be out of scope for this PR.

spencergibb: Ah, good point about it being in not only a different module, but a different project all together.  I guess the note will have to do.

wojtkiewicz:
@spencergibb thanks for the formatter, switched from spaces to tabs :-)

spencergibb:
I think it looks ok.  Any objections @dsyer?

dsyer:
There are still some things I would like to tidy up.

* Why is the default application name "default" (that's a change from "application" which is the current de facto default)?

* Why is a `TextEncryptorLocator` `@Autowired` into the `ConfigServerEncryptionConfiguration`? I guess it makes sense if the `@Bean` that is there would be `@ConditionalOnMissingBean`? Maybe that's what you meant to do?

* The `environmentController()` bean in `ConfigServerMvcConfiguration` is autowired in 2 different styles (by field in the enclosing class, and by method parameter). Pick one and stick to it?

wojtkiewicz:
Applied your suggestions :-)

dsyer:
I get load of conflicts when I try to merge this now. Mostly looks like whitespace. Maybe you could rebase again and push --force?

dsyer:
Ignore that last comment

dsyer:
I merged this after a few tweaks. It would be good if you could send another PR with some tests for the new features in `EncryptionController`.

PR #126, opened 2015-04-17
Fix typos in documentation

shakuzen:
Going through the documentation led me to find these typos. Here is a pull request to fix them.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -36,7 +36,7 @@ Server? The strategy that governs this behaviour is the
 * `{label}` which is a server side feature labelling a "versioned" set of config files.
 
 Repository implementations generally behave just like a Spring Boot
-application loading configuration files from a "spring.config.name"
+application loading configuration files from a "spring.application.name"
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -36,7 +36,7 @@ Server? The strategy that governs this behaviour is the
 * `{label}` which is a server side feature labelling a "versioned" set of config files.
 
 Repository implementations generally behave just like a Spring Boot
-application loading configuration files from a "spring.config.name"
+application loading configuration files from a "spring.application.name"
```
spencergibb: I think this is actually `spring.config.name` (see [here](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/NativeEnvironmentRepository.java#L178)).  It means find `<appname>.{yml|properties}`, rather than `application.{yml|properties} (which applies to all apps).

dsyer:
Spencer is correct, this is not a typo, if I understand correctly. If it's still confusing and you can think of a way to re-word it, please do.

PR #135, opened 2015-04-28
Fix error loading config files in config server on Windows

abrochard63:
Make sure that the "normal" file path does not have Windows separators ("\") to be coherent with the "pattern" file path


dsyer:
Did you fill out the contributor's agreement (link in README)? One of you or @marmanan needs to do it to get this fixed.

abrochard63:
Thanks for the reminder !
I have now  filled the contributor's agreement .

PR #151, opened 2015-05-16
Removed duplicate groupId and verson

jibijose:


dsyer:
Thanks. Did you fill out the contributors agreement (link in README)?

dsyer:
The JCE links were added elsewhere (I think you submitted another PR). The duplicate maven co-ordinates are intentional I believe (in case the parent changes at some point). Thanks.

Issue #155, opened 2015-05-19
Send 404 if application not found


hmichopoulos:
Hi,
I would like to configure Config Client to accept only non empty configuration sources.

Currently, if the configuration of my `FooApplication` has not reached the Config Server yet, the Config Client will receive empty configuration sources and will go on. 
I suggest it to be configurable. If `acceptEmpty` is false, it would fail and probably retry.

Thanks

hmichopoulos:
Actually, Config Server should send 404 if the application is not found

dsyer:
This is marked as fixed in #156

Sangdol:
Is this actually fixed? #156 has been closed without being merged and `acceptEmpty` option doesn't work. I cannot find the `acceptEmpty` in the code as well.

spencergibb:
No, it was closed since the PR didn't actually issue a 404

spencergibb:
@hmichopoulos never submitted another PR

PR #156, opened 2015-05-19
Add acceptEmpty option to accept or reject empty PropertySources

hmichopoulos:
Add acceptEmpty configuration option. If set to false and empty
PropertySources are returned, the client throws an
IllegalStateException.

Fixes gh-155


hmichopoulos:
Should return 404. Will submit another PR.

PR #161, opened 2015-05-28
Additional tests for new endpoints in EncryptionController

wojtkiewicz:
Added few missing tests for new endpoints in EncryptionController.


spencergibb:
@wojtkiewicz is this still valid? There is already a `EncryptionControllerMultiTextEncryptorTests` that you authored.

wojtkiewicz:
It's been a while and I don't fully remember what happened but these test cases seem to be identical...

dsyer:
I think I made these changes already

PR #162, opened 2015-05-28
Optionally decrypt secret properties before sending them to the clients

marmanan:
Add the option to disable decryption in the EnvironmentController to let clients manage decryption of the ciphered properties. In this way, secret properties go encrypted from server to client.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/ConfigServerEncryptionConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/ConfigServerEncryptionConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.config.server.encryption;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


dsyer:
This looks like it might be useful to some people. But I think the code in `EnvironmentController` has changed since you proposed the change. You might need to revise it and add a few null checks.

marmanan:
I only see the encryptor is used once in EnvironmentController and it´s null checked

Line 102
		if (environmentEncryptor != null) {
			environment = environmentEncryptor.decrypt(environment);
		}

dsyer:
Correct. OK thanks. I think I was looking at the implementation of the `EnvironmentEncryptor` itself and seeing a lot of places where a null check might be needed.

marmanan:
Hi Dave, are you planning to add this to 1.0.2?

dsyer:
I think something equivalent will be in master soon. I'm re-designing some of the encryption features.

dsyer:
I couldn't use your code because your repo doesn't exist any more (according to Github). The change is in though (with a slightly different property name).

Issue #180, opened 2015-06-30
ConcurrentModificationException when properties stored in Map after "spring.profiles"


kausnz:
In org.springframework.cloud.config.server.EnvironmentController#postProcessProperties

The code throws a ConcurrentModificationException if there's a property in the map that's stored after "spring.profiles" as the following removal doesn't update the iterator used by the foreach. Use propertiesMap.keySet().iterator().remove(..) instead.

	private void postProcessProperties(Map<String, Object> propertiesMap) {
		for (String key : propertiesMap.keySet()) {
			if (key.equals("spring.profiles")) {
				propertiesMap.remove(key);
			}
		}
	}

kausnz:
Fix:
	private void postProcessProperties(Map<String, Object> propertiesMap) {
		Iterator<String> i = propertiesMap.keySet().iterator();
		while(i.hasNext()) {
			if (i.next().equals("spring.profiles")) {
				i.remove();
			}
		}
	}

PR #185, opened 2015-07-03
Fixed ConcurrentModificationException when properties stored in Map a…

kausnz:
This is the fix for the issue I reported under:
ConcurrentModificationException when properties stored in Map after "spring.profiles" #180

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
@@ -16,15 +16,8 @@
 package org.springframework.cloud.config.server;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnvironmentController.java
```diff
@@ -16,15 +16,8 @@
...
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
spencergibb: Our convention is to not use '*' imports, please update.

dsyer:
Fixed in 7828a0.

Issue #191, opened 2015-07-08
NPE in ConfigClientAutoConfiguration when bootstrapping with a custom parent CTX


KeesKoffeman:
i.e. BeanFactoryUtils.beanNamesForTypeIncludingAncestors(..) can return a different result than BeanFactoryUtils.beanOfType(..), therefore BeanFactoryUtils.beanOfTypeIncludingAncestors(..) should be used.

KeesKoffeman:
Probably duplicate of gh-179

PR #192, opened 2015-07-08
Fixes gh-191

KeesKoffeman:


spencergibb:
Do you have a test that recreates the problem so we can verify your fix?  Also, have you signed [the contributor agreement](https://support.springsource.com/spring_committer_signup)?

KeesKoffeman:
I did not write a test as I quickly had to patch our version in order to continue the development of our project. The mismatch in API usage looked pretty obvious to me so I adjusted it for myself but didn't want do that without any notice to the project.

I do agree however that the scenerio should be covered by a unit test. In the current test only the presence of the bean in the context (including ancestors) is tested and not the actual retrieving of the bean. Having said that I need to find the time to do that my self, so it might take a while before I get to it.

Agreement is signed.

dsyer:
Merged in ccc44db. Please add a test if it makes sense later, and please put something in the git log to describe what you did and why (e.g. see the log I added for you). Thanks.

Issue #193, opened 2015-07-17
ConfigServicePropertySourceLocator should log a warning and not an error if failFast is false


cppwfs:
Currently if the config server is not available and the failFast flag is set to false Cloud Config Client will report an error, but should be a warning.  

The scenario is when a Spring-Cloud-Stream module starts up (specifically the samples) and no config server is running the first log message is an Error.  This makes it look like there is a problem when none exists. 

I will submit a PR for this for your review.

dsyer:
Fixed in 54f76bc

PR #194, opened 2015-07-17
GH-193 logs warning if failFast is false

cppwfs:
ConfigServicePropertySourceLocator should log a warning and not an error if failFast is false


dsyer:
Merged in 54f76bc2

PR #212, opened 2015-08-12
Client doesn't throw 404 exception unless all labels are checked

### LINKED ISSUE #210: Config client doesn't try next label if first one doesn't exist
If I set `spring.cloud.config.label=nonexistant,master` then the RestClient receives a 404 on the 'nonexistant' label which is converted to an HttpClientResponseException. The code is checking for a 200 but should also catch the client exception.

See:
https://github.com/spring-cloud/spring-cloud-config/blob/5f05eadbd4c0352515ede0d08427e14bd297d737/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java#L120

atoregozh:
fixes #210


spencergibb:
something's weird with your branch.  Maybe rebase against master? Where is all the history? https://github.com/atoregozh/spring-cloud-config

atoregozh:
@spencergibb my branch is here: https://github.com/atoregozh/spring-cloud-config-1

spencergibb:
my mistake

spencergibb:
Squashed and merged via fb706e327c6c1c26839261017323c1bc762fbf92

PR #218, opened 2015-08-31
Omit surrounding quotes for encrypted values in .properties file

tallstak:
Just adding to the documentation.


dsyer:
This change appears to already be in master. Thanks.

Issue #244, opened 2015-10-01
Add version to environment response


spencergibb:

shakuzen:
@spencergibb Is there going to be any update to the documentation for this or will that come when/if this is used by the client?

spencergibb:
@shakuzen yes, currently the version is only logged by the client.

PR #245, opened 2015-10-01
Add version to environment response

spencergibb:
fixes gh-244

@dsyer probably my biggest question is how to present the version to boot? Currently I'm adding the version to to the name of the property source on the server. The client isn't using version yet.  What do you think?

```
{
    "configService:https://github.com/spring-cloud-samples/config-repo/application.yml@bdcfc52819cfb29582ebb06f7b970ae6b65e4985": {
        "info.description": "Spring Cloud Samples"
    }
}
```

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -80,6 +80,11 @@ public class ConfigServicePropertySourceLocator implements PropertySourceLocator
 			for (String label : labels) {
 				Environment result = getRemoteEnvironment(restTemplate, client.getRawUri(), client.getName(), client.getProfile(), label.trim());
 				if (result != null) {
+					logger.info(String.format("Located environment: name=%s, profiles=%s, label=%s, version=%s",
+							result.getName(),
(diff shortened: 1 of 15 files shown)


dsyer:
I don't like the commit id in the property source name (it might change between fetches). Wouldn't it be better to add it as a field in `Environment`? (Not sure what the client would do with it other than log it.)

spencergibb:
It is in a field in environment. The request was for someone wanting to see what version was retrieved.

spencergibb:
A kind of auditing function.

spencergibb:
If the `EnvironmentEndpoint` could display some metadata about the propertysource besides name, we could put the version there.

spencergibb:
How about this?
```
2015-10-01 15:21:08.416  INFO 3322 --- [           main] c.c.c.ConfigServicePropertySourceLocator : Fetching config from server at: http://localhost:8888
2015-10-01 15:21:09.087  INFO 3322 --- [           main] c.c.c.ConfigServicePropertySourceLocator : Located environment: name=customers, profiles=[default], label=null, version=bdcfc52819cfb29582ebb06f7b970ae6b65e4985
2015-10-01 15:21:09.087  INFO 3322 --- [           main] b.c.PropertySourceBootstrapConfiguration : Located property source: CompositePropertySource [name='configService', propertySources=[MapPropertySource [name='https://github.com/spring-cloud-samples/config-repo/application.yml']]]
2```

Issue #106, opened 2015-03-17
Config server health endpoint always basing itself on 'default' spring profile


andreasevers:
The configserver appears with status “DOWN” for me because:
`org.springframework.cloud.config.client.ConfigServerHealthIndicator`
does a locate of the propertySources to verify if the configServer can access its own properties through the REST endpoint.

This locator is basing itself on an environment object, constructed as such:
```
this.env = new AbstractEnvironment() {};
```
Of course this takes only the default profile, so the loc properties are used.

This results in a REST call being made to:
`http://localhost:8888/application/native,dev/master`
Which of course cannot be reached from dev (which is running on a Jboss).

And because of this, the configServer appears to be “DOWN”, while it in fact is functioning perfectly.

The `env` endpoint returns (among other things) this:
```json
profiles": [
    "dev",
    "native"
  ]
```

There doesn't seem to be a way to tell the config server to use the currently active profile.

dsyer:
I'm not quite following the bit where you say "of course". Is the config server running on port 8888 or not? Is the health indicator in a client application ("application") or in the server itself?

andreasevers:
Oh yes this is probably confusing. The loc profile loads the following `bootstrap.yml`:
```yml
server:
  port: 8888
  servletPath: /*
```
This is deployed on the embedded tomcat of Spring boot.

For `dev` we managed to deploy everything on jboss, using the following `bootstrap-dev.yml`:
```yml
server:
  port: 22121
  contextPath: /configuration
```

The health indicator is in the Spring cloud config server, not another microservice being the client of the server.

In the `application.yml` I've only put things that are allowed to be shared across all microservices, as the property-file resolution also takes `application-{profile}.yml` files of the config server:
```yml
# This file is shared by all microservices
server:
  servletPath: /*
```

dsyer:
If the app is the config server, then you don't really want it to be a config client by default (we were discussing at #104 that the default for the server should be `spring.cloud.config.enabled=false`), so then there wouldn't be a health indicator at all. If I were you I'd just add `spring.cloud.config.enabled=false` to your `bootstrap.yml`.

dsyer:
For the case of a client application I think that the "default" profile is fine for testing if the server is up or not. So all in all I think this is a non-issue once we resolve #104.

seno5979:
Are we sure that this issue has already been fixed.

Because currently I still got issue on client healthIndicator, requesting 'default' profile to config server which resulting below :

````
[...]

Issue #104, opened 2015-03-14
Config Server Health Status


ccit-spence:
Everything on the config server is working.  Clients work etc...   When I pull the health it is showing `DOWN`.  Do I need to do something else to make it work?

```
"configServer": {
"status": "DOWN",
"error": "no property sources located"
},
```

This is in the `application.yml`

```
endpoints:
    restart:
        enabled: true
    shutdown:
        enabled: true
    health:
        enabled: true
        sensitive: false
```

ccit-spence:
The error that I get when I access `/admin/health` The config server is running on port 8080.
```
2015-03-13 20:34:51.286 ERROR 4542 --- [nio-8080-exec-1] c.c.c.ConfigServicePropertySourceLocator : Could not locate PropertySource: I/O error on GET request for "http://localhost:8888/application/default/master":Connection refused; nested exception is java.net.ConnectException: Connection refused
```

ccit-spence:
Figured out the problem, had to disable cloud config for the server itself.  `spring.cloud.config.enabled-false`

dsyer:
Re-opening to continue the discussion. It seems like maybe `spring.cloud.config.enabled:false` should be the default for the config server. Also, some people (@ewolff https://github.com/spring-cloud-samples/eureka/issues/11) have pointed out that "DOWN" is confusing for users as a default status. Even if we decide that isn't a problem, I would like to check that "DOWN" is only the status if the config server wasn't available on startup (i.e. it's not so critical since it only affects refreshes once the app is running).

ewolff:
I tried the Eureka server with spring.cloud.config.enabled: false and can confirm that its status is UP then.

ccit-spence:
The reason I even noticed this was integrating the Spring Boot Admin project into the setup.  The problem I had was seeing this glaring "DOWN" on the list of services.  I would think for anyone implementing an admin type console for services would experience the same confusion.  

A default of `spring.cloud.config.enabled:false` makes a lot of sense to me.  Having it `true` would seem more like the fringe cases.

spencergibb:
I don't think `spring.cloud.config.enabled:false` makes much sense at all.  I think having an explicit `spring-cloud-starter-config-client` makes more sense rather than having config client in `spring-cloud-starter`.  If you don't want config, don't include it, rather than include it and turn it off.

dsyer:
We were discussing the default value on the server though. There it makes sense that the default is false.

[...]

Issue #69, opened 2015-01-23
ConfigClientProperties exposes a modified version of its "uri" property in the getter


dsyer:
Since we use `BeanUtils` to copy properties it goes blindly to the getters and setters, but the getter is modifying the URI to pull out the credentials for a RETS call. We can fix it just by renaming that method. Problem manifests itself in the /health indicator of a config client that has a username/password for the server (fails to fetch properties for health check).

PR #268, opened 2015-11-15
Fixed issue on ConfigServerHealthIndicator on client side.

seno5979:
Spring Cloud should not try to connect to default repo on startup.

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientAutoConfiguration.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientAutoConfiguration.java
@@ -16,6 +16,7 @@
 package org.springframework.cloud.config.client;
 
 import org.springframework.beans.factory.BeanFactoryUtils;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -57,10 +58,14 @@ public class ConfigClientAutoConfiguration {
 	@ConditionalOnProperty(value = "health.config.enabled", matchIfMissing = true)
 	protected static class ConfigServerHealthIndicatorConfiguration {
 
+
+		@Autowired
+		private Environment environment;
+
 		@Bean
 		public ConfigServerHealthIndicator configServerHealthIndicator(
 				ConfigServicePropertySourceLocator locator) {
-			return new ConfigServerHealthIndicator(locator);
+			return new ConfigServerHealthIndicator(this.environment, locator);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


seno5979:
This is to fixed issue mentioned below :
https://github.com/spring-cloud/spring-cloud-config/issues/104

spencergibb:
So #104 has been closed since March 2015. Also, your change specifically reverts and breaks #69.

seno5979:
But we still got issue on configServer health indicator on client side.  Seems like with the current code it will try to request 'default' application on config server which is not available:

https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerHealthIndicator.java#L25

spencergibb:
That's the `default` profile, not application. What problem are you having?

seno5979:
My application health page always show below on config-server section :

```
"configServer": {
"status": "UNKNOWN",
"error": "no property sources located"
},
```
The console log display as ;
```
ERROR org.springframework.cloud.config.client.ConfigServicePropertySourceLocator - Could not locate PropertySource: 403 Forbidden
```

After did a little investigation, seems like it requesting the 'default' application name instead of my real application name to config server.

I suspect that the code on below made the configserverhealthindicator override the existing environment and made it uable to resolve spring application name (default value was returned) :
https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerHealthIndicator.java#L25

```
new AbstractEnvironment() {
        	@Override
        	public String[] getActiveProfiles() {
        		return new String[] {"default"};
        	}
		};
```

Note :We have enabled authentication on config server side

spencergibb:
> Note :We have enabled authentication on config server side

and you're getting a 403, which makes sense.

Like I said, as is, this change will make another issue regress, so I won't merge it as is.

seno5979:
But even though I disable the security on config-server, I will still got 404, cause the healthindicator on client side requested the wrong application name.

Is this issue have a fixed already ?

seno5979:
Btw what is the purpose of having new AbstractEnvironment on:

https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerHealthIndicator.java#L25

Why we cannot autowire the already existing environment ?

spencergibb:
@seno5979 I've linked to the reason already #69.  You're issue is not #104, it appears to be something we haven't seen before.

seno5979:
What I experiencing is exactly the same as discussed on #104.

I'm not sure why it being closed while the issue still persist.
I'm fine closing my pull request, but still would like to know how the issue need to be taken care.

spencergibb:
#104 has a similar symptom. The problem was raised on the config server itself, not a client. Please open a new issue.

seno5979:
Maybe #106 is talking the spring-cloud-config client issue. But the same, the status is close while the issue still persist.

PR #272, opened 2015-11-19
Prefix keys not stripped prior to encryption

westse:
**Expected**: When `POST`ing "{name: value}textToEncrypt" to the `/encrypt` endpoint, then only "textToEncrypt" should be encrypted.
**Actual** (broken) behavior: "{name: value}textToEncrypt" is encrypted.

Note that `EncryptionController.decrypt()` masks the problem by stripping prefixes twice, once prior to decrypting and once *after* decrypting the value.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
@@ -129,8 +129,9 @@ public class EncryptionController {
 			String input = stripFormData(data, type, false);
 			Map<String, String> keys = this.helper.getEncryptorKeys(name, profiles,
 					input);
+			String textToEncrypt = this.helper.stripPrefix(input);
 			String encrypted = this.helper.addPrefix(keys,
-					this.encryptor.locate(keys).encrypt(input));
+					this.encryptor.locate(keys).encrypt(textToEncrypt));
 			logger.info("Encrypted data");
 			return encrypted;
 		}
@@ -156,7 +157,7 @@ public class EncryptionController {
 					profiles, input);
 			TextEncryptor encryptor = this.encryptor.locate(encryptorKeys);
 			String encryptedText = this.helper.stripPrefix(input);
-			String decrypted = this.helper.stripPrefix(encryptor.decrypt(encryptedText));
+			String decrypted = encryptor.decrypt(encryptedText);
 			logger.info("Decrypted cipher data");
 			return decrypted;
 		}
--- a/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/encryption/EncryptionControllerTests.java
+++ b/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/encryption/EncryptionControllerTests.java
@@ -15,13 +15,22 @@
  */
 package org.springframework.cloud.config.server.encryption;
 
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.not;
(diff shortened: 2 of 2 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
```diff
@@ -129,8 +129,9 @@ public String encrypt(@PathVariable String name, @PathVariable String profiles,
 			String input = stripFormData(data, type, false);
 			Map<String, String> keys = this.helper.getEncryptorKeys(name, profiles,
 					input);
+			String textToEncrypt = this.helper.stripPrefix(input);
```
westse: This is the proposed fix.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
```diff
@@ -156,7 +157,7 @@ public String decrypt(@PathVariable String name, @PathVariable String profiles,
 					profiles, input);
 			TextEncryptor encryptor = this.encryptor.locate(encryptorKeys);
 			String encryptedText = this.helper.stripPrefix(input);
-			String decrypted = this.helper.stripPrefix(encryptor.decrypt(encryptedText));
+			String decrypted = encryptor.decrypt(encryptedText);
```
westse: Removing extraneous `stripPrefix()` which was masking the bug in `encrypt()`. While this is likely correct long term, you may consider leaving it in for a deprecation period because some users using `EncryptionController` for round-trip encryption may have stored values with the prefix included.

spencergibb:
@dsyer thoughts?

dsyer:
I guess I can see some of the motivation here, but I don't really know for sure what is broken. Can someone explain quickly, using curl command line examples, what should work and what doesn't?

westse:
Curl on existing endpoints (`encrypt` & `decrypt`) unfortunately don't demonstrate the issue because we can't tell the result of `encrypt` is wrong (because it is encrypted), and the `decrypt` endpoint masks the problem because it strips the prefix on the decrypted value.

The best I could come up with is the unit test. It is based on my understanding that the `TextEncryptor` should never be passed a value that includes the name/value prefixes. Is that assumption correct?

westse:
Note that `org.springframework.cloud.config.server.encryption.CipherEnvironmentEncryptor` does not mask the problem (because it strips the prefix *before* decrypting only). This is in fact how I discovered it: we use the `encrypt` endpoint to encrypt values with {name:value} prefixes for inclusion in our Git repository, but then the {name:value} prefixes show up in property values when using the Spring Cloud Config client.

westse:
Semi-realistic steps to reproduce the issue. Customize for your setup.

1. `curl http://config.mycompany.com/encrypt` with body=`{key:myEncryptionKey}mySensitiveValue`.
    1. This returns a response body like `{key:myEncryptionKey}myEncryptedValue`.
1. In `myApp-myEnv.yml` (in Git repo) pointed to by your Spring Cloud Config server, enter `{key:myEncryptionKey}myEncryptedValue` as the value of a property (`myProperty`).
1. `curl http://config.mycompany.com/config/myApp/myEnv`. Find `myProperty` in the returned JSON, and notice the value is `{key:myEncryptionKey}mySensitiveValue`, but it should be just `mySensitiveValue`.

checketts:
@dsyer I just verified this problem and @westse steps.  To restate: the `/encrypt` endpoint is incorrectly embedding the `{key:value}` prefix in the encrypted contents.

PR #277, opened 2015-11-20
enable spring-cloud-config-client extension

jeffnelson:
Netflix currently has a project which enables eureka service discovery of config server by client. 

The changes here allow easier extension of spring-cloud-config-client such that any getRemoteEnvironment method could be implemented to resolve remote environment in any way desired

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/BaseConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/BaseConfigServicePropertySourceLocator.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright 2013-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.config.client;
+
+import java.util.Arrays;
+import java.util.Map;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.cloud.bootstrap.config.PropertySourceLocator;
+import org.springframework.cloud.config.environment.Environment;
+import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.core.env.CompositePropertySource;
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.http.MediaType;
+import org.springframework.retry.annotation.Retryable;
+import org.springframework.util.StringUtils;
+import org.springframework.web.client.HttpServerErrorException;
+
+/**
+ * 
+ * @author Jeffrey Nelson
+ *
+ * @param <C> type of Context class used
+ */
+public abstract class BaseConfigServicePropertySourceLocator<C extends BaseConfigServicePropertySourceLocator.Context>
+		implements PropertySourceLocator {
+
+	private static Log logger = LogFactory.getLog(BaseConfigServicePropertySourceLocator.class);
+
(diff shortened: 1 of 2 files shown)


spencergibb:
Two things: 1) lot's of unnecessary formatting changes that make it hard to see what you have really changed. 2) I've implemented two other config clients ([1](https://github.com/spring-cloud/spring-cloud-consul/tree/master/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config), [2](https://github.com/spring-cloud/spring-cloud-zookeeper/tree/master/spring-cloud-zookeeper-config/src/main/java/org/springframework/cloud/zookeeper/config)) and I haven't had to do anything like this, just plug in to the bootstrap phase.

jeffnelson:
My apologies on the formatting changes. I will to revert those when I have a chance

jeffnelson:
@spencergibb - I have reversed the formatting-only changes. It should now be much easier to see what was actually changed in ConfigServicePropertySourceLocator.

To address your second point, I would like to clarify that the purpose of this refactor is to enable someone to implement their own ConfigServicePropertySourceLocator by extending BaseConfigServicePropertySourceLocator. They would still have to then plug that bean into the bootstrap phase. What you gain with this is code-reuse. By abstracting common code, all the implementing class needs to do is fetch the remote environment in whatever way they desire.

The entire spring-cloud-config-client package can be re-used. Setting "spring.cloud.config.enabled" to false disables the out of the box "configServicePropertySource" bean and allows a different impl to be used.

Yes, of course, you can fully implement another config client from scratch, but why do so when the only differing behavior is the way in which the remote environment is retrieved?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/BaseConfigServicePropertySourceLocator.java
```diff
@@ -0,0 +1,140 @@
...
+		}
+		logger.warn("Could not locate PropertySource: "
+				+ (errorBody == null ? error == null ? "label not found" : error.getMessage() : errorBody));
+		return null;
+
+	}
+	
+	protected C newContext(org.springframework.core.env.Environment environment) {
+		try {
+			C context = contextClass.newInstance();
```
spencergibb: I think this is a bad idea.

jeffnelson: OK, another option is to make newContext abstract and require child classes to implement. 

I did it this way because I see the Context class as a simple POJO that should always have default ctor. And in some case where that won't work, the method is already protected and could be overridden.

spencergibb: I'd prefer `newContext` to be abstract.

jeffnelson: done

spencergibb:
I still struggle with the whole purpose of the PR. Having to set `spring.cloud.config.enabled=false` is the reason a bunch of code was moved out of here and into spring-cloud-context to s-c-{consul,zookeeper} wouldn't have to set that flag.

jeffnelson:
To be completely honest, I was unaware of the spring-cloud-consul project. After just a brief review, I think there are some things in it that we will utilize, but not when it comes to config. We prefer consul for its consistency, but Spring Cloud Config due to git. We still need a config client load balancing with ribbon discovering via consul. I have written such a module extending spring-cloud-config-client in a similar way in which I have described here, but currently I have to override the entire ConfigServicePropertySourceLocator.locate method. I plan to open source what I have written, but I wanted this change in order to clean it up a bit.

Does that help?

spencergibb:
We have support for looking up config server via eureka. We also want to support that lookup via consul & zookeeper as well (though I can't find the issue for consul, maybe it never was created).

jeffnelson:
As I said, I have written something that does the job for us for now. I just thought I could clean it up a bit.

spencergibb:
Setting `spring.cloud.config.enabled=false` is a non-starter for me. I want it to be based off of the work we did for eureka and config server.

spencergibb:
I've created #302 for looking up config server via discovery client (it's done in s-c-netflix but not s-c-consul or s-c-zk). I'm thinking of closing this. @jeffnelson thoughts?

dsyer:
Superseded by #302. Thanks for the interest ,and hopefully you can use that feature once it's available.

Issue #203, opened 2015-07-27
Store properties in db instead of git


K-RAD:
Spring-Cloud-Config looks like a very interesting option and we're considering using it in our project. 

However, what we don't like is the fact that the properties are stored on Git. 
If they were stored in a database it would've been way better in our view. Git is less trackable, there's less security and permissions to push/change them there, it can be down more often than db and so on. Oh, and business knows db, not git.

Will this be available at some point?

spencergibb:
@K-RAD pull requests are welcome, we aren't planning it at the moment.  There are also other implementations that don't use config server [spring-cloud-consul](https://github.com/spring-cloud/spring-cloud-consul) and [spring-cloud-zookeeper](https://github.com/spring-cloud/spring-cloud-zookeeper)  Some of the reasons we chose git are that it already implements versioning and history, which a db doesn't have by default.  Your business may not know git, but in general, most developers do.

mreyes:
Hi, Spencer

I am facing the need to implement a database backend for Spring-Cloud-Config. Up until today I have reviewed the implementation on NativeEnvironmentRepository in the spring cloud-config-server project , and looks that the key component is the ConfigFileApplicationListener so would be a valid approach to implement that listener but accessing a DB.    Any guidance or hint on how to proceed or even if this would the best approach?

spencergibb:
`ConfigFileApplicationListener` is from spring boot. I don't think that's a good place to start. You need to implement `EnvironmentRepository` from config server.

igor-dmitriev:
Hi there, is there any guide how to create a custom environment repository??? It would be awesome to have an example, for instance with Map as environment storage.

ryanjbaxter:
There is no guide, but there are plenty of examples in the source code

igor-dmitriev:
You know, they are not simple, it shouldn't be hard to do. Actually I've found this - https://github.com/spring-cloud-incubator/spring-cloud-config-server-mongodb/commit/ecedae00e22978155606cf17ca1f2c3279769aff#diff-b109c993c15a6bd4f1aceeb3716c7a85 (MongoDB EnvironmentRepository). And I read it like an example how to implement a custom implementation

spencergibb:
> it shouldn't be hard to do

[...]

PR #289, opened 2015-11-27
MongoDB based EnvironmentRepository implementation.

venilnoronha:
Hi,

I've implemented a basic MongoDB based EnvironmentRepository (Issue #203). I've also added a simple test. 

### Configuration Properties
spring.cloud.config.server.mongodb.host - _The mongodb host (optional) (default 127.0.0.1)_
spring.cloud.config.server.mongodb.port - _The mongodb port (optional) (default 27017)_
spring.cloud.config.server.mongodb.database - _The mongodb database_
spring.cloud.config.server.mongodb.username - _The mongodb username (optional)_
spring.cloud.config.server.mongodb.password - _The mongodb password (optional)_

### Enabling MongoEnvironmentRepository
The _spring.profiles.active_ must be set to _'mongodb'_ and at least the _spring.cloud.config.server.mongodb.database_ property must be configured to enable this repository.

Due to some issues I had to close my previous pull requests. Apologies.

Please review and pull.

Thanks,
Venil

### DIFF
--- a/spring-cloud-config-server/pom.xml
+++ b/spring-cloud-config-server/pom.xml
@@ -56,6 +56,10 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.data</groupId>
+			<artifactId>spring-data-mongodb</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
--- a/spring-cloud-config-server/pom.xml
+++ b/spring-cloud-config-server/pom.xml
@@ -58,6 +58,7 @@
 		<dependency>
 			<groupId>org.springframework.data</groupId>
 			<artifactId>spring-data-mongodb</artifactId>
+			<optional>true</optional>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
@@ -16,6 +16,7 @@
 package org.springframework.cloud.config.server.config;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
@@ -102,6 +103,7 @@ public class EnvironmentRepositoryConfiguration {
 	
 	@Configuration
 	@Profile("mongodb")
+	@ConditionalOnClass(MongoTemplate.class)
 	protected static class MongoRepositoryConfiguration {
 
 		@Autowired
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
@@ -20,17 +20,20 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.config.server.environment.EnvironmentRepository;
+import org.springframework.cloud.config.server.environment.MongoEnvironmentRepository;
 import org.springframework.cloud.config.server.environment.MultipleJGitEnvironmentRepository;
 import org.springframework.cloud.config.server.environment.NativeEnvironmentRepository;
+import org.springframework.cloud.config.server.environment.SearchPathLocator;
 import org.springframework.cloud.config.server.environment.SvnKitEnvironmentRepository;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
 import org.springframework.context.annotation.Profile;
 import org.springframework.core.env.ConfigurableEnvironment;
 
 /**
  * @author Dave Syer
- *
+ * @author Venil Noronha
  */
 @Configuration
 @ConditionalOnMissingBean(EnvironmentRepository.class)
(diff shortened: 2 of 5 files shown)


venilnoronha:
Travis is ignoring my MongoEnvironmentRepositoryIntegrationTests class. Its also showing the following output.

> Warning: you are leaving 2 commits behind, not connected to any of your branches:
> 
>   479c869 Merge 811a58addc1a5adf369c7fc621887742da42f0ad into 1c69c59705e4e2e7bf22d511e0b777b75f094c42
> 
>   811a58a MongoDB based EnvironmentRepository implementation. #203.
> 
> If you want to keep them by creating a new branch, this may be a good time to do so with:
> 
>  git branch new_branch_name 479c869

Any idea why?

dsyer:
Why would we need `spring.cloud.*` configuration for Mongo (it's already provided by Spring Boot)?

venilnoronha:
If at all one needs to use a different mongo instance for just storing configs.

dsyer:
Sorry, I didn't catch that. Why would you need a cloud-specific config?

venilnoronha:
I mean, if the config server is to be clubbed with a component already working on a mongo instance, then the spring.cloud.config.server.mongodb.* properties would be useful to configure a separate mongo instance for only serving configuration. (This mongo instance may as well be deployed on another server with a different IP.)

If this usecase doesn't seem to be valid then I'll remove those properties and use the spring-boot mongo instance instead.

dsyer:
I don't think it's a common enough use case that we'd want to support it out of the box with configuration properties. If you can just use the default autowired Mongo client that would help I think. Also, did you fill in the contributor's agreement (link in README)?

venilnoronha:
OK. I'll remove those properties. And yes, I've signed the ICLA. Thanks.

venilnoronha:
I've made the necessary changes to use the spring-boot provided MongoTemplate instance in MongoEnvironmentRepository. Could you please verify and pull?

spring-cloud-config-server/pom.xml
```diff
@@ -56,6 +56,10 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.data</groupId>
+			<artifactId>spring-data-mongodb</artifactId>
+		</dependency>
```
spencergibb: Should this be optional?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
```diff
@@ -95,5 +99,30 @@ public EnvironmentRepository environmentRepository() {
 			return repository;
 		}
 	}
+	
+	@Configuration
+	@Profile("mongodb")
+	protected static class MongoRepositoryConfiguration {
```
spencergibb: ConditionalOnClass(MongoTemplate.class) to go with optional dependency above.

venilnoronha:
@spencergibb done.

DCalata:
Hey Venil,

We make us of the MongoDBEnvironmentRepository implementation and unfortunately we've found out an issue when it retrieves a 'complex' copnfiguration  (see below the example) which is stored in MongoDB. This problem occurs when we want to obtain the configuration as a property invoking the REST API.

Below an example of this scenario:
 
1)  Mongo and spring cloud configuration running locally on the default ports.

	Configuration stored in mongodb.
	'''   
	   ....
	   properties : {
	        "foo" : {
	            "bar": "fooBar"
	        }
	    }
	    ...
	'''

2)  Then when invoking the following url: http://localhost:8888/master/sample-default.properties

3)   And then the result is the below

		properties: {foo={bar=fooBar}}

	 but we expected a response like the below:

		proterties : foo.bar = fooBar

The reason We expected the aboveis  because when we invoke the yaml endpoint the response looks like this:

 		foo:
   			bar: fooBar

We have been working on fixing this issue and it would be greatful If we could contribute to this project with the fix, would you mind If we do a pull request to your repository?  If so, we would like also to make some improvements in the tests implementation by using Fongo  instead of assuming you have a mongo instance running on localhost

foo.bar: fooBar

Thanks in advance,
Cheers,

venilnoronha:
Sure @DCalata, I'll have a look at it. You can send a PR to my repository as well.

venilnoronha:
@DCalata thanks for highlighting the issue. I've fixed it in 71aee7a.

javiboo:
Hi @venilnoronha,

I'm a partner of @DCalata. I've been testing last changes introduced following their comments. 

I found an issue due `cleanSource` method introduced in b7aa7cb. I make a [test](https://github.com/SerenityDevstack/spring-cloud-config/commit/6efc393439ddfd786be0a85255c254e589323260) in a forked repository that reproduces the issue.

In order to follow the behaviour of the other profiles (native, git & subversion), I introduce this [changes](https://github.com/SerenityDevstack/spring-cloud-config/commit/2aa59ed62f38399064ad063ef27339c5ad47aa62) in the MongoEnvironmentRepository to allow the definition of any key in the name of the properties. Can you check the changes, please? What do you think about it? 

If you consider appropriate, I can make a PR with these changes to your repository so you can amend it to your PR. Thanks.

venilnoronha:
Hi @javiboo,

While I see that the label and profile values are being lost, I'm still wondering what would the use case be for needing those values. And if someone needs them, would there be 2 different profile/labels values per configuration? If not, then we can update `cleanSource` to retain those properties, which would then be returned in the API response.

My concern is that, having 2 label/profile values might cause ambiguity to the user. Also, would it not be nice to avoid nested properties at the first level itself?

Thoughts?

javiboo:
Hi @venilnoronha,

I agree with you and I don't think it's a good idea to have 2 different profile/labels values per configuration.

By defining a property named label, I'm trying to demonstrate the differences between mongo-db profile and the other profiles.

Suppose we have a configuration file `foo.yml` stored in our fs (with _native_ profile active) for an application with name `foo` with the content:
```
property1: value1
label: value2
```
so if you request your configuration through the endpoint `/{name}/{profiles}` (with values _foo_ and _default_, respectively) you obtain the response:
```
{
  "name": "foo",
  "profiles": [
    "default"
  ],
  "label": null,
  "version": null,
  "propertySources": [
    {
      "name": "classpath:/foo.yml",
      "source": {
        "property1": "value1",
        "label": "value2"
      }
    }
  ]
}
```
With mongo-db profile, how can you address the same content without nesting properties under a field and also without restricting the use of certain properties in the configuration file (considering that here the property _label_ is also being used as a configuration property, a part of the meaning of this field in Spring Cloud)?

What do you think about it?

Thanks!

venilnoronha:
Sure @javiboo, I get your point. We can add a `source` field of type `LinkedHashMap<String, Object>` to MongoPropertySource instead of `properties` (just to be consistent with the API response, see below).

```
"propertySources": [
  {
    "name": "classpath:/foo.yml",
    "source": {
      "property1": "value1",
      "label": "value2"
    }
  }
]
```

Regarding the PR, if you'd like to contribute to my repo I'm open for it. However, I'm not quite sure if you might get the attribution for it (as all of the commits of this PR would be squashed). Again, if you'd want to be sure about the attribution, it would be better if you could send in the PR to [spring-cloud/spring-cloud-config](https://github.com/spring-cloud/spring-cloud-config) directly once this PR is merged. I'm also fine with making the changes you've suggested to my repo directly, based on your fork of my repo (if you don't mind).

Thanks,
Venil

javiboo:
@venilnoronha I don't mind, you can make the changes suggested directly to your repo. Thanks!

venilnoronha:
@javiboo I've updated the implementation. :)

javiboo:
@venilnoronha thanks for the quick response. We are very interested that this PR joins the spring-cloud project, we can do something to expedite this issue?

venilnoronha:
@javiboo I'm not too sure about that. Maybe @spencergibb could tell if there are any immediate plans to merge this PR.

spencergibb:
There are not immediate plans. We'll have a discussion about it as a team.

venilnoronha:
@spencergibb sure, thanks!

spencergibb:
@venilnoronha what do you think about hosting this in your user org, or maybe in https://github.com/spring-cloud-incubator. That way we can see how users respond and gauge interest before we decide to move this into core.

venilnoronha:
How do we go about hosting this under spring-cloud-incubator?

spencergibb:
@venilnoronha https://github.com/spring-cloud-incubator/spring-cloud-config-server-mongodb, invite sent.

venilnoronha:
Thanks @spencergibb. I'll commit the code there soon.

venilnoronha:
Snapshot version of spring-cloud-config-server-mongodb is now available on jfrog oss. Check out the github repository [here](https://github.com/spring-cloud-incubator/spring-cloud-config-server-mongodb) for information on maven configuration and usage.

spencergibb:
Closing in favor of https://github.com/spring-cloud-incubator/spring-cloud-config-server-mongodb/commit/ecedae00e22978155606cf17ca1f2c3279769aff

PR #298, opened 2015-12-04
Fix typo using yml syntax for .properties files

christophe-f:


spencergibb:
It's actually valid properties syntax of you go look at the javadoc.

christophe-f:
You are right. My IDE seems to be the issue.

Issue #293, opened 2015-12-02
Windows error JGitEnvironmentRepository fails to delete .git\objects\*pack


jgolubenko:
I am trying to pull git repo from bitbucket for our @EnableConfigServer app, but it  fails in this method:  `JGitEnvironmentRepository.deleteBaseDirIfExists()`. 

While it might not be exactly spring-cloud-config's problem, because internally it calls eclipse.jgit.* library, but has anyone came across this before?

I've tried to start from eclipse/sts, from cygwin and cmd as Administrator. (I haven't tried it on Ubuntu yet, but will do in couple hours). 

`basedir` configuration property makes no difference apart from checking out in temp dir or under path relative to my app, still the same error.

my application.yml

```
spring:
  cloud:
    config:
      enabled: true
      server:
        git:
          uri: https://www.bitbucket.org/myrepo/ms-configuration
          password: pass
          username: user
          clone-on-start: true
          # basedir: cloud-config/ # if I enable this, makes no difference
```

Exception that is thrown upon startup:

```
Caused by: java.io.IOException: Could not delete file C:\cygwin\tmp\config-repo-7024054885475694217\.git\objects\pack\pack-59244256763eca9b6add9057d20587530bf2001a.pack
        at org.eclipse.jgit.util.FileUtils.delete(FileUtils.java:197) ~[org.eclipse.jgit-4.1.1.201511131810-r.jar:4.1.1.201511131810-r]
        at org.eclipse.jgit.util.FileUtils.delete(FileUtils.java:164) ~[org.eclipse.jgit-4.1.1.201511131810-r.jar:4.1.1.201511131810-r]
        at org.eclipse.jgit.util.FileUtils.delete(FileUtils.java:166) ~[org.eclipse.jgit-4.1.1.201511131810-r.jar:4.1.1.201511131810-r]
        at org.eclipse.jgit.util.FileUtils.delete(FileUtils.java:166) ~[org.eclipse.jgit-4.1.1.201511131810-r.jar:4.1.1.201511131810-r]
        at org.eclipse.jgit.util.FileUtils.delete(FileUtils.java:166) ~[org.eclipse.jgit-4.1.1.201511131810-r.jar:4.1.1.201511131810-r]
        at org.springframework.cloud.config.server.environment.JGitEnvironmentRepository.deleteBaseDirIfExists(JGitEnvironmentRepository.java:280) ~[spring-cloud-config-server-1.1.0.M3.jar:1.1.0.M3]
[...]

PR #309, opened 2015-12-20
fix  bug #293

t4gedieb:
"Windows error JGitEnvironmentRepository fails to delete .git\objects\*pack"

Windows creates a file lock at org.eclipse.jgit.internal.storage.file.PackFile:601 and prevents deleting the base dir.
If there is an existing git client it must be closed before deleting the base dir.


dsyer:
See #293.

PR #315, opened 2015-12-23
fix  bug #293

t4gedieb:
Commit 3f77d806da463bc6f747bf5ec21f7b6ce4c5bceb fixes successfully bug #293. But there is another file lock problem on Windows in the ResourceController. <code>StreamUtils</code> doesn't close the InputStreams and create file locks, which causes problems at *checkout* of another brunch.


dsyer:
Thanks. Did you fill out the contributor's agreement (link in README)?

t4gedieb:
Thanks.. I just filled the contributors agreement.

dsyer:
Merged, thanks. Please use tabs not spaces next time (if you use Eclipse there are format rules in the Spring Boot source code, for IntelliJ I think you can create something analogous).

PR #323, opened 2016-01-26
Small fixes to spring cloud config documentation

tanob:


docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -390,7 +390,8 @@ spring:
   cloud:
     config:
       server:
-        foo: bar
+        overrides:
```
spencergibb: Why add this?

tanob: The documentation says: ```To declare overrides just add a map of name-value pairs to spring.cloud.config.server.overrides```. I thought the correct would be to have the overrides map inside the `overrides` key. Isn't that true?

dsyer:
Thanks for spotting the error. I fixed it independently.

tanob:
@dsyer there was also a typo fix in the PR

PR #346, opened 2016-03-07
Add `resolvePlaceholders` request parameter to `EnvironmentController`.

spencergibb:
Specifically to the `{json|yaml|properties}` endpoints to support sidecar or non-jvm apps to have placeholders resolved automatically.

See https://github.com/spring-cloud/spring-cloud-netflix/issues/877

/cc @dsyer what do you think?

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
@@ -25,7 +25,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeMap;
-
 import javax.servlet.http.HttpServletResponse;
 
 import org.springframework.boot.bind.PropertiesConfigurationFactory;
@@ -43,10 +42,13 @@ import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 import org.yaml.snakeyaml.DumperOptions.FlowStyle;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.nodes.Tag;
+
+import static org.springframework.cloud.config.server.support.EnvironmentPropertySource.resolvePlaceholders;
 
 /**
  * @author Dave Syer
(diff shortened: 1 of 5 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
```diff
@@ -100,32 +102,44 @@ public Environment labelled(@PathVariable String name, @PathVariable String prof
...
 	public ResponseEntity<Map<String, Object>> labelledJsonProperties(
 			@PathVariable String name, @PathVariable String profiles,
-			@PathVariable String label) throws Exception {
+			@PathVariable String label,
+			@RequestParam(defaultValue = "false") boolean resolvePlaceholders)
+			throws Exception {
 		validateNameAndProfiles(name, profiles);
-		Map<String, Object> properties = convertToMap(labelled(name, profiles, label));
+		Environment environment = labelled(name, profiles, label);
+		Map<String, Object> properties = convertToMap(environment);
```
spencergibb: I didn't do json because it wasn't one big string. Two options, 1) convert to `String` and resolve like yaml or 2) resolve individual values in `convertToMap->convertToProperties`.

tkvangorder:
Spencer,

Thanks for the quick turn-around on this. I have confirmed this is working for {properties|yaml} endpoints. Since, in our case, our parent application is a node application, it would be nice if the .json endpoint was supported as well. 


One more gotcha for us is that we are having a problem using the endpoints that resolve the property place holders because our application names match the names we use in our pom files. We often have a "-" in our application name (example "backend-services").

The hyphen works fine on the endpoints that return an Envrionment (but dont resolve place holders) but does not work on any of the endpoints that do resolve the place holders: 

/backend-services/dev/master  <-- Works
/backend-services-dev.json <-- Fails because the name is "backend-services". 

Is the only option for us to rename all of our applications to remove the hyphen or is their an easy way to interpret the last hyphen as the profile?

spencergibb:
Yeah, there's a workaround for hyphens I can't recall, @dsyer do you know? Actually looking at the code, hyphens aren't supported. I didn't do `json` yet becuase it was a different impl, but if @dsyer likes it, I'll do json as well.

dsyer:
Yeah, we should do this (and the JSON). Please open a separate issue for the hyphens (there's probably a reason we don't support it though, e.g. how would you know that profiles don't contain hyphens?).

spencergibb:
@dsyer do you have a preference on strategy? convert json to `String` and resolve like yaml and properties, or resolve each map value individually?

dsyer:
I don't know. Does it matter (given we are already doing string coercion)?

spencergibb:
Likely not.

dsyer:
We've given up on content negotiation with these endpoints, so we might as well just do replacement on the whole body string, I guess.

spencergibb:
@dsyer It's been updated to do the json as well.

spencergibb:
Closed via 828df5262faae75a7c5ac306e2fee323c8714954

Issue #342, opened 2016-02-28
Client side failover reading configuration from cloud config servers


uweschaefer:
[suggestion for enhancement/discussion]

We're using a cloud config server. In order to be resilient to failure of the server, we have multiple instance behind a load balancer. However, in case the load-balancer fails, we'd need some IP-Level failover setup to guarantee access to the config servers.

We were wondering, if it would be easier to allow multiple endpoints for configuration via
spring.cloud.config.uri or a similar property, so that the client can try to configure itself with a secondary uri if the first fails.

Is this something considered of value? if so, we'd like to work on it with a little help.

dsyer:
I think resilient load balancers are free or cheap in pretty much any cloud environment, so I wouldn't make this a high priority. But if you have some code you want to contribute we can definitely take a look. It would be most likely to be successful if it stayed orthogonal to the existing code, so we don't risk making what is now quite simple into something complicated. Maybe the `LoadBalancer` abstraction from spring cloud would be useful?

spencergibb:
I've had others ask for this functionality as well. It's similar to how eureka works too.

uweschaefer:
Just wondering: If we would implement this, how exactly would one USE this?

Provide a comma separated List to "spring.cloud.config.uri" or have a secondary "spring.cloud.config.uris" that has a list?

spencergibb:
I'd start with a comma separated list. Would be simple to change if needed.

fkissel:
Thanks for your feedback!

After talking to my colleague @uweschaefer we would start working on this.

fkissel:
I signed the Individual Contributor Agreement, my confirmation number is 167520160316100217.

PR #353, opened 2016-03-16
Add support for multiple config server addresses

fkissel:
Multiple config server addresses can be set in spring.cloud.config.uri
as a comma separated list. If server discovery is enabled, all provided
service instances are taken.

The result of the first available server is used.

See https://github.com/spring-cloud/spring-cloud-config/issues/342

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -830,8 +830,8 @@ property `spring.cloud.config.uri`) and initializes Spring
 
 The net result of this is that all client apps that want to consume
 the Config Server need a `bootstrap.yml` (or an environment variable)
-with the server address in `spring.cloud.config.uri` (defaults to
-"http://localhost:8888").
+with one or multiple, comma separated server addresses in 
+`spring.cloud.config.uri` (defaults to "http://localhost:8888"). 
 
 [[eureka-first-bootstrap]]
 === Eureka First Bootstrap
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
@@ -16,10 +16,14 @@
 
 package org.springframework.cloud.config.client;
 
+import static java.util.Objects.requireNonNull;
+
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
-import org.springframework.beans.BeanUtils;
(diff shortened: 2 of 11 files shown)


spencergibb:
@fkissel thanks for the submission. It seems quite complex to me. Maybe if we used ribbon via the spring cloud `LoadBalancerClient` which supports retrying on different servers already. @dsyer thoughts?

fkissel:
@spencergibb thanks for your feedback.

I actually started with an different approach that used the LoadBalancerClient abstraction and ribbon. For a couple of reasons I ended up adding a simple for-loop in ConfigServicePropertySourceLocator: 

* Of cause, adding an explizit, non-optional dependency of ribbon to spring-cloud-config-client is not an option. As far as I understand autoconfiguation (including the use of spring.factories-files) this would immediately break existing spring boot setups that uses a different LoadBalancerClient. 
* As an alternative we could require, that an implementation of LoadBalancerClient is added as a dependency to the application. But this would have the following drawbacks:
** If spring.cloud.config.uri would contain a list of uris, fail-fast is enabled and no implementation of LoadBalancerClient is available, an exception could be thrown at startup time. But if service discovery is enabled and multiple config servers are available, only one of them (like today, the first?) could be used.
** In ConfigServicePropertySourceLocator the BasicAuthorizationInterceptor is added only, if no RestTemplate was manually injected via the setRestTemplate method. For consistency reasons we should do it in the same way when adding a LoadBalancerInterceptor. Hence, in setups with manually injected RestTemplate the failover mechanism would not work without some additional changes.

To cut a long story short, this were my must-haves:
* The configuration should be as simple as possible: use a comma separated list, that's it. No need to add additional dependency.
* No side effects to existing spring boot setups. In particular, there should be no additional dependencies which might break existing setups by adding some beans via autoconfigution. (ribbon, retry)
* It should work for service discovery, too.

On the other hand, there are - of cause - some disadvantage:
* No use of the existing LoadBalancerClient abstraction and existing implementations like ribbon.
* You are not able to configure the load balancing (e.g. IClientConfig).
* ConfigServicePropertySourceLocator became more complex (the "single responsibility principle" is violated)

To be honest, it's quite hard for me to weight the arguments against each other. If you say, using LoadBalancerClient should be the way to go (and I should not be to much afraid of breaking any existing setup), that is fine with me.

If you like, I could contribute another solution using the LoadBalancerClient abstraction.

spencergibb:
Closing this due to inactivity. Feel free to reopen if you want to go down the `LoadBalancerClient` route.

Syy0n:
Why was that feature not merged to the current version? The spring config server is not really useful in a production environment if there is no clientside failover...

spencergibb:
@Syy0n it is explained above why we didn't merge in its current form. Our advice so far is to put it behind a platform load balancer.

Issue #363, opened 2016-04-05
config server cache properties for same client and force refresh by git webhook


akaala:
Hi, i adopt spring cloud config server as my configuration management center. 
There are 50+ applications and they ask config server for latest properties at the interval of 1 minute (we want they could notice the changes ASAP).

Then the problem is that config server would pull git frequently (it seems config server pull git per request). This make our git server quite busy.

So could we make config server **cache ** properties for same client (by its application name and profile)? If  we need force refresh the cache, we could add git webhook to trigger that. 
(git webhook function seems added in spring-cloud-config-monitor project, but not designed for this kind of features)

I'm not quite deep into config server codes, If there is misuse or better solution, please let me know.
If this a good and reasonable feature, please add your advice. I could try contribute some codes on this feature.

spencergibb:
Why don't you think spring-cloud-config-monitor with git webhooks is a good enough solution?

akaala:
Thank you for your fast response.

The real problem is that my other applications are not spring projects. They just send HTTP request to spring-config-server to load properties. They are designed to uncouple from configuration service, which means they pull properties at short interval. There is also no "/refresh" API to force refresh runtime properties.

By the way, spring-cloud-config-monitor is only in the latest release, and not available in central maven repositories. Even spring-cloud-config-monitor is deployed to public, lots of codes things should be done if we decide to convert all applications to spring projects.

So i want to do some cache things in config server, make less "git pull" request to git server. 
This is the main propose.

spencergibb:
How different is this than #157

spencergibb:
I think the difference is where the cache is happening, server (this issue) vs. client. Config server is stateless, that's part of it's simplicity. You could certainly add caching to your implementation of config server.

akaala:
@spencergibb thank your for your advice.
I will try to add cache feature in my forked codes.

akaala:
hi, guys, i commit a PR on #366 for this config-server cache feature.

But the implement is a bit wired. The detail is that i have to create a @Autowired cacheService to make "@Cachalbe" works, because the EnvironmentRepository.findOne() is not called by Spring AOP  proxy in EnvironmentController. The origin codes are:

```	@RequestMapping("/{name}/{profiles}/{label:.*}")
	public Environment labelled(@PathVariable String name, @PathVariable String profiles,
			@PathVariable String label) {
		if (label != null && label.contains("(_)")) {
[...]

PR #366, opened 2016-04-11
fix issue #363: config-server cache properties.

jacobke:
make config-server able to cache properties.
details is in issue #363.
(contributor's agreement signed)

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
@@ -27,9 +27,11 @@ import java.util.Map.Entry;
 import java.util.TreeMap;
 import javax.servlet.http.HttpServletResponse;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.bind.PropertiesConfigurationFactory;
 import org.springframework.cloud.config.environment.Environment;
 import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.cloud.config.server.support.ConfigServerCacheService;
 import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.MutablePropertySources;
 import org.springframework.http.HttpHeaders;
@@ -71,6 +73,9 @@ public class EnvironmentController {
 
 	private boolean stripDocument = true;
 
+	@Autowired
+	ConfigServerCacheService configServerCacheService;
+
 	public EnvironmentController(EnvironmentRepository repository) {
 		this(repository, new ObjectMapper());
 	}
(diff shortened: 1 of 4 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
```diff
@@ -71,6 +73,9 @@
 
 	private boolean stripDocument = true;
 
+	@Autowired
```
spencergibb: Use constructor argument

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/ConfigServerCacheService.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.cache.annotation.CachePut;
+import org.springframework.cache.annotation.Cacheable;
+import org.springframework.cloud.config.environment.Environment;
+
+/**
+ * @author: akaala
+ */
+public class ConfigServerCacheService {
```
spencergibb: Seems like you could use the Spring `CacheManage` directly rather than create a class like this.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
```diff
@@ -104,8 +109,20 @@ public Environment labelled(@PathVariable String name, @PathVariable String prof
 			// by Spring MVC
 			label = label.replace("(_)", "/");
 		}
-		Environment environment = this.repository.findOne(name, profiles, label);
-		return environment;
+
+		if (configServerCacheService == null) { // to pass-by EnvironmentControllerTests
```
spencergibb: Caching should be opt-in. There should be a property that enables it.

pivotal-issuemaster:
@jacobke Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=366)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=366) to manually synchronize the status of this Pull Request.

ylquan:
Hi @spencergibb , any update for this feature. I met the same issue with @jacobke , config server pull git per request will make config server very busy.  BTW, do we have any official performance report about config-server. like how many client can support by one config-server. I know there will other solution for HA. but for many clients scene, it is important to know the performance.
I have do some performance test using tools like ab. but the result is not positive. I am a little worry about the performance. Do you have any suggestion? Thanks a lot

PR #391, opened 2016-05-04
WIP! (Do Not Merge) Initial VaultEnvironmentRepository

spencergibb:
### TODO

- [X] Config Server Vault without Watch
- [X] Config Client Token
- [X] Config Server Watch Consul (Basic)
- [X] Config Client Watch
 - [X] scheduled long polling
- [ ] Vault/Token Tests
- [ ] Config Server Watch Consul (Advanced)
- [ ] Documentation

### HOWTO

Run vault.

```
$ vault server -dev
...
Root Token: 9e00b619-6674-cf41-c01f-d31b502c08b1 // copy
...
```

In another terminal:
```
$ export VAULT_ADDR='http://127.0.0.1:8200'
$ vault write secret/application foo=bar baz=bam
$ vault write secret/myapp foo=myappsbar
```
Run a configserver on 8888 with `--spring-profiles-active=vault`.

#### Config Client
In a config client set `spring.cloud.config.token=9e00b619-6674-cf41-c01f-d31b502c08b1`

If you view `/env` you should see something like
```
{
    "profiles": [],
    "configService:vault:myapp": {
        "foo": "myappsbar"
    },
    "configService:vault:application": {
        "baz": "bam",
        "foo": "bar"
    },
    "commandLineArgs": {
        "spring.application.name": "myapp",
        "spring.cloud.config.token": "9e00b619-6674-cf41-c01f-d31b502c08b1",
    },
    "servletContextInitParams": {},
    "systemProperties": {
    },
    "systemEnvironment": {
    },
    "applicationConfig: [classpath:/bootstrap.properties]": {
    }
}
```

#### CLI

Config with no specifics
```
$ http :8888/foo/default X-Config-Token:9e00b619-6674-cf41-c01f-d31b502c08b1
HTTP/1.1 200 OK
Content-Type: application/json;charset=UTF-8
Date: Wed, 04 May 2016 23:04:43 GMT
Server: Apache-Coyote/1.1
Transfer-Encoding: chunked
X-Application-Context: configserver:vault:8888

{
    "label": null,
    "name": "foo",
    "profiles": [
        "default"
    ],
    "propertySources": [
        {
            "name": "vault:application",
            "source": {
                "baz": "bam",
                "foo": "bar"
            }
        }
    ],
    "version": null
}
```

Config for myapp
```
$ http :8888/myapp/default X-Config-Token:9e00b619-6674-cf41-c01f-d31b502c08b1
HTTP/1.1 200 OK
Content-Type: application/json;charset=UTF-8
Date: Wed, 04 May 2016 22:51:31 GMT
Server: Apache-Coyote/1.1
Transfer-Encoding: chunked
X-Application-Context: configserver:vault:8888

{
    "label": null,
    "name": "myapp",
    "profiles": [
        "default"
    ],
    "propertySources": [
        {
            "name": "vault:myapp",
            "source": {
                "foo": "myappsbar"
            }
        },
        {
            "name": "vault:application",
            "source": {
                "baz": "bam",
                "foo": "bar"
            }
        }
    ],
    "version": null
}
```

yml for myapp (notice foo=myappsbar, the most specific value).

```
$ http :8888/myapp-default.yml X-Config-Token:9e00b619-6674-cf41-c01f-d31b502c08b1
HTTP/1.1 200 OK
Content-Length: 24
Content-Type: text/plain;charset=UTF-8
Date: Wed, 04 May 2016 23:06:28 GMT
Server: Apache-Coyote/1.1
X-Application-Context: configserver:vault:8888

baz: bam
foo: myappsbar
```


roblally:
We’ve been looking at the question of Vault and Consul and whether it stores data with predictable paths today and have generally positive news.

Attached is a screenshot of the Consul UI where you can see I stored two keys/value pairs ‘a’ and ‘b’. The keys are not munged, and they’re not likely to be if the closing of this issue [ https://github.com/hashicorp/vault/issues/697 ] holds true. 

But, the overall path doesn’t seem to be completely predictable. When I stored the two keys they ended up in the following paths: 

> vault/logical/EF75A9FF-688A-7786-9446-A8E88F4DB8F5/a
> vault/logical/EF75A9FF-688A-7786-9446-A8E88F4DB8F5/b

The first path segment “vault” comes from a configuration parameter that I used when initializing vault:

```
backend "consul" { address = "127.0.0.1:8500" path = "vault” }
```

The path segment “logical” is one of a number of different paths that vault creates under the “vault” root and seems stable across instantiations. 

The next path segment “EF75A9FF-688A-7786-9446-A8E88F4DB8F5” is a UUID. It is randomly generated when the first Vault node mounts the Consul backend.

https://github.com/hashicorp/vault/blob/master/vault/mount.go#L184-L189

When Vault is restarted it reads the UUID from the backend where it appears to be stored, encrypted, under the Consul path 'vault/core/mounts'

https://github.com/hashicorp/vault/blob/master/vault/mount.go#L477

From what I can discern it is there to allow multiple Vault instances to use the same Consul cluster with the same path prefix without treading on each other. So, the value will be stable, but unpredictable without opening up Consul and manually finding it.

Since we'd need Vault and Consul to be installed before setting up Spring Cloud Config Server in this mode, we can, as you suggested, make the UUID a configuration parameter to Spring Cloud Config Server.

jefferai:
Hi there,

Just to confirm: the paths are stable but are not intended to be public, as I believe the mount UUID is used as a salt in some cases. So I recommend being careful about exposing them publicly. The mounts use UUIDs so that you can have multiple mounts of the same type with different storage; if we used the same mount paths as within Vault, there'd be a risk of overlap if someone managed to unmount a mount but for some reason the storage wasn't wiped as it should be (so it's safety more than practicality).

spencergibb:
To enable continuous polling of the config server in config client set `spring.cloud.config.watch.enabled=true` (default false).
 
Options:
 `spring.cloud.config.watch.initialDelay` is millis to wait to start polling. Default 180000 (3 minutes)
 
  `spring.cloud.config.watch.delay` is millis to wait to start the next watch after the previous has completed. Default 500.

I can tweak the defaults.

roblally:
Hey Spencer,

I've been experimenting with the code and I've a couple of thoughts/questions so far:

- I think we might need to add an extra level of nesting to allow the server to be used by multiple application. At the moment "application" would be shared by all of the applications storing configuration in Vault.
- We'd provisionally discussed using Consul and watches rather than polling - both for scaling and timeliness. What are your thoughts there now?
- With polling any beans with @RefreshScope are getting reloaded every 500 ms, which is likely to be a problem. What are your thoughts there? Watches in Consul would only have been triggered when there were actual changes. At one point, we discussed manually taking a digest of the configuration on the server and only emitting change notifications if it was different. We could implement that idea on the client side.

spencergibb:
The way config server works is that `application.yml` is shared to all apps, this mirrors that. You don't need to give apps access if you don't want any apps to share configuration. You can also use profiles, so `application-<profile>` can be used to share to certain groups of applications. What kind of nesting are you thinking?

A consul watch using the http api blocks up to the timeout. The max is 10 minutes, the default is 5, I've set the default to 3 minutes, that can be changed. You shouldn't be seeing refresh beans every 500ms. `fixedDelayString` is the delay "between" the end of the last run (currently max 3min) and the start of the next.

Currently I see I'm missing on actually checking that the state is different on the client side before calling refresh :-)

Yes, we talked about a more complex algorithm for determining state change rather than just the consul watch index. The client can't know about any algorithm. The state needs to be opaque. It should only know how to determine if it is different.

spencergibb:
Another thing we could do is create a method in `ContextRefresher` that doesn't refresh refresh scope, but only updates the environment. 

@dsyer would `@ConfigurationProperties` get updated without the refresh scope update?

roblally:
You're right about the polling, I had the configuration wrong and it was using the vault polling mechanism which does reload the config every time.

On the application.yml: one of the issues we perceived with the git implementation is that we would need a repo/server per client application because sharing an application.yml we would rarely if every want to do, and certainly something that we wouldn't want to do across, potentially, hundreds of applications.

I had been thinking that, instead of:

```sh
vault write secret/application foo=bar baz=bam
```

We could have an extra level of path:

```sh
vault write secret/myapp/application foo=bar baz=bam
```

That would allow us to use a single vault/consul cluster for many different applications without them treading on each other's configuration namespace.

roblally:
One other area I have a question about:

Here:

https://github.com/spring-cloud/spring-cloud-config/blob/vault/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepository.java#L151

we're mapping the vault response to a Map<String, String>. Vault supports map/json properties which we were hoping to use to be able to map to properties of the form a.b.c.d. 

Is there a structural issue in doing this with Spring Cloud Config?

spencergibb:
## Config Server Vault configuration 
The following are available under `spring.cloud.config.server.vault`
- `enabled` defaults to `false`
- `host` defaults to `127.0.0.1`
- `port` defaults to `8200`
- `scheme` defaults to `http`
- `backend` defaults to `secret`
- `defaultKey` defaults to `application`. Set to empty to disable.
- `profileSeparator` defaults to `,`

spencergibb:
Now supports nested keys in data. For example:
```
$ echo -n '{"appA": {"secret": "appAsecret"}, "bar": "baz"}' | vault write secret/testConsulVaultAppA -
Success! Data written to: secret/testConsulVaultAppA

$ vault read -format=yaml secret/testConsulVaultAppA
data:
  appA:
    secret: appAsecret
  bar: baz
lease_duration: 2592000
```

spencergibb:
Closed via 753db0ed224af3e634fe9d29cb0dfd7f71205650

ryanjbaxter:
We should also call out that `spring.cloud.config.token` needs to be placed in `bootstrap.yml` in order to work.

Issue #364, opened 2016-04-07
Exception is not logged when pulling


prademak:
We are using spring-cloud-config 1.1.0.M3, and currently repeatedly see a warning being logged:
`Could not pull remote for master (current ref=Ref[refs/heads/master=386bfc56eb3b12de6e638db20180aedabcdef]), remote: git@bitbucket.org:company/config.git`

However, since the Exception is not logged it makes it more difficult to find the actual cause of the problem. 

Therefore I would like to propose the following change at JGitEnvironmentRepository, line 220:
```java
catch (Exception e) {
	this.logger.warn("Could not pull remote for " + label + " (current ref=" + ref
			+ "), remote: " + git.getRepository().getConfig().getString("remote",
			"origin", "url"));
}
```
change to:
```java
               catch (Exception e) {
			this.logger.warn("Could not pull remote for " + label + " (current ref=" + ref
					+ "), remote: " + git.getRepository().getConfig().getString("remote",
							"origin", "url"), e);
		}
```

skhome:
I would like to see the error logging being improved as well.

We faced a similar problem which we were not able to track down to it's root cause.
We had an exception being thrown at the place mentioned above and suspected also an exception being thrown in `MultipleJGitEnvironmentRepository#getLocations` but the 
```
catch (Exception e) {
	continue;
}
```
did hide the root cause.

While I totally agree that the config-server should be as resilient as possible and should rather serve requests from a local potentially stale cache then not at all, having more information available to fix the root cause would be helpful.

spencergibb:
@dsyer any objections to the above changes?

dsyer:
I think it would be too noisy to log the whole stack trace because it's not an error. The multi repo service is trying to find a matching repository for the request. It has to try all of them till it finds one that works. Unless I'm missing something this is quite intentional.

skhome:
My problem is not in the way the server handles the situation, but that it does not provide any clues that help me find the root cause.
How do you feel about logging the exception on DEBUG level, so I could then temporary increase the log verbosity when needed.

dsyer:
I don't like the idea of logging the stack trace at all. We could log the fact that there was an exception at DEBUG level. Do you want to propose a change?

tjuchniewicz:
It would be great to log (even in DEBUG without stacktrace) error in `MultipleJGitEnvironmentRepository#getLocations`

```java
catch (Exception e) {
    continue;
}
```

Now the only way for me to find root cause was debugging.

PR #393, opened 2016-05-06
Improve logging of JGit exceptions

skhome:
This pull request is suggestion related to Issue #364 (https://github.com/spring-cloud/spring-cloud-config/issues/364)

In some situations the JGit library can fail to update the local Git repository.
While the config server should gracefully handle these situations some added
information from the root cause might be helpful to pinpoint why changes in the
remote repository do not get reflected in the REST API.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -205,8 +205,14 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 	}
 
 	private boolean shouldPull(Git git, Ref ref) throws GitAPIException {
-		return git.status().call().isClean() && ref != null && git.getRepository()
-				.getConfig().getString("remote", "origin", "url") != null;
+		final boolean isWorkingTreeClean = git.status().call().isClean();
+		final String originUrl = git.getRepository().getConfig().getString("remote",
+				"origin", "url");
+		if (!isWorkingTreeClean) {
+			this.logger.info("Cannot pull from remote " + originUrl
+					+ ", the working tree is not clean.");
+		}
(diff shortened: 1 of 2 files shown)


dsyer:
In #364 I suggested DEBUG level. Can we go with that for the multi-repo try/catch?

spencergibb:
ping @skhome

skhome:
I have made the change to the log level. Any additional comments on this PR?

spencergibb:
@dsyer do you want both `debug` in `MultipleJGitEnvironmentRepository`?

dsyer:
Yes, please, both logs have to be the same level. They should be protected with `if(logger.isDebugEnabled())` as well.

skhome:
Sorry, I am confused. I only aded one log statement in `MultipleJGitEnvironmentRepository`, the other one was already there and on `info` level.
Why do you you want to hide even more information that is needed for trouble shooting?

dsyer:
It's better to hide it IMO because it's normal for repositories not to contain any given file you ask them for. I couldn't see any difference between the two log statements from that point of view. Maybe I was wrong?

skhome:
Hmm, I seem to be not qualified for this kind of pull request. I was talking the whole time about exceptions coming from the underlying JGit library. Cases where JGit simply leaves corrupted Git repositories and the config server is unable to recover from this situation itself.

I have no idea what you mean with "it's normal for repositories not to contain any given file you ask them for" in the context of exception handling.

Currently in our setup with an in-house hosted BitBucket server backend the config server behaves in some non-deterministic fashion (not serving new commits, not recognizing new or deleted tags, sometimes recognizing new commits only on the 2nd request). In this state I can not use it for a production environment.

Hiding the exceptions does not help me in any way to trouble shoot what is causing this.

I rest my case, since this goes into the opposite direction as was my intention.

dsyer:
I don't understand why you don't want logging if you have a problem you need to diagnose. Do you want to re-submit?

spencergibb:
Closed with some polish via 0dda21df749350234d2b6ba6a5fcce010f3ababf

PR #416, opened 2016-06-09
Uses ./mvnw instead of mvn

mavenraven:
Uses mvnw to prevent users from getting stuck due to errors (like I did).

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -32,7 +32,7 @@ Start the server:
 
 ----
 $ cd spring-cloud-config-server
-$ mvn spring-boot:run
+$ ./mvnw spring-boot:run
 ----
 
 The server is a Spring Boot application so you can run it from your


dsyer:
Good catch. Thanks. You might notice the "do not edit" banner at the top of this file. The actual content is probably here: https://raw.githubusercontent.com/spring-cloud/spring-cloud-build/master/docs/src/main/asciidoc/building-base.adoc (and it already refers to the wrapper, so this one is probably just waiting for a refresh).

mavenraven:
Thanks @dsyer! It looks like this block is incorrect as well
```
$ curl localhost:8888/foo/development
{"name":"development","label":"master","propertySources":[
  {"name":"https://github.com/scratches/config-repo/foo-development.properties","source":{"bar":"spam"}},
  {"name":"https://github.com/scratches/config-repo/foo.properties","source":{"foo":"bar"}}
]}
```
It looks like it's listening on `8080`, and I don't get anything served at `/foo/development`.

dsyer:
If you follow the instructions, it runs on port 8888 I think. I noticed a reference to "mvn" not in that build link as well, so I'll fix that independently.

PR #422, opened 2016-06-21
using array index properties, e.g. YAML arrays (at least with a singl…

mbenson:
…e element), causes the existing array value to be modified; guard the static array from accidental modification


mbenson:
This also keeps the config server from appending the default locations.

PR #426, opened 2016-06-24
Cache config client health indicator.

spencergibb:
So config server isn't bombarded with requests.

Should `health.config.enabled` default to `false`?

/cc @dsyer @marcingrzejszczak @adriancole 

I'd also bump version to 1.2.0 for Camden.


spencergibb:
Copied from boot's health mvc endpoint caching.

spencergibb:
@ryanjbaxter yeah I think so. I need to rebase.

spencergibb:
@ryanjbaxter rebased, check it out and merge if you approve.

ryanjbaxter:
@spencergibb I think it would be a good idea to add something to the docs about this.  Property to enable the functionality (it is disabled by default right?), default cache length, and property to configure the cache length.

spencergibb:
Will do

spencergibb:
@ryanjbaxter added docs section

PR #442, opened 2016-07-12
Incorrect version returned on first request after git commit

rlynch:
There is a bug where the version returned on the first request after an update is committed to git incorrectly returns the previous version (commit ID).  This was because the Ref returned was from the checkout prior to the pull.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -148,6 +148,7 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 			Ref ref = checkout(git, label);
 			if (shouldPull(git, ref)) {
 				pull(git, label, ref);
+				ref = checkout(git, label);
 			}
 			return ref;
 		}


spencergibb:
Usually when we fix a bug, we create a test.

rlynch:
I actually tried to create a test but the exiting unit tests only use a local repository so as a result the code to do a git pull will never be exercised.

rlynch:
Some more details on the lack of unit testing for this regarding this bug.

The findOne_FileAddedToRepo_FindOneSuccess test in JGitEnvironmentRepositoryIntegrationTests is a good example of a unit test which demonstrates the difference between testing with a local git repo vs a remote repo.  The test actually adds a new file to git (locally) and commits it which on the surface would be a perfect test.  I actually started implementing this fix by adding an additional assertion at the end of this test to verify the version changes before and after the commit.  I first tested it prior to applying the fix to verify the additional assertion triggers the bug.  Unfortunately, in my testing I found the bug didn't present itself because the call to shouldPull() inside of the refresh() method always returns false with a local repository so therefore the pull() method never gets called.  It works in a local repository because the unit test has already committed the file locally thus the call to checkout() return a ref with the current version.  With a remote repository shoudPull() always returns true so the pull method gets called which is when the updated version becomes available.  Below is the code snippet of the refresh() method prior to the fix.


		try {
			git = createGitClient();
			git.getRepository().getConfig().setString("branch", label, "merge", label);
			Ref ref = checkout(git, label);
			if (shouldPull(git, ref)) {
				pull(git, label, ref);
			}
			return ref;

rlynch:
Hi @spencergibb, how do I move forward on this?  Should I just convert this to a bug or do you have recommendations on how we can approach getting coverage for this in the unit testing?

daniellavoie:
I am also struggling on finding a way to test a remote repository. I think creating a Bare repository and setting it as a remote could do the trick. This way the pull could be simulated locally. I'll let you know if I come up with something.

daniellavoie:
As promised, take a look at this [PR](https://github.com/spring-cloud/spring-cloud-config/pull/483/commits/ea14ec439adc37e3164f5045e2608688ac1cd85d) I made. I am using a bare repository to operate push and pull commands. Should answer your testing needs.

rlynch:
local environment mess up so submitting a new pull request. closing this one.

Issue #481, opened 2016-08-19
Git Environment Repository does not handle remotes updated from a forced push.


daniellavoie:
If you ever force push on a branch served by the config server using an ssh git backend, the config server local repository will be left in a conflicting state. Accessing the HTTP endpoint will even serve you the json ENV properties with git conflict markers `>>>>>>>>>>` and `==========`.

I suggest that the `JGitEnvironmentRepository` operates a `git fetch` then `git reset --hard` operation rather than a simple `git pull`.

daniellavoie:
I can come up with a PR but I'll need some inspiration to mock and reproduce the git operations on a remote. I am curious to figure out which approach was taken to unit test the `JGitEnvironmentRepository`. I doubt I'll be able to simulate a remote who diverged with a `git push --force`. Guess I'll have to stick with actual manual testing and make sure the current tests does not encounter any regression.

daniellavoie:
Hello @ryanjbaxter. Any chances this will be part of Candem M2 ? Ciao !

spencergibb:
I was just starting to look at this again

ryanjbaxter:
@daniellavoie once @spencergibb finished taking a look at it I am sure we can get it merged

PR #483, opened 2016-08-21
Handle Git Environment Repository updated from a forced push.

daniellavoie:
Fixes gh-481.

The unit test might not seem obvious so feel free to ask about it. If you want to fully understand the reason behind this PR, checkout my branch, revert the `JGitEnvironmentRepository` changes I made and run the `JGitEnvironmentRepositoryIntegrationTests.pullDivergedRepo()` test. You will see that the local repositroy will be left in a conflicting state. It represents the issue I encountered in production. 

In order to fix this issue, I changed the `git pull` operation performed by `JGitEnvironmentRepository` to a `git fetch` then `git reset --hard`. IMO, hard is reset is acceptable since the local repository should not be modified.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -24,14 +24,18 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
-import org.eclipse.jgit.api.CloneCommand;
 import org.eclipse.jgit.api.CheckoutCommand;
+import org.eclipse.jgit.api.CloneCommand;
 import org.eclipse.jgit.api.CreateBranchCommand.SetupUpstreamMode;
+import org.eclipse.jgit.api.FetchCommand;
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.ListBranchCommand;
 import org.eclipse.jgit.api.ListBranchCommand.ListMode;
 import org.eclipse.jgit.api.PullCommand;
+import org.eclipse.jgit.api.ResetCommand;
+import org.eclipse.jgit.api.ResetCommand.ResetType;
 import org.eclipse.jgit.api.Status;
+import org.eclipse.jgit.api.StatusCommand;
 import org.eclipse.jgit.api.TransportCommand;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.api.errors.RefNotFoundException;
@@ -55,6 +59,7 @@ import com.jcraft.jsch.Session;
  * @author Dave Syer
  * @author Roy Clarkson
  * @author Marcos Barbero
+ * @author Daniel Lavoie
  */
 public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 		implements EnvironmentRepository, SearchPathLocator, InitializingBean {
@@ -166,6 +171,14 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 			Ref ref = checkout(git, label);
 			if (shouldPull(git, ref)) {
 				pull(git, label, ref);
+
+				if (!isClean(git)) {
+					logger.warn("The local repository is dirty. Reseting it to origin/"
+							+ label + ".");
+
+					fetch(git, label, "origin");
+					resetHard(git, label, "refs/remotes/origin/" + label);
+				}
 			}
 			return ref;
 		}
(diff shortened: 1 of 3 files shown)


daniellavoie:
Built on my side with a 1.8 JDK with 1.7 compatibility mode. I'll switch to a 1.7 JDK to fix the travis issues.

dsyer:
Thanks for the code. I think I prefer to make this an opt in, and keep the old behaviour unless the user sets a flag. Does that make sense? I suppose we could do the opposite, but I'd like to keep the old code anyway.

daniellavoie:
No problem ! You are afraid of the overhead of `git fetch + reset --hard` over the classic `git pull` ?

The problem I had with this issue is that is hard to investigate You really need to dig down on your config server payload to find out the nasty `=====`. It doesn't blow up to your face. Properties are still served without errors by the Config Server. The symptoms you encounter are outdated configuration. Activating this fix by default might save a lot of time to a lot of people. 

I understand you prefer limiting the impact of the existing users of the Config Server. It's a hard choice.

daniellavoie:
I have updated the PR by implementing the `pullOverReset` property which allows to revert back to the original pull operation rather than fetch and reset. I have describe the property within ascidoc. Feel free to comment. I will have to fork Spring Cloud Server with this fix for our production requirements until this PR is merged.

ryanjbaxter:
This looks good to me.  A small nit about the pull request....there are a couple of spots where there are new lines or changes to code formatting that don't really seem necessary.  If you wouldnt mind cleaning those up, that would be great.

daniellavoie:
Hello @ryanjbaxter 

I agree with you on the code formatting. I just applied Spring Cloud code formatter from Eclipse. It feels to me those classes were not previously using the latest formatter. I can remove these but then it won't be possible to apply the formatter to my latest contribution. I've got mixed feeling on that.

ryanjbaxter:
@daniellavoie OK if its from the code formatter no big deal!

spencergibb:
You should be able to selectivity format code

daniellavoie:
But the way, wait before merging. I've came accross an interesting case that I need to rework for. I need to add the force update to the fetch operation. My unit test passes because my local repository is already aware of the commits from the remote. On a real scenario, the fetch would fail with a reject.

daniellavoie:
@ryanjbaxter @spencergibb 

I went through the code and found out I didn't had to configure any force update parameter onto the fetch command as this parameter is part of the refspec. That refspec is built on the clone operation and is set to force by default (the little + sign in front of the resources). The Egit plugin of Eclipse raised me some doubts as I came across some rejected fetch command. This PR is free to go unless you want me to rework the non modified sections formatting.

ryanjbaxter:
👍 LGTM

spencergibb:
> I think I prefer to make this an opt in, and keep the old behaviour unless the user sets a flag.

This currently looks like opt-out, it should be opt-in.

spencergibb:
Also, how is this different from 218d140c7e8e329d11d3c7e9a181215b58092cad?

spencergibb:
Disregard last comment.

daniellavoie:
@dsyer argues that it should be it should be deactivated by default. Given the trouble I had to investigate this issue, I feel it would save a kitten or two if it is activated by default. It might not be obvious since most people have a clean workflow (or relies on HTTP protocol). But from to moment you are on ssh and you push force a conflicting commit, you will be left in a conflicting state. 

Users of the framework won't have the reflex to activate this option as symptoms are not obvious  (no error, only possible old configs being served). You have some concerns on the reliability of the `fetch / reset --hard` over classic `pull` ?

spencergibb:
@dsyer said
 > I think I prefer to make this an opt in, and keep the old behaviour unless the user sets a flag.

Your code does not `keep the old behaviour unless the user sets a flag`, it uses the new behavior _unless_  a flag is set. All I'm asking is to change the default of `pullOverReset=true` and change the docs to match.

daniellavoie:
I understand what you are asking but I don't agree that think it is the best. As I said multiple times, the old behavior leads you to a bug if you run into the conditions I described. That bug is hard to investigate to people won't have the reflex to activate the new behavior.

daniellavoie:
I've put some though into it and I came up with this. Tell me what you think. 

I drop the option thing, and keep the old behavior by default. BUT, after executing the `pull` comannd, I could call `isClean` on the git repository to check if it is in a clean state. If not, I trigger my new `reset --hard` mechanism. This way, new behavior will only be triggered on my special edge case. This way, no confusing obscure magic parameter. It will just work whatever the case.

dsyer:
Sounds promising. Does the clean test cost any network traffic (if it does then it might slow things down compared to today)?

daniellavoie:
According to EGit, the clean attribute is performed on a `StatusCommand`. Which ends up being a `git status` command. I can get the `isClean` info by running a `git.status().call().isClean().

This operation is already executed before `pull`. Network traffic should not occur when running a git status. It only computes the states of the local branches.

spencergibb:
@daniellavoie sounds like a good idea.

daniellavoie:
Here we go ! Final revision in my opinion. Let me know what you think about it.

PR #485, opened 2016-08-23
fix svn repo to also check branches and tags folders for label

transamericamoon:
Currently {label} only works with 'trunk' for svn repository (using the standard svn folder layout), This fix allows to look for label inside branches and tags folder.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SvnKitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SvnKitEnvironmentRepository.java
@@ -183,9 +183,9 @@ public class SvnKitEnvironmentRepository extends AbstractScmEnvironmentRepositor
 		File svnPath = new File(workingDirectory, label);
 		if(!svnPath.exists()) {
 			svnPath = new File(workingDirectory, "branches" + File.separator + label);
-		}
-		if(!svnPath.exists()) {
-			svnPath = new File(workingDirectory, "tags" + File.separator + label);
+			if(!svnPath.exists()) {
+				svnPath = new File(workingDirectory, "tags" + File.separator + label);
+			}
 		}
 		return svnPath; 
 	}
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SvnKitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SvnKitEnvironmentRepository.java
@@ -179,7 +179,15 @@ public class SvnKitEnvironmentRepository extends AbstractScmEnvironmentRepositor
 
 	private File getSvnPath(File workingDirectory, String label) {
 		// use label as path relative to repository root
-		return new File(workingDirectory, label);
+		// if it doesn't exists check branches and then tags folders
+		File svnPath = new File(workingDirectory, label);
+		if(!svnPath.exists()) {
+			svnPath = new File(workingDirectory, "branches" + File.separator + label);
+		}
+		if(!svnPath.exists()) {
+			svnPath = new File(workingDirectory, "tags" + File.separator + label);
+		}
+		return svnPath; 
 	}
 
 }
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@transamericamoon Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=485)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=485) to manually synchronize the status of this Pull Request.

spencergibb:
How about a test?

pivotal-issuemaster:
@transamericamoon Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=485)!

transamericamoon:
here's couple.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SvnKitEnvironmentRepository.java
```diff
@@ -179,7 +179,15 @@ protected File getWorkingDirectory() {
 
 	private File getSvnPath(File workingDirectory, String label) {
 		// use label as path relative to repository root
-		return new File(workingDirectory, label);
+		// if it doesn't exists check branches and then tags folders
+		File svnPath = new File(workingDirectory, label);
+		if(!svnPath.exists()) {
+			svnPath = new File(workingDirectory, "branches" + File.separator + label);
+		}
+		if(!svnPath.exists()) {
```
ryanjbaxter: Won't this logic always execute if the path doesn't exist?  What is the point of the previous `if`?

transamericamoon: I'm not sure what your asking.  It checks the root folder for the label, if it doesn't exists it checks the branches folder for the label, if label doesn't exists there it checks the tags folder for the label.

transamericamoon: I see what your saying, i guess i can nest it.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SvnKitEnvironmentRepository.java
```diff
@@ -183,9 +183,9 @@ private File getSvnPath(File workingDirectory, String label) {
 		File svnPath = new File(workingDirectory, label);
 		if(!svnPath.exists()) {
 			svnPath = new File(workingDirectory, "branches" + File.separator + label);
-		}
-		if(!svnPath.exists()) {
-			svnPath = new File(workingDirectory, "tags" + File.separator + label);
+			if(!svnPath.exists()) {
```
ryanjbaxter: is the intention to return the branches or tags path if the original path does not exist?  what if both the branches and tags paths don't exist then we return the path to tags, is that intentional? 

I guess my question is should we only be setting `svnPath` if either the branches or tags directories exist, otherwise leave the behavior as it is today?

transamericamoon: This is covered in IntegrationTest - public void invalidLabel(), if the folder doesn't exists an exception will be thrown.  Either way you will be returning a path that does not exists (whether it is the tags folder or trunk folder should be irrevelant)

transamericamoon:
I guess it makes more sense to throw the exception immediately if the label doesn't exist.

ryanjbaxter:
Can we add some tests for the error cases where branches and labels don't exist?

transamericamoon:
Test already exist, look at: SVNKitEnvironmentRepositoryIntegrationTests:invalidLabel()

ryanjbaxter:
Hum the test already exists....is it testing your new code which also throws the same exception?  Sorry not so familiar with the code, just trying to review it.

transamericamoon:
Yes, but now it gets the exception sooner rather than later.

ryanjbaxter:
OK I see now, it was here https://github.com/spring-cloud/spring-cloud-config/blob/2670aaaf244501af1f9cd8e083549578a3f65757/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SvnKitEnvironmentRepository.java#L98.

So can we take out the check for exists in `getPaths` since `getSvnPath` will do the checking?

transamericamoon:
No, we can't take it out because getSearchLocations():

https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/AbstractScmAccessor.java#L175

 appends the search paths, so we still need to check whether the path exists once the search locations are appended.

Issue #480, opened 2016-08-19
resolvePlaceholder functionality incorrect and inconsistent when serving alternative formats


rlynch:
I am seeing problems with the resolvePlaceholder functionality in Spring Cloud Config Server.  A set of examples is best.  First here is an example yml file I have in my git repo....

```
jsk:
  servlet:
    healthcheck:
      dirToWatch: '${catalina.home:.}/conf/_health'
```
When I call the server using the path used by the Spring Client the correct results are returned with the ${catalina.home:.} placeholder not being resolved.

**_http://localhost:8888/sp_boot_sample/foo/ryan_test_**
```
{
  "name": "sp_boot_sample",
  "profiles": [
    "foo"
  ],
  "label": "ryan_test",
  "version": "ed19d558ff52e08aeb72a517e46685a344389fd1",
  "propertySources": [
    {
      "name": "https://github.intuit.com/java-service-kit/sp-boot-sample-config/application.yml",
      "source": {
        "jsk.servlet.healthcheck.dirToWatch": "${catalina.home:.}/conf/_health"
      }
    }
  ]
}
```

It also behaives correctly when I request the configuration as properties with the resolvePlaceholder=false
**_http://localhost:8888/ryan_test/sp_boot_sample-foo.properties?resolvePlaceholders=false_**
```
jsk.servlet.healthcheck.dirToWatch: ${catalina.home:.}/conf/_health
```

YML and JSON is where we have problems since it is trying to replace the placeholder.

**_http://localhost:8888/ryan_test/sp_boot_sample-foo.json?resolvePlaceholders=false_**
```
{
  "jsk": {
    "servlet": {
      "healthcheck": {
        "dirToWatch": "./conf/_health"
      }
    }
  }
}
```

**_http://localhost:8888/ryan_test/sp_boot_sample-foo.yml?resolvePlaceholders=false_**
```
jsk:
  servlet:
    healthcheck:
      dirToWatch: ./conf/_health
```

The issue gets more problematic when you remove the resolvePlaceholders as the behavior becomes inconsistent. 
**YML**
**_http://localhost:8888/ryan_test/sp_boot_sample-foo.yml_**
```
jsk:
  servlet:
    healthcheck:
      dirToWatch: ./conf/_health
```
**JSON**
**_http://localhost:8888/ryan_test/sp_boot_sample-foo.json_**
```
{
  "jsk": {
    "servlet": {
      "healthcheck": {
[...]

PR #492, opened 2016-09-06
Fixes handling of default values in placeholders for json and yaml files.

rlynch:
Fixes handling of default values in placeholders for json and yaml files.  Fix was to create the properties map manually instead of using the PropertiesConfigurationFactory which forces default value evaluation.  This might be somewhat more efficient too but I haven't tested that yet. 
Completes the fix for gh-480 which had already been closed but didn't cover this case.


ryanjbaxter:
The approach sounds good, but I am having a problem digesting what is happening in `recursiveKeyValueToMap`.  Maybe you can break this one method up into some smaller more concrete methods?

rlynch:
Okay, I have broken up the method, added more comments, and renamed some variables to try and improve readability of this pr.  Please let me know if you have any questions.

ryanjbaxter:
@rlynch thanks for refactoring the code.  It looks better and the comments certainly help.  Would you mind merging in the changes I pushed a few minutes ago?  They will enable builds on Circle CI and also do generate some code coverage information.  I am curious to see how much of our tests are going to cover the new code, especially the edge cases.

spencergibb:
I'd like @dsyer to chime in.

dsyer:
Seems like a big change when I'm not sure what the problem is. Can we get a crisp explanation of what the behaviour is before and after?

rlynch:
Placeholders with default values always get processed even when resolvePlaceholders=false for json and yaml (properties work).  Using your unit test as an example, if have a property a.b.c=${foo:spam} and I request it with with resolvePlaceHolders=false like so, http://localhost:8888/a_test/sample-foo.json?resolvePlaceholders=false, it will return with a.b.c=spam when it should return with a.b.c=${foo:spam}.

The existing unit tests in EnvironmentContorllerTests acknowledge this shortcoming in its comments....

```
@Test
	public void placeholdersResolvedInJsonFromSystemPropertiesWhenNotFlaggedWithDefault() throws Exception {
		whenPlaceholdersSystemPropsWithDefault();
		String json = this.controller.jsonProperties("foo", "bar", false).getBody();
		// If there is a default value we can't prevent the placeholder being resolved
		assertEquals("{\"a\":{\"b\":{\"c\":\"spam\"}}}", json);
	}
```

rlynch:
@ryanjbaxter - I have merged with latest from master and you can see the coverage.  We could add some tests to handle bad arrays but overall think the coverage looks pretty good on the new code but I will leave the final call you to guys :-)

rlynch:
Please hold off on merging this PR. I found a bug not converted in the test cases. Simple fix related to arrays with missing elements. Fix is simple and doesn't alter the overall approach so I would still like feedback.

ryanjbaxter:
@rlynch I agree the coverage is pretty good.  I would consider adding a test case for the `NumberFormatException`.  That code path is not hit.

rlynch:
I have added additional unit tests (and associated fixes) for array edge cases.  The following use cases are now tested. This also covers your ask of hitting the `NumberFormatException`

- Missing Array Elements
-- a.b[0].c=d
-- a.b[2].c=e

- Array elements with negative numbers
-- a.b[-1].c=d

- Array with invalid number
-- a.b[non_a_number].c=d

ryanjbaxter:
@rlynch what else does this fix address in addition to what @dsyer fixed already in [f8fc4e1](https://github.com/spring-cloud/spring-cloud-config/commit/f8fc4e19375d3b4c0c2562a71bc49ba288197100)?

rlynch:
@ryanjbaxter

Placeholders with default values always get processed even when resolvePlaceholders=false for json and yaml (properties work). if have a property a.b.c=${foo:spam} and I request it with with resolvePlaceHolders=false like so, http://localhost:8888/a_test/sample-foo.json?resolvePlaceholders=false, it will return with a.b.c=spam when it should return with a.b.c=${foo:spam}.

ryanjbaxter:
I understand the problem, but it seems like @dsyer already submitted a fix for this, did it not work?

rlynch:
@ryanjbaxter it fixed the bug partially. Placeholders with no default were fixed.

rlynch:
@ryanjbaxter 
To further clarify.  The fix provided by @dsyer - https://github.com/spring-cloud/spring-cloud-config/commit/f8fc4e19375d3b4c0c2562a71bc49ba288197100 removed system and environment properties from placeholder resolution which reduced the impact of this issue but didn't completely fix the problem. The remaining issue is that placeholders with default values always have the placeholder replaced with the default value even if you have replacePlaceholders=false.  If the placeholder doesn't have a default value then the current approach seems to work.

The problem (I believe) stems from the fact that there are two separate code paths when handling properties vs json and yaml.  For properties (which work) the environment is retrieved and then simply converted to properties within the EnvironmentController.  If placeholder resolution is requested then a separate method is called to perform that action.  For yaml and json the environment is retrieved and then the configuration is converted to a map using PropertiesConfigurationFactory.  The PropertiesConfigurationFactory has the side effect of always processing all placeholders with default values which I believe to be incorrect in this use case.  

The change I made was to have json/yaml processed in a similar way as properties by dealing with the conversion internally and thus avoiding the side effect caused by PropertiesConfigurationFactory.

dsyer:
Maybe we should be fixing `PropertiesConfigurationFactory`? Or maybe not (I never liked that step, really, but it was expedient), but only if this change is really covering all the corner cases.

rlynch:
@dsyer - I completely agree that this does introduce some risk since I am assuming `PropertiesConfigurationFactory` has been around the block for a while now and I totally see the benefit of being expedient :-)  I have tried to add a good hunk of additional use cases especially in how it handles arrays which is where a lot of the logic and complexity reside.
 
One other area I would like to bring up. `PropertiesConfigurationFactory` also treats corner cases differently than I believe Spring Cloud should but some folks could certainly disagree.  For some array edge cases `PropertiesConfigurationFactory` will throw a `org.springframework.beans.InvalidPropertyException` exception in the following use cases.
-- Array elements with negative numbers
-- Array with invalid number

My solution is to treat these not as array failures but as just to not interpret them as arrays.  This PR could throw an error for this situation but I don't think the `InvalidPropertyException` is appropriate.

ryanjbaxter:
It personally feels wrong to me to reimplement most of the functionality in `PropertiesConfigurationFactory` just to avoid the issue instead of trying to change it to fit our needs.  However my experience here is limited and I think @dsyer can make the call as to which direction is right.

rlynch:
Certainly is a good issue to discuss and not a slam dunk either way.  IMHO, `PropertiesConfigurationFactory` existence is about binding properties to beans not about conversion of an environment to json/yaml.  The need for the addArray method in EnvironmentController (eliminated in this PR) emphasises that `PropertiesConfigurationFactory` is an imperfect solution.  @dsyer I would guess this is what you meant by "I never liked that step, really, but it was expedient" but my wife always told me I am horrible at reading minds.  

The more relevant point in my mind is that this PR is a fix to a bug that is available today.  The nature of the change doesn't really build up tech debt since it would be simple to revert it to use `PropertiesConfigurationFactory` should it better support this use case in the future.

rlynch:
@dsyer @ryanjbaxter - Is this still on your radar?  This is a bug we are trying to get resolved and I would like to avoid dealing with a forked code base.  I am happy to make any changes you feel are needed to ensure the safety of this change.  Let me know how I can help.

dsyer:
I feel like the fix should be in `PropertiesConfigurationFactory`. I'll see if I can add a flag there.

PR #508, opened 2016-09-22
Better support custom PropertyPathNotificationExtractor

whboyd:
Currently, the way EnvironmentMonitorAutoConfiguration is set up in spring-cloud-config-monitor does not allow the configuration of custom PropertyPathNotificationExtractor beans without overriding and replacing the default PropertyPathNotificationExtractors for Github, Bitbucket, etc. This is due to a combination of how spring boot autoconfig bean overriding works coupled with the fact that the PropertyPathNotificationExtractor beans are declared within the same configuration class that autowires the collection of PropertyPathNotificationExtractors (namely, PropertyPathNotificationExtractor).

This change moves the default PropertyPathNotificationExtractor beans into a separate configuration class, which allows users to define their own custom PropertyPathNotificationExtractors which will then be added to the list and processed in addition to the default PropertyPathNotificationExtractors rather than instead of them.


spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
```diff
@@ -31,7 +31,7 @@
  */
 @Configuration
 @ConditionalOnWebApplication
-@Import(FileMonitorConfiguration.class)
+@Import({FileMonitorConfiguration.class, PropertyPathNotificationExtractorConfiguration.class})
```
spencergibb: Can you try with a `protected static class` in this same class rather than another class with an import?

whboyd: done :)

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
```diff
@@ -41,23 +42,28 @@
...
+                        return new GithubPropertyPathNotificationExtractor();
+                }
 
-	@Bean
-	@ConditionalOnProperty(value="spring.cloud.config.server.monitor.github.enabled", havingValue="true", matchIfMissing=true)
-	public GithubPropertyPathNotificationExtractor githubPropertyPathNotificationExtractor() {
-		return new GithubPropertyPathNotificationExtractor();
-	}
+                @Bean
+                @ConditionalOnProperty(value="spring.cloud.config.server.monitor.gitlab.enabled", havingValue="true", matchIfMissing=true)
```
spencergibb: Formatting, we use tabs

whboyd: fixed

PR #518, opened 2016-09-29
Support nested arrays when translating properties to map

chiknrice:
I've replaced the previous implementation which only supports array at leaf level of yaml.  The new implementation is based on how snakeyaml builds a nested map from a yaml file (i.e. LinkedHashMap to represent the dictionaries and ArrayList to represent arrays).  The new implementation also removes the need for switching off placeholder resolution as the operation of converting an environment to a map doesn't really involve setting placeholder values.

### DIFF
--- a/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/EnvironmentControllerTests.java
+++ b/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/EnvironmentControllerTests.java
@@ -267,6 +267,8 @@ public class EnvironmentControllerTests {
 		map.put("a.b[0].d[2]", "yy");
 		map.put("a.b[0].d[0]", "xx");
 		map.put("a.b[2].c", "y");
+		map.put("a.b[3][0]", "r");
+		map.put("a.b[3][1]", "s");
 		this.environment.add(new PropertySource("one", map));
 		Mockito.when(this.repository.findOne("foo", "bar", null)).thenReturn(this.environment);
 		String yaml = this.controller.yaml("foo", "bar", false).getBody();
@@ -281,7 +283,9 @@ public class EnvironmentControllerTests {
 				"  - null\n" +
 				"  - c: y\n" +
 				"    e:\n" +
-				"    - d: z\n";
+				"    - d: z\n" +
+				"  - - r\n" +
+				"    - s\n";
 		assertThat("Wrong output: " + yaml, yaml, is(expected));
 	}
 
--- a/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/EnvironmentControllerTests.java
+++ b/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/EnvironmentControllerTests.java
@@ -15,7 +15,11 @@
  */
 package org.springframework.cloud.config.server.environment;
 
-import static org.hamcrest.Matchers.*;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
>  The new implementation also removes the need for switching off placeholder resolution as the operation of converting an environment to a map doesn't really involve setting placeholder values.

That makes it sound like it might be related to #480, is it?

rlynch:
@ryanjbaxter - yes it is very similar to #480

ryanjbaxter:
Kind of wondering what the added benefit is then, is it just the support for nested arrays?

rlynch:
@ryanjbaxter - Looks like.  #480 was focused on solving the bug that default placeholders were always being processed even when handlePlaceholders=false.  This PR is effectivly doing the same thing but to solve the nested array problem which I think #480 handled as well but I am not sure. The interesting part is that this PR took the same approach as #480 by removing `PropertiesConfigurationFactory` as the mechanism to convert properties to a map.   It would be good to at least fold in the additional unit test #480 had to this PR to gain confidence that it is handling some of the edge cases.

chiknrice:
@ryanjbaxter @rlynch The change is mainly about supporting nested arrays - yes that's the only benefit of this change.  The removal of the resolvePlaceholders param in convertToMap method is just a side effect of the change (which is a good because the resolvePlaceholders concern is being performed in the endpoint methods which require it).  And no, the change doesn't relate to, or trying to fix #480

chiknrice:
@rlynch - Referring to your comment "Its during the bindPropertiesToTarget that the placeholder resolution occurs" in #480 then the changes in this PR would definitely make that issue in convertToMap irrelevant.  I also investigated what fixes were applied related to #480 (which is 53e049a) and the unit test provided by that change definitely passes, or are there other commits related to #480?

chiknrice:
@rlynch Apologies, I'm not that informed on how the workflow is in github (collaborating with others).  Now I realize you have been fixing this in your forked src (seeing all the links to the commits at the bottom of #480).  Just let me know if you have a use for this change or if you prefer I just close and abandon it.  Thanks.

rlynch:
@chiknrice No worries. The Spring team decided to use a change to `PropertiesConfigurationFactory ` instead of the one I provided to fix the placeholder issue I was having. My PR was somewhat similar to yours so I was just adding context.

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/EnvironmentControllerTests.java
```diff
@@ -15,10 +15,7 @@
  */
 package org.springframework.cloud.config.server.environment;
 
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.hasSize;
-import static org.hamcrest.Matchers.not;
-import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.*;
```
spencergibb: We don't use `*` imports.

chiknrice:
@spencergibb I've addressed both code review comments.

spencergibb:
@dsyer concerns about eliminating the use of `PropertiesConfigurationFactory`?

dsyer:
I'm not sure. Didn't we decide in #480 it was better to fix the spring boot side, and not try to change too much here? Wouldn't the same argument apply to this change?

chiknrice:
@spencergibb the role of `PropertiesConfigurationFactory` was to bind the properties after part of `convertToMap` has created a skeleton/structure of the nested Map.  Was thinking why not set the values while building the structure?

@dsyer I believe that's why the `convertToMap` method was created is because spring boot doesn't have a functionality where properties/Environment are translated back to their yaml form so not sure where it could be fixed in spring boot side?

chiknrice:
The change also supports multi dimension arrays, FYI.

chiknrice:
All checks pass now, any chance of approval? :)

spencergibb:
@chiknrice I marked it for team discussion. We'll talk about it soon.

PR #523, opened 2016-10-11
Document how to use custom RestTemplate

### LINKED ISSUE #501: Document how to use custom RestTemplate in config-client
Is there any way to use custom RestTemplate object that is used in config-client?

The reason why I need this is that we are working in distributed system. So if we would like to connect some https application then we are creating restTemplate with ClientHttpRequestFactory. Within ClientHttpRequestFactory we are setting sslConnectionSocketFactory property in such a way that we don't need to do deal with installing trust store etc. 

In normal applicatin it works fine, but Now we are moving config-server to https, So to connect with it config-client use RestTemplate object and that is generated within config-client. 

I want a some way that config-client use restTemplate object that is generated by me.

ryanjbaxter:
Fixes #501

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -1162,9 +1162,53 @@ spring:
 
 ----
 
-If you use another form of security you might need to provide a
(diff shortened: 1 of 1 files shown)


ryanjbaxter:
@spencergibb could you please review for accuracy, thanks!

PR #532, opened 2016-10-24
Added /encrypt/validate endpoint

mlueders:
We've had a number of instances where someone has checked in an encrypted value which ended up being incorrect (wrong key, character accidentally deleted before commit, etc).  We've added this endpoint to our config server which allows us to run validation checks on config repo commit.

I figured if we ran into this issue, other people may as well, so figured I'd raise a PR.  If the functionality is something y'all would like to add but in a different form, let me know.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
@@ -121,6 +121,17 @@ public class EncryptionController {
 		return encrypt(this.defaultApplicationName, this.defaultProfile, data, type);
 	}
 
+	@RequestMapping(value = "/encrypt/validate", method = RequestMethod.POST)
+	public ResponseEntity<String> validate(@RequestBody String data,
+			@RequestHeader("Content-Type") MediaType type) {
+		try {
+			decrypt(data, type);
+			return new ResponseEntity("Valid", HttpStatus.OK);
+		} catch (IllegalStateException ex) {
+			return new ResponseEntity<String>("Invalid", HttpStatus.BAD_REQUEST);
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@mlueders Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=532)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=532) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mlueders Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=532)!

ryanjbaxter:
This seems useful to me.  We would want to add documentation about the endpoint.

@dsyer @spencergibb what do you think?

spencergibb:
Seems a bit redundant to me. Would just calling decrypt do the same thing?

Blackbaud-MikeLueders:
@spencergibb yes and no.  It would have the same affect but it would also return the decrypted value to the client.  I want to open this endpoint up on the production config server such that it can be called from a jenkins build which can be run to validate whether the configuration is valid, after any change to the config repository.  A proxy prevents decrypt from being called from our build environment.

Does this use case make sense?

Also, once I get confirmation this is something y'all would want to add, I'll update the PR with endpoint documentation.

spencergibb:
OK, makes sense. Why the two different usernames?

Blackbaud-MikeLueders:
Ah, didn't realize - this is my work account, other is personal.  Largely depends on what time of day I'm doing things :)

Blackbaud-MikeLueders:
I'll work on getting documentation added in the next day or two

spencergibb:
ping @mlueders on docs

spencergibb:

Issue #439, opened 2016-07-09
Config Server not fetching new remote branches


imberda:
There is currently an issue with the spring-cloud-config-server (1.1.0-RELEASE) when using a Git backed repository. The config-server does not allow getting config for remote branches added after the config-server initially clones the repository.

Steps to re-produce:
1. Start config server using remote Git repository
2. Make initial request to config server for existing branch (i.e. master)
3. Create a new branch in GitHub (e.g. new_branch)
4. Make a request to config-server for /some-app-name/default/new_branch

Expected behavior is the config for the "new_branch" is returned. Actual behavior is `NoSuchLabelException`.

The issue seems to stem from the fact that `JGitEnvironmentRepository` does not perform a fetch (`FetchCommand`) to find newly added branches created remotely, prior to attempting a checkout

This issue can _seem_ hard to re-produce if the config-server is being actively used and configuration is being requested for branches which are already known locally. This is because this causes a pull (`PullCommand`) which implicitly causes a fetch to occur.

This issue is likely the same (root cause) as #265, but I think this issue more succinctly describes the issue without other factors being involved.

As an aside, the rationale for wanting newly added remote branches to be found without a restart relates to the way the config-server is used within my development team. We have a single deployed instance of the config-server which all developers use as a shared service. Each developer starts whatever service they are working on locally (in their IDE) specifying a config-client  label to be their current feature branch. They then expect to be able to start their service and receive the config for that label.

Krustie101:
Somewhat related to this issue in its origin is a problem with moving tags in the repository. It might be possible to fix both issues in one go. In any case, the root cause is to be found in the refresh method of the JGitEnvironmentRepository.

1. Start server using a git repository with an existing tag.
2. Make initial request to config server for this existing tag.
3. Change configurations in the remote git repository and move the tag in the remote git repository to the new commit.
4. Make a subsequent request to the config server for this tag.

We expect that the new configurations will be loaded, but the old ones will be loaded because the tag is not moved in the git repository on the config server. In the refresh method, the checkout call for a tag returns a null ref, so that no pull is performed.

rlynch:
I have submitted a PR which I believe addresses this issue. #538

spencergibb:
Closed via 4c9806b4d699c5e27a9a62c5d1eeaf8029bbba47

PR #538, opened 2016-10-25
New Branch/Tags Not Available and Git Version Not Updated On First Request

rlynch:
This addresses issue #439 along with some other issues.  This PR was originally addressed just to fix a bug I found where the first request to spring cloud server after a git update results in the prior version number being returned along with the new config.  However, during further testing, I realized that the fix was only partial.  Here is a list of issues it resolved...

- SCCS would return an incorrect version on the first request after the following use cases.
  - An update to an existing branch on the Git server.
  - When retrieving a tag the version would be null
  - When a hardReset was done due to a merge error.
- SCCS would return a NoSuchLabelException when requesting a tag or branch that was created after the initial clone. #439

I have also added some easy setup code to create tests which operate off a local and remote repository which allows for more robust testing of "real world" git scenarios.  I have also added many new unit tests to cover these use cases.


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -171,6 +171,7 @@ private Ref refresh(String application, String label) {
 			Ref ref = checkout(git, label);
 			if (shouldPull(git, ref)) {
 				pull(git, label, ref);
+				ref = git.getRepository().getRef(ref.getName());
```
spencergibb: Please put a comment here as to why this is needed.

spencergibb:
Awesome test BTW

rlynch:
@spencergibb - requested comment is done.  Please review.

rlynch:
@ryanjbaxter and @spencergibb 
After doing some more testing with this fix I discovered some uncovered use cases where this fix didn't return the correct version. Specifically, the system would return the prior version (with correct config) on the first request after the local repo was pulled.  

While looking into this I also discovered why new branches/tags were not showing up if they were created after the initial clone as described in #439. The reason was twofold.
1. Since checkout was being called prior to the pull, any new branches or tags didn't exist locally so it would throw a NoSuchLabelException.
2. The JGit pull doesn't offer the ability to add a --tags so even if we did the pull prior to the checkout it still wouldn't see new tags.  
The fix was to first do a fetch prior to the checkout (unless shouldPull is false) so we can see the new tags/branches.  We also no longer do a pull, but instead do a separate fetch(so we can add the --tags option) and merge.  A pull is basically just a fetch and merge so this is fine.

Fixing both issues were fairly intertwined so I decided to submit this as one PR.  

As an added bonus I have added some good test setup code to make testing git calls with a local repo that replicates with a remote repo we can make changes to.  This makes the unit test more "real".  Several new unit tests were added to cover server-side git changes.

Krustie101:
@rlynch 
In your last comment, you mention you do a separate fetch allowing you to add the "--tag" flag but I do not see this in your code (the TagOpts.FETCH_TAGS option is not set on the FetchCommand). As a consequence, a moved tag will not be fetched from the remote repo.

rlynch:
Wow, it was included in my brain. I will add it along with a test case. Thanks for pointing out my miss!

rlynch:
@Krustie101 - TagOpts.FETCH_ALL added to fetch and unit test extended to cover this case.  Thanks again for finding this.

Krustie101:
@rlynch 
Thank you for putting in the effort to improve the config server. Much appreciated!

rlynch:
@spencergibb @ryanjbaxter - Guys, any comment(s) on this?

spencergibb:
@rlynch I haven't had a chance to review the new changes. It's in my queue.

Issue #586, opened 2016-12-12
Use Spring Vault for VaultEnvironmentRepository


mp911de:
`VaultEnvironmentRepository` uses `RestTemplate` and plain Jackson to access Vault. A migration to `VaultClient` or even `VaultTemplate` from https://github.com/spring-projects/spring-vault can help removing extra code and benefit from error handling. Spring Vault provides additional configuration options (such as HTTP-client specific SSL/Timeout configuration).

PR #587, opened 2016-12-13
Use Spring Vault for VaultEnvironmentRepository

mp911de:
`VaultEnvironmentRepository` now uses Spring Vault and VaultClient to access Vault. Using `VaultEnvironmentRepository` requires the `org.springprojects.vault:spring-vault-core` dependency.

See gh-586.


spencergibb:
ug, why no circle build

PR #595, opened 2016-12-19
Support for composite environment repos

### LINKED ISSUE #495: Support a composite EnvironmentRepository
Then, for example, you could have normal config data in git, with secrets in vault.

ryanjbaxter:
Fixes #495


docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -556,6 +556,58 @@ $ vault write secret/application foo=bar baz=bam
...
+spring:
+  profiles:
+    active: git, svn
+  cloud:
+    config:
+      server:
+        svn:
+          uri: file:///path/to/svn/repo
+		  order: 2
+		git:
```
spencergibb: spacing looks finny here

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
```diff
@@ -83,6 +81,10 @@ public MultipleJGitEnvironmentRepository environmentRepository() {
 	}
 
 	@Configuration
+	@Profile("git")
```
spencergibb: clever

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/OrderedEnvironmentRepository.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.springframework.cloud.config.server.environment;
+
+import org.springframework.core.Ordered;
+
+/**
+ * A prioritized {@link EnvironmentRepository}.
+ * @author Ryan Baxter
+ */
+public interface OrderedEnvironmentRepository extends EnvironmentRepository, Ordered {
```
spencergibb: I think this interface isn't needed is it? I think it could be removed and all your tests would still pass.

ryanjbaxter: I wanted an interface that combined both the `EnvironmentRepository` and `Ordered` so I can inject them into the configuration class.

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -556,6 +556,58 @@ $ vault write secret/application foo=bar baz=bam
...
+The priority order of a repository will help resolve any potential conflicts between
+repositories that contain values for the same properties.
+
+NOTE:  Any type of failure when retrieving values from an environment repositoy
+will result in a failure for the entire composite environment.
+
+NOTE:  When using a composite environment it is important that all repos contain
+the same label(s).  If you have an environment similar to the one above and you request
+configuration data with the label `master` but the SVN
+repo does not contain a branch called `master` the entire request will fail.
```
dsyer: This seems like an unwanted restriction. I don't want to have to go to all repositories and create a new label every time I do it in one of them.

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -556,6 +556,58 @@ $ vault write secret/application foo=bar baz=bam
...
+
+===== Custom Composite Environment Repositories
+
+It is also possible to provide your own `EnvironmentRepository` bean
+to be included as part of a composite environment in addition to
+using one of the environment repositories from Spring Cloud.  To do this your bean
+must implement `OrderedEnvironmentRepository`.  For convenience sake, there is
```
dsyer: Why do we need a new interface? Surely it just has to implement `EnvironmentRepository` and `Ordered`?

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -556,6 +556,58 @@ $ vault write secret/application foo=bar baz=bam
...
+an abstract implementation of this interface you can extend called
+`AbstractOrderedEnvironmentRepository`.  If you extend `AbstractOrderedEnvironmentRepository`
```
dsyer: ...so the abstract base class is probably not helpful (and I'm not sure we want to advertise it in the user guide anyway)?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
```diff
@@ -76,6 +76,10 @@ public void add(PropertySource propertySource) {
 		this.propertySources.add(propertySource);
 	}
 
+	public void add(List<PropertySource> propertySources) {
```
dsyer: "addAll"?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/CompositeConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+	@Primary
+	@ConditionalOnBean(SearchPathLocator.class)
+	public SearchPathCompositeEnvironmentRepository compositeEnvironmentRepository() {
+		return new SearchPathCompositeEnvironmentRepository(environmentRepos);
+	}
+
+	@Bean
+	@Primary
+	@ConditionalOnMissingBean(SearchPathLocator.class)
+	public CompositeEnvironmentRepository compositeEnvironmentRepository2() {
```
dsyer: Lame bean name. Surely we can name it better than that?

dsyer:
Looking good. Some comments inline above.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/CompositeConfiguration.java
ryanjbaxter: some of my hacking escaped my cleanup :)

docs/src/main/asciidoc/spring-cloud-config.adoc
ryanjbaxter: @spencergibb had a similar concern.  I also felt like a new interface was not necessary.  However I want to be able to inject only the `EnvironmentRepository`s that also implement `Ordered` to be used in the `CompositeEnvironment` [configuration](https://github.com/spring-cloud/spring-cloud-config/pull/595/files#diff-88c6b4614cc8766d200234a86a6f0741R56).

The other way of doing this would be to make `EnvironmentRepository` extend `Ordered`, but I didn't want to do that in order to maintain backward compatibility.  If we are not concerned about backward compatibility (which we may not be since this will be in Dalston) than I can do that instead.

The final way of doing this, would be to inject all `EnvironmentRepository`s and check if they implement `Ordered`.  If they don't we could just stick them at the end of the ordered list (essentially giving them the lowest priority possible).

docs/src/main/asciidoc/spring-cloud-config.adoc
dsyer: Yeah, but a missing label is probably not really an error in this scenario, I guess?

docs/src/main/asciidoc/spring-cloud-config.adoc
dsyer: Spring already does that, so I think we can rely on the library to provide a comparator (unordered has a standard default order). Just tell users that unordered has the normal Spring semantics.

docs/src/main/asciidoc/spring-cloud-config.adoc
spencergibb: Unless the client sends a label (which it doesn't by default), each `EnvironmentRepository` should already choose a default.

ryanjbaxter: @dsyer yes a missing label should not be an error.

spencergibb:
👍

PR #614, opened 2017-01-23
Add ability specify base dir which will not be cleaned after restart

vasilievip:
Here is the case:
- my git repo is inaccessible
- I've got nodes with config server crashing (require new instance startup)

I'll probably loose my config server under these conditions in production. It would be nice to 

a) allow to specify some base path so I can use mounted folder to keep files
b) use same folder on startup

Does this make any sense to implement? I tried to use repo uri as the key for each checkout, so if there is overlapping repos by key, they might get corrupted

NOTE: Not sure who decided to instantiate MultipleJGitEnvironmentRepository via ConfigurationProperties, but this is really strange decision. Its complaining that  PatternMatchingJGitEnvironmentRepository must have default constructor? Really?

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerProperties.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerProperties.java
@@ -68,6 +68,11 @@ public class ConfigServerProperties {
 	private String defaultProfile = "default";
 
 	/**
+	 * Base directory for local working copy of repository. If not set the default temporary-file directory will be used.
+	 */
+	private String baseDir;
+
+	/**
 	 * Decryption configuration for when server handles encrypted properties before sending them to clients.
 	 */
 	private Encrypt encrypt = new Encrypt();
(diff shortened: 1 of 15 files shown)


pivotal-issuemaster:
@vasilievip Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=614)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=614) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@vasilievip Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=614)!

ryanjbaxter:
Not sure I understand the need for this.  Restarting the server or creating a new instance would just cause the config repo to be cloned again.

Your git repo does not need to be remote today, and we already have the ability to use the native file system for configuration properties.

vasilievip:
native file system might be the option (while I have to maintain repo on some shared store)

PR #619, opened 2017-01-27
Fixing a typo in the documentation

eliasgomes:
There's a typo in the **Vault Server** section:
...
"When using Vault as a backend you can share configuration with all applications by placing configuration in `html5/application`. For example, if you run this Vault command"
...

When it should be:
...
"When using Vault as a backend you can share configuration with all applications by placing configuration in `secret/application`. For example, if you run this Vault command"
...

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -622,7 +622,7 @@ part of the server.
 
 When using Vault as a backend you can share configuration with
 all applications by placing configuration in
-`{backend}/application`.  For example, if you run this Vault command
+`secret/application`.  For example, if you run this Vault command
 
 [source,sh]
 ----


pivotal-issuemaster:
@eliasgomes Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=619)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=619) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@eliasgomes Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=619)!

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -622,7 +622,7 @@ part of the server.
 
 When using Vault as a backend you can share configuration with
 all applications by placing configuration in
-`{backend}/application`.  For example, if you run this Vault command
+`secret/application`.  For example, if you run this Vault command
```
ryanjbaxter: Hum must be some weird variable replacement issue.  

I cant remember off the top of my head but I think it could be any backend not just `secret`, I think that was the purpose of `{backend}`.  I think what we want to change it to is `\{backend}`.

eliasgomes: I can imagine that was probably the idea, but everywhere else in the file is hardcoded to `secret`. Which is the default value anyway and used by all the examples.

ryanjbaxter: Thats a good point, and we also specify that by default the backend is `secret` and how to change it, so I guess this makes sense.  Thanks!

PR #626, opened 2017-02-02
Add exception stacktrace to logs

vasilievip:
Its hard to guess root cause by having only "Could not fetch" in logs

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -197,7 +197,7 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 			return git.getRepository().getRef("HEAD").getObjectId().getName();
 		}
 		catch (RefNotFoundException e) {
-			throw new NoSuchLabelException("No such label: " + label);
+			throw new NoSuchLabelException("No such label: " + label, e);
 		}
 		catch (GitAPIException e) {
 			throw new IllegalStateException("Cannot clone or checkout repository", e);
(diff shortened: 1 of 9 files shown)


spencergibb:
We've had this discussion before. The multiple git environment will have errors by design (not found). Adding these exceptions will produce false errors and fill logs in some cases. Putting them as debug would let you dynamically change the level to see problems.

spencergibb:
I don't agree on changing warns to errors either

vasilievip:
Ok, I changed stacktrace logging only when debug is enabled and switched back from error to warn.
Regarding EncryptionController - I found IllegalArgumentException or IllegalStateException cases only on decrypt, don't see where it can be thrown on encrypt - there is key validation done upfront which throws KeyNotInstalledException.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -484,6 +486,14 @@ private boolean containsBranch(Git git, String label, ListMode listMode)
 		return false;
 	}
 
+	protected void warn(String message, Exception ex) {
+		if(logger.isDebugEnabled()) {
+			logger.warn(message, ex);
```
spencergibb: should be `.debug()`.

vasilievip: why? 
if debug enabled new data will be sent but message level will stay as warn. If I search for errors and warnings in kibana I will not find this message anymore, it will be confusing.

spencergibb: It's a standard we follow, we don't check if one level is enabled, then log another. Feel free to log the warn AND then the debug.

vasilievip: done

spencergibb:
Closed with some polish via 8c8591cf53755622b7eceee467c638d7e2634baa

PR #627, opened 2017-02-03
Add jgit health indicator

vasilievip:
Add HEAD commit info from jgit repository to /health endpoint

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AbstractScmEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AbstractScmEnvironmentRepository.java
@@ -16,6 +16,9 @@
 
 package org.springframework.cloud.config.server.environment;
 
+import java.text.DateFormat;
+import java.util.Date;
+
 import org.springframework.cloud.config.environment.Environment;
 import org.springframework.cloud.config.server.support.AbstractScmAccessor;
 import org.springframework.core.Ordered;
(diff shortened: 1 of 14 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AbstractScmEnvironmentRepository.java
```diff
@@ -44,8 +47,15 @@ public synchronized Environment findOne(String application, String profile, Stri
...
-		return this.cleaner.clean(result, getWorkingDirectory().toURI().toString(),
-				getUri());
+		StringBuilder state = new StringBuilder();
+		if(locations.getTimestamp() > 0) {
+			Date date = new Date(locations.getTimestamp());
+			state.append(DateFormat.getDateTimeInstance().format(date));
+			state.append(" ");
+		}
+		state.append(locations.getInfo());
+		result.setState(state.toString());
```
spencergibb: The `Environment.state` was included for client side state, not server side.

vasilievip: not sure I understood the problem. If Environment is client data structure, why then its used everywhere on server? If field state is client only, why there is no ClientEnvironment with this field and ServerEnvironment w/o this field. Should I add 1 or2 more fields to Environment to transfer this information? Why this field can't be used on server side?

spencergibb: It's a representation of the client state.

vasilievip: Added new field

PR #535, opened 2016-10-24
Add configuration option to allow for a passphrase to be specified.

cfraser:
- Add `passphrase` configuration for SSH connections
- add `strict.host.key.checking` configuration flag
- fixes #431
- addresses `strictHostKeyChecking` request in #101

PR #630, opened 2017-02-07
Default StrictHostKeyChecking to true

cfraser:
* defaulted strict host key checking to true and updated test


cfraser:
As suggested by @shakuzen in #535.

Sorry this took so long, it completely slipped my mind. 😦

Issue #631, opened 2017-02-08
Proposal for failover design


asarkar:
I'm looking at [JGitEnvironmentRepository](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java) and pretty much everything happens in the `refresh` method. There are more than one problems with it:

1. It clones the repo every time. See https://github.com/spring-cloud/spring-cloud-config/issues/566
2. It doesn't have failover behavior.
3. It is a private method, thus ruling out AOP that clients may want to do to handle the above items.

I propose that the method be made `public`, and that it falls back to using the local repo if remote clone fails. Question I've is, **what version should be returned** in that case? The one retrieved from local repo or something to indicate that the clone operation failed (I'm thinking `UNKNOWN`)?

Optionally, we could add a retry logic using project Reactor keeping it consistent with Spring 5, with configurable properties. This wouldn't be in my first pass, because making it public means the user can do AOP to attempt retries.

ryanjbaxter:
This sounds useful.  I suppose if you are going to fallback to using the local repo, you try and use the same version as what was requested from the remote repo.  If that is not possible maybe it should just fail?  

+1 to the retry logic as well, although I suggest you use Spring Retry, as that is what we are using in spring cloud netflix.

asarkar:
@ryanjbaxter 
> If that is not possible maybe it should just fail?

I don't want it to fail no matter what. According to the 12-factor app, we must design keeping network problems in mind and failing config would in turn fail all apps that depend on it. That's not acceptable. What do you suggest returning the version in that case?

spencergibb:
In another issue, it was mentioned to enhance the CompositeEnvironmentRepository as a way to failover. Have a normal git repo, then a native one that could fall through as the backup.

There is retry already on the client side.

asarkar:
@spencergibb I believe the retry on the client side is different. The client retries to connect to config server if not successful. AFAIK, there's no retry for a request which actually triggers the git clone.
What's the other issue you referred to?

spencergibb:
Yup, retry is client side. This is the issue https://github.com/spring-cloud/spring-cloud-config/issues/617#issuecomment-276683893

asarkar:
[...]

PR #634, opened 2017-02-09
Making refresh modifier public so that clients can do AOP failover

asarkar:


spencergibb:
Why do you need to do AOP on `refresh()`? It's only called from one place `getLocations()`, couldn't you do it there?

asarkar:
@spencergibb Not without extending `AbstractScmAccessor`, methods from which are used in `getLocations`.

asarkar:
@spencergibb What does the team think? This is really a trivial change, so if you don't see a problem with it, I'd like to get it merged. It'd help the ambitious clients.

spencergibb:
@asarkar can you explain in more detail why AOP wouldn't work around `getLocations()` while it would around `refresh()`?

asarkar:
It's not that `getLocations` can't be intercepted, that's not hard. The question is what the alternative implementation might be when we don't want to execute the original method.  Currently `getLocations` uses `protected` methods from `AbstractScmAccessor`, thus making it harded to provided an alternative without having another subclass of `AbstractScmAccessor`.
If you have a suggestion for an alternative `getLocations` without subclassing `AbstractScmAccessor`, I can do that.

spencergibb:
I think it would be helpful to step back and see **why** you are trying to do this rather than implementation details. That would help drive how to achieve your purposes.

asarkar:
I believe this https://github.com/spring-cloud/spring-cloud-config/issues/631 answers your question.

dendron8:
@asarkar do you have an example of how you used this to provide failover? did you failover to a native/local repo?

PR #644, opened 2017-02-16
Use RestTemplate for Vault access.

mp911de:
Use RestTemplate for Vault access and remove Spring Vault dependency. VaultEnvironmentRepository used VaultClient but Spring Vault no longer provides VaultClient.

See also gh-586, gh-587.


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepository.java
```diff
@@ -15,48 +15,43 @@
  */
 package org.springframework.cloud.config.server.environment;
 
+import static org.springframework.cloud.config.client.ConfigClientProperties.*;
```
spencergibb: no * imports

Issue #650, opened 2017-02-23
Allow custom token name in ConfigServicePropertySourceLocator


mouellet:
My config-server is setup behind an API Gateway and registered clients are required to pass an `apikey ` token header but the header name is hard coded in `ConfigClientProperties`.

I need a way to customize the header name, ideally with configuration properties.

ryanjbaxter:
Are you referring to its use in `ConfigServicePropertySourceLocator`?

mouellet:
Yes, in the `getRemoteEnvironment` private method, not in the `VaultEnvironmentRepository`... 

Here is a similar issue in spring-security-oauth (https://github.com/spring-projects/spring-security-oauth/issues/296)

spencergibb:
I'm not sure adding a configurable token name is the right thing to do here. That header name has specific meaning at least in `VaultEnvironmentRepository` and adding an option to change the header name in `ConfigClientProperties` would require a corresponding change on the server side. Maybe using a custom `RestTemplate` with an interceptor would make more sense.

mouellet:
Your right, changing that header name doesn't look like the right thing to do. However, I would really prefer not having to use a custom `RestTemplate` for all my clients...

What about a map of optional headers in the `ConfigClientProperties`? Those could be added to the request in `ConfigServicePropertySourceLocator`.

mouellet:
By adding a `Map` in `ConfigClientProperties`:
```java
private Map<String, String> headers;
```
And adding those properties in the `getRemoteEnvironment` private method of `ConfigServicePropertySourceLocator`:
```java
HttpHeaders headers = new HttpHeaders();
for (Entry<String, String> customHeader : properties.getHeaders().entrySet()) {
    headers.add(customHeader.getKey(), customHeader.getValue());
}
```
Now any client can have custom headers without using a custom `RestTemplate`:
```yaml
spring:
  cloud:
    config:
      headers:
        Authorization: Basic ...
        X-Example-Version: ...
        X-Example-Client-Id: ...
```
The `BasicAuthorizationInterceptor` (and maybe the `GenericAuthorization` interceptor with some logic) could no longer be needed.

I believe this would be the most extensible and developer friendly way to add custom headers to a config client request.

Thoughts?

Thanks!

spencergibb:
Closed via 45336cd

dtitov:
Is this feature implemented? I can't find any docs mentioning:
```yaml
spring:
  cloud:
    config:
      headers:
[...]

Issue #613, opened 2017-01-20
Allow simpler way to provide custom RestTemplate to ConfigServicePropertySourceLocator 


spencergibb:
See http://cloud.spring.io/spring-cloud-static/Camden.SR4/#custom-rest-template

Currently requires disabling the client and creating your own bootstrap configuration and a custom ConfigServicePropertySourceLocator.

Not really developer friendly.

PR #656, opened 2017-02-27
Adds ability to add static headers to config client request.

mouellet:
This commit provides an alternative implementation for creating
config client request using 'ConfigurationProperties' instead
of having to create custom 'RestTemplate' when adding header to
the request.

See discusions at gh-650

Not covering gh-613

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -16,10 +16,10 @@
 
 package org.springframework.cloud.config.client;
 
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Map.Entry;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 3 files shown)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -187,67 +182,38 @@ public void setRestTemplate(RestTemplate restTemplate) {
...
-		}
-		else if (authorization != null) {
-			template.setInterceptors(Arrays.<ClientHttpRequestInterceptor> asList(
-					new GenericAuthorization(authorization)));
-		}
-
 		return template;
 	}
 
-	private static class BasicAuthorizationInterceptor implements
```
ryanjbaxter: Is there a reason why we couldnt have kept this interceptor and just generalized it to add all the headers?

mouellet: Nope, Is it better to add headers using interceptors then `RequestEntity`?
It didn't seems right have 2 ways to add headers for a request and I've chosen the `RequestEntity` way. But I have no problems going for the interceptor, just let me know and I'll make the changes.

ryanjbaxter: I don't think your approach is necessarily wrong, it would just be less churn in the code if you kept the interceptor and generalized it to add the new headers as well.

mouellet: @ryanjbaxter, I pushed the changes you ask, let me know if you want me to squash my commits!

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -191,62 +193,47 @@ private RestTemplate getSecureRestTemplate(ConfigClientProperties client) {
 		SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
 		requestFactory.setReadTimeout((60 * 1000 * 3) + 5000); //TODO 3m5s, make configurable?
 		RestTemplate template = new RestTemplate(requestFactory);
+		String username = client.getUsername();
 		String password = client.getPassword();
 		String authorization = client.getAuthorization();
+		Map<String, String> headers = client.getHeaders();
```
spencergibb: I would create a new Map here and copy values. There are no guarantees that the map is mutable.

mouellet:
Should I also mark the `spring.cloud.config.authorization` property as deprecated (with `@DeprecatedConfigurationProperty`)?

spencergibb:
Squashed and merged via b559735377c48c137d347d74b0151c0ffb61a86d

PR #548, opened 2016-11-12
Add AWS CodeCommit authentication support

dlaidlaw:
Added new code for an AWS CodeCommit credential provider, updated existing code to use same.
The code only creates the AWS credential provider if the git uri matches the AWS CodeCommit pattern, like: https://git-codecommit.$(AWS_REGION}.amazonaws.com/${repoPath}.

Also update the documentation.

See gh-334 and the class documentation for additional details.

PR #664, opened 2017-03-08
Add missing test for credential provider.

dlaidlaw:
The credential provider for AWS CodeCommit cannot be executed without the change in the setCredentialsProvider method of JGitEnvironmentRepository.java. This test got dropped in the original pull request. The original pull request was #548


spring-cloud-config-client/.gitignore
```diff
@@ -0,0 +1 @@
+/.apt_generated/
```
ryanjbaxter: Is this needed?

dlaidlaw: It was generated somewhere. I did not add it. I googled and it says that that is where code generated by Annotations is placed. And I do believe that there is code generated by Lombok in there. So it is probably a good idea, but I can remove it if that is preferable. There are a couple of .apt_generated folders generated by the build on my machine.

ryanjbaxter: Hum, it seems odd because nothing has changed.  I think it should be removed.

dlaidlaw: I removed it, and added some tests to improve code coverage, and test that the change actually performed as expected.

Issue #584, opened 2016-12-12
Config Server cannot pull from remote repo


magiccrafter:
Hi,

After almost 3 months without any issues the cloud config server suddenly failed pulling from the repo.
This happened silently on Saturday and on Monday we found out after checking the logs that the new config changes were not being picked from the config-server. 

Exception thrown: `Cannot pull from remote https://repo.git, the working tree is not clean.`

No changes were pushed to the repo or new version of cloud server deployed anytime close to the silent crash. The good thing is that the issue can be resolved just by restarting the cloud config server(s). After that everything goes back to normal but now the question is how to monitor and get notified for such issues. It would be nice if there is a HealthIndicator reporting the health of the config server taking into consideration such state or/and the logging level changed to ERROR.

Detailed exception:
```
2016-12-10 03:47:09,118 [http-nio-8888-exec-5] INFO  o.s.c.a.AnnotationConfigApplicationContext Refreshing org.springframework.context.annotation.AnnotationConfigApplicationContext@498613ae: startup date [Sat Dec 10 03:47:09 CET 2016]; root of context hierarchy
2016-12-10 03:47:09,123 [http-nio-8888-exec-5] INFO  o.s.c.c.s.e.NativeEnvironmentRepository Adding property source: file:/tmp/config-repo-7239745589329148613/application.properties
2016-12-10 03:47:09,123 [http-nio-8888-exec-5] INFO  o.s.c.a.AnnotationConfigApplicationContext Closing org.springframework.context.annotation.AnnotationConfigApplicationContext@498613ae: startup date [Sat Dec 10 03:47:09 CET 2016]; root of context hierarchy
2016-12-10 03:48:08,949 [http-nio-8888-exec-10] INFO  o.s.c.c.s.e.MultipleJGitEnvironmentRepository Cannot pull from remote https://repo.git, the working tree is not clean.
2016-12-10 03:48:08,983 [http-nio-8888-exec-10] INFO  o.s.c.a.AnnotationConfigApplicationContext Refreshing org.springframework.context.annotation.AnnotationConfigApplicationContext@11af7f6b: startup date [Sat Dec 10 03:48:08 CET 2016]; root of context hierarchy
2016-12-10 03:48:08,988 [http-nio-8888-exec-10] INFO  o.s.c.c.s.e.NativeEnvironmentRepository Adding property source: file:/tmp/config-repo-7239745589329148613/application.properties
[...]

PR #677, opened 2017-03-30
Changed Health-Endpoint-Status if git-scm not available and extend logging-settings to set level to error if repo not available

marbon87:
- Added property spring.cloud.config.server.git.error-log-on-failing-fetch
  to set logging type to error if fetch failed

- Updated Health-Status to set to out-of-service if git repo is currenty
  not fetchable

- Created new Variable 'cached' for Class Environment

Fixes gh-584

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
@@ -46,6 +46,12 @@ public class Environment {
 
 	private String state;
 
+	/**
+	 * Marks an environment as cached means that the last request to update the property sources failed
+	 * and the last existing information is used instead.
+	 */
+	private Boolean cached = false;
+
 	public Environment(String name, String... profiles) {
 		this(name, profiles, "master", null, null);
 	}
@@ -56,6 +62,7 @@ public class Environment {
 	 */
 	public Environment(Environment env) {
 		this(env.getName(), env.getProfiles(), env.getLabel(), env.getVersion(), env.getState());
+		this.setCached(env.getCached());
 	}
 
 	@JsonCreator
@@ -128,12 +135,21 @@ public class Environment {
 		this.state = state;
 	}
 
+	public Boolean getCached() {
+		return cached;
+	}
+
+	public void setCached(Boolean cached) {
+		this.cached = cached;
+	}
+
 	@Override
 	public String toString() {
 		return "Environment [name=" + name + ", profiles=" + Arrays.asList(profiles)
 				+ ", label=" + label + ", propertySources=" + propertySources
 				+ ", version=" + version
-				+ ", state=" + state + "]";
+				+ ", state=" + state
+				+ ", cached=" + cached + "]";
 	}
 
 }
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerHealthIndicator.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerHealthIndicator.java
@@ -19,6 +19,7 @@ import org.springframework.util.CollectionUtils;
 
 /**
  * @author Spencer Gibb
+ * @author Mark Bonnekessel
  */
 @ConfigurationProperties("spring.cloud.config.server.health")
 public class ConfigServerHealthIndicator extends AbstractHealthIndicator {
@@ -53,6 +54,11 @@ public class ConfigServerHealthIndicator extends AbstractHealthIndicator {
 				HashMap<String, Object> detail = new HashMap<>();
 				detail.put("name", environment.getName());
 				detail.put("label", environment.getLabel());
+
+				if (environment.getCached()) {
+					builder.unknown();
+				}
+
 				if (environment.getProfiles() != null && environment.getProfiles().length > 0) {
 					detail.put("profiles", Arrays.asList(environment.getProfiles()));
 				}
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -38,6 +38,7 @@ import org.eclipse.jgit.api.StatusCommand;
 import org.eclipse.jgit.api.TransportCommand;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.api.errors.RefNotFoundException;
+import org.eclipse.jgit.api.errors.TransportException;
 import org.eclipse.jgit.lib.Ref;
 import org.eclipse.jgit.transport.CredentialsProvider;
 import org.eclipse.jgit.transport.FetchResult;
@@ -66,6 +67,7 @@ import static org.springframework.util.StringUtils.hasText;
  * @author Marcos Barbero
  * @author Daniel Lavoie
  * @author Ryan Lynch
+ * @author Mark Bonnekessel
  */
 public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 		implements EnvironmentRepository, SearchPathLocator, InitializingBean {
@@ -90,7 +92,7 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 	private JGitEnvironmentRepository.JGitFactory gitFactory = new JGitEnvironmentRepository.JGitFactory();
 
 	private String defaultLabel = DEFAULT_LABEL;
-	
+
 	/**
 	 * The credentials provider to use to connect to the Git repository.
 	 */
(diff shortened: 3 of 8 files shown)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
```diff
@@ -70,6 +77,7 @@ public Environment(@JsonProperty("name") String name,
 		this.label = label;
 		this.version = version;
 		this.state = state;
+		this.cached = false;
```
ryanjbaxter: Is there a reason to set the variable in the constructor as well?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -56,8 +59,6 @@
 
 import com.jcraft.jsch.Session;
 
-import static org.springframework.util.StringUtils.hasText;
```
ryanjbaxter: static imports should be imported last

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -170,13 +195,19 @@ public void afterPropertiesSet() throws Exception {
 	/**
 	 * Get the working directory ready.
 	 */
-	private String refresh(String label) {
+	public String refresh(String label) {
```
ryanjbaxter: why are you making this method public, it looks like it is still only used in this class

marbon87:
I Fixed the issues in the branch.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
marbon87: nope, was a accidental duplicate initilization

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
marbon87: fixed

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
marbon87: Fixed. It was a temporary idea to use that method in the health check which was not a good idea :)

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
```diff
@@ -16,13 +16,13 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonProperty;
-
```
ryanjbaxter: Sorry if I missed this before, but please dont change the import order.  You can get the import settings here https://github.com/spring-cloud/spring-cloud-netflix/tree/master/eclipse

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerHealthIndicator.java
```diff
@@ -53,6 +54,11 @@ protected void doHealthCheck(Health.Builder builder) throws Exception {
 				HashMap<String, Object> detail = new HashMap<>();
 				detail.put("name", environment.getName());
 				detail.put("label", environment.getLabel());
+
+				if (environment.getCached()) {
+					builder.outOfService();
```
ryanjbaxter: Should we consider the config server out of service just because the repo couldnt be reached?  I mean technically it can still serve configuration data right?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
marbon87: My mistake. The settings don't work so well with intellij (imports are grouped to .* by intellij) and i forgot to switch back to them. I will fix that!

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerHealthIndicator.java
marbon87: You are right. Regarding to the documentation out-of-service means "Status indicating that the component or subsystem has been taken out of service and should not be used.".

But down and unknown also don't fit well. Maybe unknown. Should we use a custom status like outdated or cached?

ryanjbaxter: Maybe `unknown` but not down.  A custom status might be OK as well.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
ryanjbaxter: There is an Eclipse formatting plugin for IntelliJ that you can use to import Eclipse formatting settings.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerHealthIndicator.java
marbon87: I changed it to unknown

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/environment/Environment.java
marbon87: fixed

marbon87:
I changed the status and imports in my fork, but the build is not triggered again...

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -307,6 +341,17 @@ private FetchResult fetch(Git git, String label) {
 			}
 			return result;
 		}
+		catch (TransportException ex){
+			if(errorLogOnFailingFetch){
```
spencergibb: I don't think I like a branch here. This is too into the weeds for a configuration item. If you're going to log something, pick the one that works and be done.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SearchPathLocator.java
```diff
@@ -35,15 +36,23 @@
 		private final String label;
 		private final String[] locations;
 		private final String version;
+		private Boolean latest;
```
spencergibb: Should be `boolean`.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
marbon87: We could instead change the logging level from WARN to ERROR?

spencergibb: I don't know. Is a `TransportException` thrown if a repository doesn't exist? That is a regular occurrence that isn't an error.

marbon87: The Jgit documentation for the method  fetch says:
> @throws org.eclipse.jgit.api.errors.TransportException 
> when an error occurs during transport

I think that is an error and so we should log it as error

spencergibb: But I'm saying that we try for branches that don't have to exist, therefore aren't necessarily errors.

marbon87:
@spencergibb what did you mean by your last comment? Is there anything that should be changed?

spencergibb:
I didn't mean to approve the changes, I meant to request changes.

marbon87:
> I didn't mean to approve the changes, I meant to request changes.
ahh, ok sorry

we could check if the request branch is the default-branch and only then log it as error?

manderson23:
Any update on this PR and whether it would be accepted if completed? The health indicator reporting down if the git repository cannot be accessed is something that might interest me.

ryanjbaxter:
We will try and discuss this durning our team meeting this week

sashadidukh:
Take a look at the use case described below.

We have `Config Server application` and several `Config Client applications`, some of the `Config Client applications` gets configuration from our `Config Server application` in the JSON format 
(via [labelledJsonProperties](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java#L144) `EnvironmentController` method) and they don't want to convert all `Environment` response by himself (in case of using [labelled](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java#L100) `EnvironmentController` method).
In this case, `Config Client applications` don't know is returned configuration cached or not.
Could we add some information to the Response Header when Git does not available and cached configuration is returned?

For more details see [this](https://github.com/spring-cloud/spring-cloud-config/issues/801) issue.

spencergibb:
After reviewing this PR again, we decided to close it. If the health indicator is not needed you can disable it. I also propose an option that will change the usage of `down()` to `unknown()` rather than changing all environment repositories (this only changes jgit).

PR #687, opened 2017-04-14
Remove unreachable assertNull()

izeye:
`assertNull()` is unreachable, so removing them looks more sensible.


spencergibb:
Yes, the tests did pass.

PR #691, opened 2017-05-08
Add ability to set JGit TransportConfigCallback

rterentiev:
Hey guys,

I need to do some custom configuration for JGit's ```Transport``` object. And JGit provides configuration callback ```TransportConfigCallback``` for this case.

This pull request adds ability to set this ```TransportConfigCallback```.

Thank you,
Roman

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -444,8 +444,13 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 
 	private void configureCommand(TransportCommand<?, ?> command) {
 		command.setTimeout(this.timeout);
-		command.setTransportConfigCallback(this.transportConfigCallback);
-		command.setCredentialsProvider(getCredentialsProvider());
+		if (this.transportConfigCallback != null) {
+			command.setTransportConfigCallback(this.transportConfigCallback);
+		}
+		CredentialsProvider credentialsProvider = getCredentialsProvider();
+		if (credentialsProvider != null) {
+			command.setCredentialsProvider(credentialsProvider);
+		}
 	}
 
 	private CredentialsProvider getCredentialsProvider() {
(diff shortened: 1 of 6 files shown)


rterentiev:
I rebased my branch with master and codacy now says that my test should include ```assert() or fail()```.  I use ```assertEquals``` for verifying results. Other tests also use ```assertEquals```. 
Did I miss something or its just problem with codacy?

ryanjbaxter:
I dont think it is complaining about the type of assert you are using, rather it is complaining that you don't have one.  You are verifying certain things in that test so I think it is fine.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -430,20 +442,26 @@ protected void configure(Host hc, Session session) {
...
-		} else if (hasText(getUsername())) {
-			cmd.setCredentialsProvider(
-					new UsernamePasswordCredentialsProvider(getUsername(), getPassword()));
-		} else if (hasText(getPassphrase())) {
-			cmd.setCredentialsProvider(
-					new PassphraseCredentialsProvider(getPassphrase()));
-		}
+	private void configureCommand(TransportCommand<?, ?> command) {
+		command.setTimeout(this.timeout);
+		command.setTransportConfigCallback(this.transportConfigCallback);
```
spencergibb: Is setting null here OK?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -430,20 +442,26 @@ protected void configure(Host hc, Session session) {
...
+		command.setCredentialsProvider(getCredentialsProvider());
```
spencergibb: This needs to be protected by a null check, since that is what was done previously.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
rterentiev: Yeah, TransportCommand does not have callback by default. 
Anyway, I added null check in case something changed later. Thanks.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
rterentiev: Fixed. Thanks!

PR #734, opened 2017-06-30
Feature Request: Entire folder as tar.gz and binary endpoints without…

transamericamoon:
Hi all,

We added these additional features and thought we would give back to the community if you guys are interested.

1. We use LOTS of docker images, most are based on alpine linux and only come with busybox version of wget.  This version does not allow us to pass content-type headers.  We write scripts that go directly to configserver to pull down files during application startup, so we added direct endpoints for binary files.  We also had issues pulling down camel XML files (that we did not want placeholders resolved by configserver) because there is no way to pass the content-type header from a spring configuration @ImportResource("URL")

2. Some of our apps we need download an entire configuration folder with nested folders, and expand that in our docker container before the app starts, so we added an endpoint for downloading an entire folder off the config server and returning it as a tar.gz file.  (binary or resolve placeholders).  This save us alot of time because each file request updates the svn/git repo, now we can grab 100 files in a single request.  (It was really slow with svn)

example URLs:

binary: 
http://localhost:8888/binary/app/default/master/subdir/test.txt

tar.gz (binary)
http://localhost:8888/binarytargz/app/default/master/subdir

tar.gz (resolved placeholders)
http://localhost:8888/targz/app/default/master/subdir

These are kinda special use cases so let me know what you think?  (if you guys like it i'll write some UTs)

### DIFF
--- a/spring-cloud-config-server/pom.xml
+++ b/spring-cloud-config-server/pom.xml
@@ -66,6 +66,11 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>1.13</version>
+        </dependency>
 	</dependencies>
 
 	<properties>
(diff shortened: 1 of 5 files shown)


spencergibb:
Hmm. Not sure this is a direction we want to go. At this point, the lines between a configuration server and a object storage system (AWS S3 or GCP Storage) are starting a blur.

spencergibb:
I think there could be a more elegant way than separate paths, maybe the accept header and a maybe a parameter for the 3rd.

spencergibb:
Sorry for the delay. The team agrees that it feels like we're headed down the object/blob storage API. You should be able to add this as a `@Controller` in config server. If we're missing hooks to make this possible let us know.

transamericamoon:
No prob.  I will close it.

spencergibb:
Sorry, I meant to hit close and didn't ;-)

Issue #566, opened 2016-11-24
Config Server pulling repo in each request


fjvieira:
In the Config Server, every time it tries to fullfil a GET request for a GIT repository, it calls refresh method of JGitEnvironmentRepository, that results in a PULL to GIT repo, what takes a lot of time and can generate issues in case of many concurrent requests. Maybe it should be better to have a timeout threshold and a scheduled task to do it.

rlynch:
At my company we have a similar problem and are working on a PR which will use messaging to proactively update the locally cloned repos on a git update and almost entirely eliminate the need for a pull request during a confit get request. We want to do some internal testing and hardening before submitting it here but look for it soon.  We are not committers so there is no guarantee that it will ever be accepted in SCCS so don't run your business counting on this but we will have something for the community to review in the near future.

spencergibb:
@rlynch how different is that from the monitor functionality?

fjvieira:
@spencergibb The monitor functionality is nice. I adapted the monitor solution to use SQS/SNS instead of Cloud Bus and it works very well. My point is that should be a way of overpassing what I described. Nowadays the average response time is more than 1 second. And considering that health check of each Config Client do a GET to print the list of resources, it is an eternity. I disabled this health check in my clients and the response time of /health now is less than 20ms.

rlynch:
Not much, only difference is it updates the configuration servers instead of the clients. Then we can "mostly" remove the call to git during the findOne method.  This also allows us to reduce the locking on the findOne method which really hindered scalability for us. It's coded up but still needs to be tested so it only works in unit test land at this point.

marcellodesales:
@fjvieira that's definitely a great finding for sure! We were asked to investigate the health check as well, but we considered not using it... I've been working with @rlynch and his patch should help help with your requirement 👍

kuipercm:
@rlynch  (How) Can we help? This is a feature that would be most welcome in our environment as well. The load and response time caused by the git pull is too much at the moment, so avoiding git pulls would be very helpful.

asarkar:
I've a different thought about this. If we could add caching to config server then the constant Git pull can be reduced. This would have been real easy to do using Spring AOP is `JGitEnvironmentRepository..refresh` wasn't a private method. AFAIK, Spring AOP can't proxy private methods. Perhaps a small PR can change that.
[...]

PR #749, opened 2017-07-21
Feature request: Adding refresh rate to GIT repositories

karl-ravn:
To avoid overload on GIT backend servers, I have added a RefreshRate setting to the git repositories. This is a naive implementation, which in case of an error will wait another refreshRate until checking again. With this, the configuration server can handle much more load in case of the fact that there are multiple servers asking for updated configuration (or, as in our case, very fine grained configuration from hundreds of files).

The code is completely passive unless activated by configuring the git backend with a "refreshRate" property (number of seconds as value). It is placed next to the "timeout" property, so the same value will be used, just as the timeout.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -82,6 +82,16 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 	private int timeout = 5;
 
 	/**
+	 * Time (in seconds) between refresh of the git repository
+	 */
+	private int refreshRate = 0;
+
+	/**
+	 * Time of the last refresh of the git repository
+	 */
+	private long lastRefresh;
+
+	/**
 	 * Flag to indicate that the repository should be cloned on startup (not on demand).
 	 * Generally leads to slower startup but faster first query.
 	 */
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@karl-ravn Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=749)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=749) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@karl-ravn Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=749)!

lizell:
This would be super useful, we are having constant problems with a server that is being hit quite frequently, due to a lot of different profiles and settings.

karl-ravn:
Wouldn't this PR solve the purpose for this old issue? https://github.com/spring-cloud/spring-cloud-config/issues/566

p-zalejko:
any chance to push this topic further?

karl-ravn:
The interest seems to be a little bit low. We have in the mean-time developed a different solution where we jack into the Spring ApplicationEnvironmentPreparedEvent and performs a double-clone of the repository and rewrites the config server repo to be a local repo instead. The double-clone is to make sure that the repo that Config Server is looking at is only doing refresh against the other local repo on each invocation. The repository that config server is not checking out itself, is indeed a remote repo and refreshed every X seconds.

We had a drop in response times from ~250ms on a small repo to ~80ms with the other fix. It is more complex code however, and the porting is not equally easy.

ogrodnek:
I'd love to see this merged.  What else needs to be done?

karl-ravn:
🤷‍♂️Noone of the maintainers seem to care about what some users want. 

We are moving away from this project since it takes forever to get anything through. 8 months and still “team discussion”. Frickin abandonware!

spencergibb:
Merged via 17a2cadabba5543c125103eeed921e0d558093e6

pitinga:
Hello...

This feature was released in version 1.4.4?

I did not found in documentation how to set the value of "refreshRate".

How do I do this?

spencergibb:
It is in version 2.0.0

pitinga:
Hi @spencergibb 

And the feature request #1001? If you see above it was merged in 1.4.x on May.

PR #760, opened 2017-07-27
Fix symmetric encryption

ojhughes:
Setting property `encrypt.key` to enable symmetric encryption does not work since Dalston SR2.  This can be verified using the endpoint `/encrypt/status` Fixed by creating a default `TextEncryptorLocator` when the '`encrypt.key` is set

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/SymmetricKeyEncryptor.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/SymmetricKeyEncryptor.java
@@ -1,0 +1,29 @@
+package org.springframework.cloud.config.server.encryption;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.crypto.encrypt.TextEncryptor;
+
+/**
+ * Provide a default {@link TextEncryptorLocator} when a symmetric key is configured
+ *
+ * @author Ollie Hughes
+ *
+*/
+@Configuration
+@ConditionalOnProperty(value = "encrypt.key", matchIfMissing = false)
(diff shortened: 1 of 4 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/SymmetricKeyEncryptor.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @author Ollie Hughes
+ *
+*/
+@Configuration
+@ConditionalOnProperty(value = "encrypt.key", matchIfMissing = false)
+public class SymmetricKeyEncryptor {
+
+	@Autowired
+	private TextEncryptor encryptor;
```
dsyer: The class name `SymmetricKeyEncryptor` implies it is something to do with symmetric keys, but the `@Autowired` will succeed for any `TextEncryptor`. I'm kind of worried that this is duplicating a code path in spring-cloud-commons somewhere.

dsyer:
I see why this is needed now. But for consistency can we put the new config class in `EncryptionAutoConfiguration` with all the others (assuming that it's still possible to meet the conditional constraints)?

ojhughes:
@dsyer I originally had it in`EncryptionAutoConfiguration` but there was a circular reference with https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EncryptionAutoConfiguration.java#L81. Moving it to a class not loaded in the bootstrap context resolved it.

Using a Qualifier could be an option but felt that might preclude how other clients might be providing a `TextEncryptorLocator`

dsyer:
Yeah, we'd need to add another condition to prevent the cycle. We need it anyway because if the user provides both `encrypt.key` and `encrypt.key-store.location` then we'll end up with 2 `TextEncryptorLocators` won't we?

ojhughes:
should having both `encrypt.key` and `encrypt.keystore` be disallowed as there would be no way to choose which encryption method to use?

dsyer:
Yes, I guess it would make sense to fail in a helpful way (rather than just a no-unique bean exception). That's a nice to have.

dsyer:
I took your test and organized the config files a bit differently in fe42a1d. Thanks.

PR #774, opened 2017-08-15
Fix encryption of string containing curly braces

mnyeste:
Steps to reproduce:

- ecrypt a string with a single closing curly brace, e.g. "ab}cd"
- decrypt it
- the expected result would be the same string, however the result is "cd"

The root cause is the  `stripPrefix` method of `EnvironmentPrefixHelper` class called before the encryption which is currently removing all characters until the first closing curly brace including it. 

It seems the intent was to remove all `{foo:bar}` like expressions from the beginnig of the string, fix is looking for that pattern. Note, the implementation won't allow to enrypt texts starting with {foo:bar} sequence, where foo is not a prefix keyword. To make the implementation perfect all the known keyword:value pairs should be removed one by one.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EnvironmentPrefixHelper.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EnvironmentPrefixHelper.java
@@ -117,7 +117,7 @@ class EnvironmentPrefixHelper {
 		if (value.contains(ESCAPE)) {
 			return value.substring(value.indexOf(ESCAPE) + ESCAPE.length());
 		}
-		return value.substring(value.lastIndexOf("}") + 1);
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@mnyeste Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=774)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=774) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mnyeste Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=774)!

PR #775, opened 2017-08-15
Added additional options for GIT SSH configuration property configuration

yasmans:
- Added possibility to specify a custom known_hosts file which should fix Issue 101.
_spring.cloud.config.server.git.knownHostsFile_
- Added possibility to specify the order of SSH authentication methods. This will resolve problems when SSH server has high priority for interactive authentication methods and application presents a login-prompt. 
_spring.cloud.config.server.git.preferredAuthentications_


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/PreferredAuthenticationsValidator.java
```diff
@@ -0,0 +1,57 @@
...
+ * JSR-303 Cross Field validator that ensures that a {@link SshUriProperties} bean for the constraints:
+ * - Verifies that PreferredAuthentications property is properly formed
+ *
+ * Beans annotated with {@link PreferredAuthenticationsIsValid} and {@link Validated} will have the constraints applied.
+ *
+ * @author Edgars Jasmans
+ */
+public class PreferredAuthenticationsValidator implements ConstraintValidator<PreferredAuthenticationsIsValid, SshUriProperties> {
+
+    private static final String propertyPattern = "([\\w -]+,)*([\\w -]+)";
```
spencergibb: Couldn't a `javax.validation.constraints.Pattern` be used rather than a custom annotation?

yasmans: Yes, you are right - I could put _Pattern_ constraint on _SshUri.preferredAuthentications_ property. It's just that all other validation constraints are on subclass _SshUriProperties_ and I wanted them to be in one place. 
Maybe it is an overkill. Should I use _Pattern_?

PR #826, opened 2017-10-23
Add support for default label on native environment repository

daniellavoie:
Brings support for `spring.cloud.config.server.defaultLabel` on the `NativeEnvironmentRepository`.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
@@ -49,13 +49,15 @@ import org.springframework.util.StringUtils;
  * @author Dave Syer
  * @author Roy Clarkson
  * @author Venil Noronha
+ * @author Daniel Lavoie
  */
 @ConfigurationProperties("spring.cloud.config.server.native")
-public class NativeEnvironmentRepository implements EnvironmentRepository, SearchPathLocator, Ordered {
+public class NativeEnvironmentRepository
+		implements EnvironmentRepository, SearchPathLocator, Ordered {
 
 	private static Log logger = LogFactory.getLog(NativeEnvironmentRepository.class);
-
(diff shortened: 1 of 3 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
```diff
@@ -49,13 +49,15 @@
...
 @ConfigurationProperties("spring.cloud.config.server.native")
-public class NativeEnvironmentRepository implements EnvironmentRepository, SearchPathLocator, Ordered {
+public class NativeEnvironmentRepository
+		implements EnvironmentRepository, SearchPathLocator, Ordered {
 
 	private static Log logger = LogFactory.getLog(NativeEnvironmentRepository.class);
-
-	private static final String DEFAULT_LABEL = "master";
+	
+	private String defaultLabel;
```
spencergibb: I think this still needs to default to `master`. If a user creates a `NativeEnvironmentRepository` bean, they wouldn't know they might need to update this.

daniellavoie: `DEFAULT_LABEL` was never used. Exposed through `getDefaultLabel` but never used.

spencergibb: right. `private String defaultLabel = "master"` is all I'm saying.

daniellavoie: In fact, the `profile` parameter will only be used if the users specify a custom location with `{label}` placeholder.

Updating to default to master.

daniellavoie: I guess if anyone relied on calling the public `getDefaultLabel()`, it's better to not break it. Fair enough.

PR #828, opened 2017-10-26
Support for plain text resource without extension

daniellavoie:
Current mapping definition would cause the wrong environment mapping to be resolved rather the expected resource controller.


spencergibb:

PR #833, opened 2017-10-31
Extract RequestTimeout property in `ConfigServicePropertySourceLocator`

tristanchanson:
Extract RequestTimeout property in `ConfigServicePropertySourceLocator` to allow for configuration of the timeout

- Extract RequestTimeout Property
- Refactor to remove call to deprecated method
- Move header name constants to use `HttpHeaders` constants

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
@@ -16,11 +16,6 @@
 
 package org.springframework.cloud.config.client;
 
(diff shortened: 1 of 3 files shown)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
```diff
@@ -134,6 +140,14 @@ public void setUri(String url) {
 		this.uri = url;
 	}
 
+	public int getRequestTimeout() {
```
ryanjbaxter: indentation

indraneelb1903:
As this has already been implemented , should this be closed?

PR #840, opened 2017-11-06
Make CompositeConfiguration conditional on missing bean

shakuzen:
Previously, `CompositeConfiguration` (which is imported by `ConfigServerAutoConfiguration`) would unconditionally create a `CompositeEnvironmentRepository` bean marked as `@Primary`. This makes it difficult for users, in their code, to provide their own `CompositeEnvironmentRepository` bean to be used in various places. This adds a `ConditionalOnMissingBean` condition to the `Configuration` to make it easy to provide a primary `CompositeEnvironmentRepository` in user code.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/CompositeConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/CompositeConfiguration.java
@@ -32,6 +32,7 @@ import org.springframework.context.annotation.Primary;
  * @author Ryan Baxter
  */
 @Configuration
+@ConditionalOnMissingBean(CompositeEnvironmentRepository.class)
 public class CompositeConfiguration {
 
 	private List<EnvironmentRepository> environmentRepos = new ArrayList<>();
--- a/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepositoryTests.java
+++ b/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepositoryTests.java
@@ -16,10 +16,17 @@
 package org.springframework.cloud.config.server.environment;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import org.junit.Test;
 import org.springframework.cloud.config.environment.Environment;
 import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.cloud.config.server.config.CompositeConfiguration;
+import org.springframework.cloud.config.server.config.ConfigServerHealthIndicator;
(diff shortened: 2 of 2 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/CompositeConfiguration.java
```diff
@@ -32,6 +32,7 @@
  * @author Ryan Baxter
  */
 @Configuration
+@ConditionalOnMissingBean(CompositeEnvironmentRepository.class)
```
ryanjbaxter: Actually I think the problem is below on the `compositeEncironmentRepository` method.  Looks like I added the wrong conditional class there.

shakuzen: Both bean methods in this class produce `CompositeEnvironmentRepository` beans (it just happens one is the more specific type `SearchPathCompositeEnvironmentRepository`). If the condition were moved to the `compositeEnvironmentRepository()` method, then users would still be unable to provide a primary `CompositeEnvironmentRepository` if the `SearchPathCompositeEnvironmentRepository` bean is made.

If you change the test configuration `OverrideCompositeConfig` to the following, you can see that happen (because the `MultipleJGitEnvironmentRepository` bean is a `SearchPathLocator` bean):

```java
@Bean
@Primary
CompositeEnvironmentRepository customCompositeEnvironmentRepository() {
	return new CompositeEnvironmentRepository(Arrays.<EnvironmentRepository>asList(new TestOrderedEnvironmentRepository(1, new Environment("app", "dev"), null), someGitRepo()));
}

@Bean
MultipleJGitEnvironmentRepository someGitRepo() {
	MultipleJGitEnvironmentRepository gitRepo = new MultipleJGitEnvironmentRepository(new MockEnvironment());
	gitRepo.setUri("some-uri");
	return gitRepo;
}
```

Without guarding beans marked primary with some `ConditionalOnMissingBean`, the only option for users wanting a different primary bean is excluding the configuration that creates them, as far as I know. In this case, that means excluding the `ConfigServerAutoConfiguration`.

ryanjbaxter: Ah yes, I forgot they were subclasses of each other makes sense.

shakuzen:
I guess I should have opened this against the `1.3.x` branch. Should I open another pull request, or will you take care of merging to 1.3.x (assuming this qualifies as a fix for that line) and 2.0.x?

ryanjbaxter:
NP I just cherry picked the commit https://github.com/spring-cloud/spring-cloud-config/commit/b5022e3660516e4fd7c5f3f9d4ff7e1bdcb3a3b2

Issue #866, opened 2017-12-05
Add support for propery source origin introduced in boot 2


spencergibb:
This would be helpful in debugging configuration problems.

@dsyer @twicksell

spencergibb:
Boot's env endpoint changed for a simple key -> value, to key -> {value, origin}. Config server would need to be backward compatible. Two options, either a new prefix (v2) or a content type (which is what boot uses). I think I lean towards the content type.

Boot's content type `application/vnd.spring-boot.actuator.v2+json`, so I would propose `application/vnd.spring-cloud.config-server.v2+json`.

Example output:
```
// 20171206103612
// http://localhost:8888/foo/default

{
  "name": "foo",
  "profiles": [
    "default"
  ],
  "label": null,
  "version": "a611374438e75aa1b9808908c57833480944e1a8",
  "state": null,
  "propertySources": [
    {
      "name": "https://github.com/spring-cloud-samples/config-repo/foo.properties",
      "source": {
        "foo": {
          "value": "from foo props",
          "origin": "URL [file:/tmp/config-repo-7662911556124581061/foo.properties]:1:6"
        },
        "democonfigclient.message": {
          "value": "hello spring io",
          "origin": "URL [file:/tmp/config-repo-7662911556124581061/foo.properties]:2:27"
        }
      }
    },
    {
      "name": "https://github.com/spring-cloud-samples/config-repo/application.yml",
      "source": {
        "info.description": {
          "value": "Spring Cloud Samples",
          "origin": "URL [file:/tmp/config-repo-7662911556124581061/application.yml]:2:16"
        },
        "info.url": {
          "value": "https://github.com/spring-cloud-samples",
          "origin": "URL [file:/tmp/config-repo-7662911556124581061/application.yml]:3:8"
        },
        "eureka.client.serviceUrl.defaultZone": {
          "value": "http://localhost:8761/eureka/",
          "origin": "URL [file:/tmp/config-repo-7662911556124581061/application.yml]:7:20"
        },
        "foo": {
          "value": "baz",
          "origin": "URL [file:/tmp/config-repo-7662911556124581061/application.yml]:9:6"
        }
      }
    }
  ]
}
```

spencergibb:
We could enhance the origin depending on the environment. For example, git blame information could be added, etc...

spencergibb:
From `/actuator/env`
```
{
  "activeProfiles": [],
  "propertySources": [
    {
[...]

PR #870, opened 2017-12-15
Adds support for property origin tracking in config server/client.

spencergibb:
Adds a new format if requested using v2 Accept header, otherwise the
format is backwards compatible so older clients will work with new
servers.

fixes gh-866

/cc @twicksell

requires https://github.com/spring-cloud/spring-cloud-commons/pull/583

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -110,9 +110,11 @@ public class ConfigServicePropertySourceLocator implements PropertySourceLocator
 																// xml
 						for (PropertySource source : result.getPropertySources()) {
 							@SuppressWarnings("unchecked")
-							Map<String, Object> map = translateOrigins((Map<String, Object>) source.getSource());
-							composite.addPropertySource(new OriginTrackedMapPropertySource(source
-									.getName(), map));
+							Map<String, Object> map = translateOrigins(
+									(Map<String, Object>) source.getSource());
+							composite.addPropertySource(
+									new OriginTrackedMapPropertySource(source.getName(),
+											map));
 						}
 					}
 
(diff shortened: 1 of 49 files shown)


dsyer:
It seems like we need this feature, and this looks like a good start, and we clearly have all the data we need. I don't like the format of the JSON though, and it seems like we could maybe even make a more incremental change that didn't require a new media type (or maybe not, but it's worth thinking about). Repeating the file location/name in every property value seems wasteful, and we could use integer fields for the line and column values, to make them more machine readable, and then convert to canonical "Boot" style format in the client.

spencergibb:
I agree we could probably not repeat the location, but I don't think we can go straight to integers. Ints would work for the file based env repos, like git, but wouldn't work for jdbc or vault. Also, git blame info could go on each origin for git.

spencergibb:
@dsyer @ryanjbaxter @OlgaMaciaszek @marcingrzejszczak @TYsewyn I'm reviving this again.

Two options I can see for the format:

1) file not repeated
```
{
    "label": null,
    "name": "foo",
    "profiles": [
        "cloud"
    ],
    "propertySources": [
        {
            "name": "https://github.com/spring-cloud-samples/config-repo/application.yml (document #1)",
            "source": {
                "application.domain": {
                    "origin": "27:11",
                    "value": "${APPLICATION_DOMAIN:cfapps.io}"
                },
                "endpoints.restart": {
                    "origin": "29:12",
                    "value": "enabled"
                }
         }
    ]
}
```

or repeated
```
{
    "label": null,
    "name": "foo",
    "profiles": [
        "cloud"
    ],
    "propertySources": [
        {
            "name": "https://github.com/spring-cloud-samples/config-repo/application.yml (document #1)",
            "source": {
                "application.domain": {
                    "origin": "[https://github.com/spring-cloud-samples/config-repo/application.yml]:27:11",
                    "value": "${APPLICATION_DOMAIN:cfapps.io}"
                },
                "endpoints.restart": {
                    "origin": "[https://github.com/spring-cloud-samples/config-repo/application.yml]:29:12",
                    "value": "enabled"
                }
         }
    ]
}
```

spencergibb:
Once we get the paired down one (I think I prefer it), we could provide different fields for origin, line and column for origins that support it and an opaque string for others.

ryanjbaxter:
+1 to the paired down version

TYsewyn:
I agree on the paired down version, only question that I have now is whether or not we still need to have that ` (document #1)` suffix in the property source its name since we now have the exact location of the property within that file.

spencergibb:
That comes from spring boot actually

spencergibb:
Here's a sample client error message
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Binding to target org.springframework.boot.context.properties.bind.BindException: Failed to bind properties under 'democonfigclient' to com.example.democonfigclient.DemoconfigclientApplication$Config failed:

    Property: democonfigclient.message
    Value: hello from dev profile
    Origin: https://github.com/spring-cloud-samples/config-repo/foo-dev.yml:3:27
    Reason: size must be between 0 and 10
```

PR #875, opened 2017-12-21
Also check if the branch is ahead of origin and if so reset the local branch

transamericamoon:
use case:

Cloud config server is running.

Junior programmer checked in a password to git in clear text

Supervisor reverts the change and force pushes to remove it from history:
git reset --hard HEAD~
git push -f

Configserver is accessed and does a pull, but still shows password because the repo is clean but it is now ahead of origin by one commit.


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -492,10 +492,13 @@ private CredentialsProvider getCredentialsProvider() {
 		return null;
 	}
 
-	private boolean isClean(Git git) {
+	private boolean isClean(Git git, String label) {
 		StatusCommand status = git.status();
 		try {
-			return status.call().isClean();
+            boolean isBranchAhead = false;
```
spencergibb: We use tabs, not spaces

transamericamoon:
Should be good now.  tabs... :(

Issue #877, opened 2017-12-28
Git backend: deleted "feature" branches not skipped during properties lookup


vasilievip:
According to doc:
https://github.com/spring-cloud/spring-cloud-config/blob/master/docs/src/main/asciidoc/spring-cloud-config.adoc#locating-remote-configuration-resources

its possible to specify on client:
```
spring.cloud.config.label=myfeature,develop
```

This leads to ability to deploy service using feature branch and then merge feature branch to master with removal of feature branch, leaving spring.cloud.config.label pointing to missing branch. 
Local repo on config server still has deleted branch and serves settings from this branch which no longer exists. 
Now, assumption is that deletion of branch is OK and if branch is missing - config server will fallback to master and spring.cloud.config.label can contain missing branch. 

There issue is that branch is deleted only in main git repo and if master branch gets updated, w/o restart of config server - changes are not visible by services which have deleted feature branch in startup options.

ryanjbaxter:
So this sounds like if you are going to start merging and deleting branches you will need to restart the config server?  I think this is something we can just document.

vasilievip:
@ryanjbaxter if you do dozens or hundreds of deployments per day using feature branches, keeping those branches is non practical. Having dozens or hundreds of restarts for config service per day just to flush out stale branches seems not practical as well (from automation point of view either)
I would not say that this is documentation only concern.

tdanylchuk:
As this feature is closed and will be available only in  **2.0.0.M9 milestone** is there any possibility to cherry-pick this feature to 1.4.3 and make release, since we would happy to see this feature on prod w\o using M* version.

PS: I've already cherry-picket it already on my repo and tested in 1.4.3 version, everything seems to be smooth.

ryanjbaxter:
@tdanylchuk IDK if it applies so cleanly.  Looks like there are a number of classes that need to be moved over in order to cherry pick this PR even compile.  The major problem I see is that one of the tests appears to be dependent on a JGit API that is only available in JGit 4.* and that required Java 1.8 which we cant swing in the 1.4.x branch.  Maybe the test can be modified, I will have to dig a bit deeper, unless you already did this in your repo and can make a suggestion.

tdanylchuk:
@ryanjbaxter could you check please https://github.com/spring-cloud/spring-cloud-config/pull/947 ?

PR #947, opened 2018-03-21
1.4.x delete untracked branches

tdanylchuk:
cherry-picked from https://github.com/spring-cloud/spring-cloud-config/commit/c59de61394dc4f9f37b0153af9b6bf931ab0a3ab

issue - https://github.com/spring-cloud/spring-cloud-config/issues/877

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -20,6 +20,9 @@ import static org.springframework.util.StringUtils.hasText;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -27,6 +30,7 @@ import java.util.Set;
 import org.eclipse.jgit.api.CheckoutCommand;
 import org.eclipse.jgit.api.CloneCommand;
 import org.eclipse.jgit.api.CreateBranchCommand.SetupUpstreamMode;
+import org.eclipse.jgit.api.DeleteBranchCommand;
 import org.eclipse.jgit.api.FetchCommand;
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.ListBranchCommand;
@@ -47,8 +51,10 @@ import org.eclipse.jgit.transport.CredentialsProvider;
 import org.eclipse.jgit.transport.FetchResult;
 import org.eclipse.jgit.transport.JschConfigSessionFactory;
 import org.eclipse.jgit.transport.OpenSshConfig.Host;
+import org.eclipse.jgit.transport.ReceiveCommand;
 import org.eclipse.jgit.transport.SshSessionFactory;
 import org.eclipse.jgit.transport.TagOpt;
+import org.eclipse.jgit.transport.TrackingRefUpdate;
 import org.eclipse.jgit.transport.UsernamePasswordCredentialsProvider;
 import org.eclipse.jgit.util.FileUtils;
 import org.springframework.beans.factory.InitializingBean;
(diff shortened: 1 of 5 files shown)


spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepositoryTests.java
```diff
@@ -716,6 +727,111 @@ public void shouldSetTransportConfigCallbackOnCloneAndFetch() throws Exception {
...
+		MergeResult mergeResult = mock(MergeResult.class);
+		MergeResult.MergeStatus mergeStatus = mock(MergeResult.MergeStatus.class);
+		MergeCommand mergeCommand = mock(MergeCommand.class);
+		when(git.merge()).thenReturn(mergeCommand);
+		when(mergeCommand.call()).thenReturn(mergeResult);
+		when(mergeResult.getMergeStatus()).thenReturn(mergeStatus);
+		when(mergeStatus.isSuccessful()).thenReturn(true);
+
+		// refresh()->return
+		// git.getRepository().findRef("HEAD").getObjectId().getName();
```
ryanjbaxter: You commented this out because it wont compile right?
I think it should be changed to `when(repository.getRef(anyString())).thenReturn(headRef);`
I am surprised this test even passed without it....so that is quite concerning

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -161,6 +177,14 @@ public void setForcePull(boolean forcePull) {
 		this.forcePull = forcePull;
 	}
 
+	public boolean isDeleteUntrackedBranches() {
+		return deleteUntrackedBranches;
```
ryanjbaxter: Have you tested that this actually gets set when you set the property to `true`?  In `master` there is a configuration properties object and `deleteUntrackedBranches` is set in the constructor.  I am concerned that this is not the case here.  Maybe we should add an integration test?

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepositoryTests.java
tdanylchuk: actually this expectation is redundant, just take a look a bit lower and you will see actual mock
```when(repository.getRef(anyString())).thenReturn(headRef);``` 
As for test and how it passes - it's a big trouble of this class, since it's huge and getting bigger, so mock&verify correctly is very hard.

File: spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java (hunk 2 above)
Comment (tdanylchuk) [reply spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java]: As for flow of setting this property:
It's a bit tricky, since whole ```JGitEnvironmentRepository``` is ```ConfigurationProperties```, for this you can take a look on ```MultipleJGitEnvironmentRepository``` which is inheriting this class, so properties set by ```ConfigurationPropertiesBindingPostProcessor```.
As for tests:
you can take a look on PR class ```JGitEnvironmentRepositoryIntegrationTests```. I've added two tests: ```testShouldReturnEnvironmentFromLocalBranchInCaseRemoteDeleted``` with unset property and exact the same test but with set property - ```testShouldFailIfRemoteBranchWasDeleted```

File: spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepositoryIntegrationTests.java [hunk 3]
```diff
@@ -503,6 +504,42 @@ public void shouldSetTransportConfigCallback() throws IOException {
...
+				.prepareClonedGitRepository(TestConfiguration.class);
+
+		String branchToDelete = "branchToDelete";
+		testData.getServerGit().getGit().branchCreate().setName(branchToDelete).call();
+
+		Environment environment = testData.getRepository().findOne("bar", "staging", "branchToDelete");
+		assertNotNull(environment);
+
+		testData.getServerGit().getGit().branchDelete().setBranchNames(branchToDelete).call();
+		testData.getRepository().findOne("bar", "staging", "branchToDelete");
```
ryanjbaxter: Shouldn't the result be set to `environment`?

tdanylchuk: Nice catch! Sorry for that - fixed now.
In any case main point is not to get ```NoSuchLabelException```

tdanylchuk:
Thank you for accepting this feature.
Can I expect 1.4.3 release in nearest future? ;D

ryanjbaxter:
Yup soon

PR #952, opened 2018-03-27
Refactoring of JGitEnvironmentRepository.java class:

tdanylchuk:
- split into several classes
- moved git related classes into dedicated package
- refactored JGitEnvironmentRepositoryTests.java: left & enriched unit-test + moved integration tests to JGitEnvironmentRepositoryIntegrationTests.java

PS: this PR brings zero new functionality. It's just a minor step to decrease amount of logic in this class. 
This change should lead to better unit-test coverage and simplicity of changing & adding new tests.


tdanylchuk:
One thing that I wanted to add, that this change should be one of the several in order to make code simpler and easier to add new features. So I guess (IMHO) that   the time has come - to refactor ;)

spencergibb:
Unfortunately, we missed the boat for this to add breaking changes in Finchley.

PR #954, opened 2018-03-28
Fix delete untracked branch feature if git repo unavailable

tdanylchuk:
Root cause:
If git repository is not available - fetch command throws exception and null is returned instead of FetchResult object.
In this case enabled "delete untracked branch feature" will brake flow of retrieving local copy.
Resolution summary:
Added null check for fetch result.

PS: sorry, I've missed this earlier...


ryanjbaxter:
Could we perhaps make this change against the 1.4.x branch?  We backported your original change to that branch as well

tdanylchuk:
That's for sure, will do PR, right after this will pass all rounds.

ryanjbaxter:
We just need this PR submitted against the 1.4.x branch, and then we can just merge it forward.  No reason to submit 2 PRs.

PR #960, opened 2018-04-02
Send 404 if application not found

Chids-gs:
Added a check to see if PropertySources is empty. If yes, Cloud Config server would return HTTP 404. 
Modified existing JUnit Testcases to send PropertySources in Environment
Fixes gh-155

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -105,7 +105,7 @@ If there are profile-specific YAML (or properties) files, these are also applied
 Higher precedence translates to a `PropertySource` listed earlier in the `Environment`.
 (These same rules apply in a standalone Spring Boot application.)
 
-You can set spring.cloud.config.server.acceptEmtpy to false so that Server would return a HTTP 404 status, if the application is not found .By default, this flag is set to true.
+You can set spring.cloud.config.server.accept-empty to false so that Server would return a HTTP 404 status, if the application is not found.By default, this flag is set to true.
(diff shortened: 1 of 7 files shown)


spencergibb:
I agree, this needs to be opt-in

Chids-gs:
Have made the implementation config driven (spring.cloud.config.server.acceptEmtpy ). By default this flag would be set to true. When set to false, server would return 404, if application is not found

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -105,6 +105,7 @@ If there are profile-specific YAML (or properties) files, these are also applied
 Higher precedence translates to a `PropertySource` listed earlier in the `Environment`.
 (These same rules apply in a standalone Spring Boot application.)
 
+You can set spring.cloud.config.server.acceptEmtpy to false so that Server would return a HTTP 404 status, if the application is not found .By default, this flag is set to true.
```
ryanjbaxter: @spencergibb should this be `spring.cloud.config.client.acceptEmpty` since you are configuring the client behavior?

spencergibb: I don't think so, this is all server-side behavior. It should be `accept-empty`, BTW.

Chids-gs: Have incorporated the review comment. Also, corrected a typo in the variable name

Chids-gs:
Have incorporated the review comment. Also, corrected a typo in the variable name

PR #982, opened 2018-04-17
#974 remove one or two escapes for supporting JSON

### LINKED ISSUE #974: Escaped variables not properly returned in .json responses
With an escaped variable in a config file, e.g.

```
a=\${hello}
```

requesting the .yml or .properties will properly remove the `\` before returning the config, returning e.g. `a=${hello}`

requesting .json leaves a single `\` (i.e. `{"a": "\$hello"}` (which also ends up being invalid json (invalid escape sequence)).

I believe this is due to the jackson serialization that happens before the property resolution [EnvironmentController.labelledJsonPropertes](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java#L169).

[objectMapper.writeValueAsString](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java#L169) will double up the `\`s to make the resulting JSON valid (`\$` would be an invalid escape), so you end up with `\\${hello}`

[EnvironmentPropertySource.resolvePlaceholders](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/EnvironmentPropertySource.java#L25) only deals with one `\`.

So, .yml/.properties end up with correct/unescaped vars, but .json will return the `\` (and possibly cause downstream errors).

ogrodnek:
Hi.  This is an attempt to fix #974 (cc @ryanjbaxter )

As mentioned previously in the ticket, the original code tries to mask out escaped variables (i.e. `\${hello}`) before performing replacement, and then finally removing the escapes before returning.  

This works fine for .yaml or .properties, but not for JSON.

The jackson serialization that happens before the property resolution ends up adding an extra `\` (so that the JSON is well-formed) [EnvironmentController.labelledJsonPropertes](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java#L169).

I've updated the `resolverPlaceholders` to accept one or two escapes (`\` or `\\`).

I realize this will now treat `\\` as an escape in yml or properties, which may be unintended, but seems unlikely.

An alternative might be to add a flag to this method to indicate whether or not to allow double escape, and updating the `labelledJson` method to pass the flag?

I'm open to any feedback on this approach or another.

Thanks!

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/EnvironmentPropertySource.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/EnvironmentPropertySource.java
@@ -1,6 +1,7 @@
 package org.springframework.cloud.config.server.support;
 
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import org.springframework.cloud.config.environment.Environment;
 import org.springframework.core.env.PropertySource;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@ogrodnek Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=982)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=982) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ogrodnek Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=982)!

PR #989, opened 2018-04-25
Enable skipping ssl validation for vault

pivotal-dylan-roberts:
If skipSslValidation flag is set for a Vault environment repository, configure its RestTemplate accordingly.


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryFactory.java
```diff
@@ -35,8 +45,28 @@ public VaultEnvironmentRepositoryFactory(ObjectProvider<HttpServletRequest> requ
...
+					.loadTrustMaterial(null, (certificate, authType) -> true)
+					.build();
+			CloseableHttpClient httpClient
+					= HttpClients.custom()
+					.setSSLContext(sslContext)
+					.setSSLHostnameVerifier(new NoopHostnameVerifier())
+					.build();
+			HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
+			requestFactory.setHttpClient(httpClient);
+			return new RestTemplate(requestFactory);
```
ojhughes: It would be useful (especially with SCS proxy support) to specify additional `HttpClientFactory` settings.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryFactory.java
```diff
@@ -15,9 +15,19 @@
  */
 package org.springframework.cloud.config.server.environment;
 
+import java.security.KeyManagementException;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import javax.net.ssl.SSLContext;
 import javax.servlet.http.HttpServletRequest;
 
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
```
spencergibb: Is Apache http client a required dependency?

pivotal-dylan-roberts: I believe so. The Spring HttpComponentsClientHttpRequestFactory depends on it and this is where I'm hooking in to configure SSL validation on the RestTemplate.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryFactory.java
```diff
@@ -35,8 +45,28 @@ public VaultEnvironmentRepositoryFactory(ObjectProvider<HttpServletRequest> requ
...
+			SSLContext sslContext = new SSLContextBuilder()
+					.loadTrustMaterial(null, (certificate, authType) -> true)
+					.build();
+			CloseableHttpClient httpClient
+					= HttpClients.custom()
+					.setSSLContext(sslContext)
+					.setSSLHostnameVerifier(new NoopHostnameVerifier())
+					.build();
+			HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
+			requestFactory.setHttpClient(httpClient);
```
pivotal-dylan-roberts: @spencergibb As far as I can tell, this is the recommended way to configure the RestTemplate SSL validation.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryFactory.java
spencergibb: While it is the recommended way, there is no required apache httpclient dependency. It happens to be on the classpath for jgit
```
[INFO] +- org.eclipse.jgit:org.eclipse.jgit:jar:4.8.0.201706111038-r:compile
[INFO] |  +- org.apache.httpcomponents:httpclient:jar:4.5.5:compile
```

https://github.com/spring-projects/spring-framework/blob/master/spring-web/spring-web.gradle#L47

You'll need to move this to auto-configuration with a `@ConditionalOnClass(HttpClient)` (spring-cloud-netflix has to do this) with a setter perhaps on `VaultEnvironmentRepository` to set the `RestTemplate`.

pivotal-dylan-roberts:
@spencergibb I've taken a stab at it. Let me know what you think!

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
```diff
@@ -214,9 +248,10 @@ public SvnEnvironmentRepositoryFactory svnEnvironmentRepositoryFactory(Configura
 	}
 
 	@Bean
-	public VaultEnvironmentRepositoryFactory vaultEnvironmentRepositoryFactory(ObjectProvider<HttpServletRequest> request,
-																						EnvironmentWatch watch) {
-		return new VaultEnvironmentRepositoryFactory(request, watch);
+	public VaultEnvironmentRepositoryFactory vaultEnvironmentRepositoryFactory(
```
ryanjbaxter: I dont understand the need for this `Bean`, what is using it?

ryanjbaxter:
FYI https://github.com/spring-cloud/spring-cloud-commons/pull/360

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
pivotal-dylan-roberts: @ryanjbaxter With the new list-based composite configuration, we're creating bean definitions dynamically in the `CompositeEnvironmentBeanFactoryPostProcessor`. These factories are used in those bean definitions.

ryanjbaxter: OK could we use it above in the `vaultEnvironmentRepository` method as well?  Seems odd that we are recreating the factory there

PR #994, opened 2018-05-02
[fix] discovery data for config server never updates after startup #514

### LINKED ISSUE #514: discovery data for config server never updates after startup
It appears that in discovery-first mode, the config client is not updating its location information for the config server when that data changes in eureka (i.e. when a config server instance is removed or added in eureka). Is this currently expected behavior, or am I doing something wrong?

```
spring:
    cloud:
        config:
            enabled: true
            failFast: true
            discovery:
                enabled: true
eureka:
    client:
        serviceUrl:
            defaultZone: [[my eureka defaultZone]]
```

When I take down the first config server instance, the configserver healthcheck in the app begins to fail despite there being a separate working node. The same occurs if I restart the config server node with a new password in the eureka metadataMap, in which case the client keeps trying to use the old password.

dyc87112:
fix #514


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/RefreshConfigClientPropertiesEvent.java
```diff
@@ -0,0 +1,15 @@
+package org.springframework.cloud.config.client;
```
ryanjbaxter: You need the copyright header here

PR #1000, opened 2018-05-03
[FIX #514] discovery data for config server never updates after startup

### LINKED ISSUE #514: discovery data for config server never updates after startup

dyc87112:
Fix issue #514 : discovery data for config server never updates after startup

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -95,7 +95,10 @@ public class ConfigServicePropertySourceLocator implements PropertySourceLocator
 					logger.warn("No instances find : " + properties.getDiscovery().getServiceId());
 					throw new Exception(message);
 				} else {
-					ServiceInstance instance = discoveryClient.getInstances(properties.getDiscovery().getServiceId()).get(0);
+                    // choose a random instance
+                    int index = (int) (Math.random() * instances.size());
+                    ServiceInstance instance = discoveryClient.getInstances(properties.getDiscovery().getServiceId()).get(index);
+                    // produce url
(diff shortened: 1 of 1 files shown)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -79,8 +86,37 @@ public ConfigServicePropertySourceLocator(ConfigClientProperties defaultProperti
...
+
 		try {
+			if (properties.getDiscovery().isEnabled() && discoveryClient != null) {
+				List<ServiceInstance> instances = discoveryClient.getInstances(properties.getDiscovery().getServiceId());
+				if (instances.size() == 0) {
+					String message = "No instances find : " + properties.getDiscovery().getServiceId();
+					logger.warn("No instances find : " + properties.getDiscovery().getServiceId());
+					throw new Exception(message);
+				} else {
+					ServiceInstance instance = discoveryClient.getInstances(properties.getDiscovery().getServiceId()).get(0);
```
spencergibb: Why always the first one? Maybe it would be better to get a random one?

xiaods:
need testing.

spencergibb:
I think this needs to be reworked so as not to duplicate functionality. @dyc87112 ping me if you want to know what I'm thinking.

Issue #993, opened 2018-04-28
Vault K/V Version 2 Secret Engine


HarryEMartland:
I have just started playing around with Vault backend from what I understand there is a new version of the key value secret engine in vault which has changed the api. This has broken the integration with spring boot config server.

I tried setting `spring.cloud.config.server.vault.backend=secret/data` which pulled back the properties but they were all prefixed with data. I think this is due to the api change having two nested data fields.

Does anyone know a way to get the new version of vault working using the new secret engine?

I am happy to look into this and do a pull request. If so should the current VaultEnvironmentRepository support both versions or should a new Vault2EnvironmentRepository be created?

ryanjbaxter:
@markpollack any ideas?  I was not aware of this.

spencergibb:
@mp911de is the right one. He's been working on things in spring-vault and spring-cloud-vault. I wasn't aware that it wasn't backward compatible.

ryanjbaxter:
Ooops my bad, to many Mark's :)

mp911de:
Basically what @HarryEMartland said: Vault 0.10.0 introduced a versioned key-value backend (k/v backend version 2) that exposes a different API than earlier versions (Vault 0.9.6, k/v backend version 1). There are a couple of aspects that play into this Vault change:

1. The versioned key-value backend requires a `data/` segment between the mount path and the actual context path. So `secret/my-app/foo` (v1) becomes `secret/data/my-app/foo` (v2).
2. The versioned key-valued backend secrets are wrapped in a `data` JSON object. So `{"key": "foo"}` (v1) becomes `{"data": {"key": "foo"}}` (v2).
3. Vault 0.10.0 allows activation of the version 1 (non-versioned) and the version 2 (versioned) k/v backends. By default, a newly spun up Vault server mounts the versioned k/v backend at `secret/` which breaks all existing client implementations.
4. There's no way to discover which version is mounted by the request/response itself. Vault 0.10.0 had support for a request header where all requests were served in the new format. However, Vault 0.10.1 removes this request header/request downgrading (`X-Vault-Kv-Client: v2`) and so we're left with a broken change. 

The native Vault client performs a preflight request to mount metadata (`GET /v1/sys/internal/ui/mount/{path}`, read `options.version` from the JSON response) to use the proper API for key-value backends.

[...]

PR #1018, opened 2018-05-18
Added support for Vault's 0.10.0+ versioning enabled kv engine.

HPacquee:
Vault 0.10.0 introduced a v2 versioning engine of sorts. The Vault
Server mounts the versioning enabled k/v backend at secret/ breaking any
previous interaction with spring-cloud-config's Vault backend integration.
This change provides a config flag and the necessary changed to use either
the v1 (non-versioned) or the v2 (versioned) backend.

Fixes gh-993


docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -546,12 +546,17 @@ The following table describes configurable Vault properties:
 |profileSeparator
 |,
 
+|versioningEnabled
```
mp911de: I think I'd name it along the Vault K/V backend version. Right now, we assume the only change is versioning but we should stay open for the future and not introduce a possible ambiguity. Therefore, let's switch to `kvVersion` (`kvVersion=1`, `kvVersion=2`).

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
```diff
@@ -42,6 +43,10 @@
      */
     private boolean skipSslValidation = false;
     private int order = Ordered.LOWEST_PRECEDENCE;
+    /**
+     * Flag to indicate that the used version of Vault is using the versioning-enabled kv enginge. Defaults to false.
+     */
+    private Boolean versioningEnabled = false;
```
mp911de: Any reason why this is a reference type instead of a primitive?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepository.java
```diff
@@ -291,4 +295,45 @@ public void setRenewable(boolean renewable) {
 			this.renewable = renewable;
 		}
 	}
+
+	static class VersionedVaultResponse extends VaultResponse {
```
mp911de: How about combining URL and response processing in a strategy object? We could have `VaultKvAccessStrategy` that accepts `RestOperations` and a path and returns a `VaultResponse` object. Using strategy objects is a bit more focused by keeping all functionality in a single place for a particular strategy without the need to distribute behavior across multiple types (`VersionedVaultResponse`, `VaultVersioning`, making `data` `protected`).

mp911de:
Thanks a lot for updating the pull request. We should expose as little public API as possible through `VaultKvAccessStrategy` meaning that a single `getData(backend, key)` method should be sufficient. The other aspects (full path, data extraction) should not be visible from outside. 

I will push a commit on your branch and then we can do the merge.

mp911de:
Pushed a polishing commit. This pull request is ready to merge.

HPacquee:
Thanks. I've learned a lot from your `Polishing` commit.

Can I do something to make the codecov check pass?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultKvAccessStrategyFactory.java
```diff
@@ -1,15 +1,93 @@
...
+		}
+
+		@Override
+		public String extractDataFromBody(VaultResponse body) {
+			return body.getData() == null ? null : body.getData().toString();
+		}
+	}
+
+	/**
+	 * Strategy for the key-value backend API version 1.
```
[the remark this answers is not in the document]
mp911de: Good catch, that's fixed now.

mp911de:
Thanks for having another look. The coverage/diff issue arises because I moved `VaultResponse` around which is a lot of code consisting only of getters and setters. I'd leave it as-is for now.

PR #1034, opened 2018-05-31
Made RequestReadTimeout and whether to sendState configurable

indraneelb1903:
Made RequestReadTimeout and whether to send State configurable.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -1121,7 +1121,7 @@ If you use HTTP basic security on your Config Server, it is currently possible t
 
 === Configuring Read Timeouts
 
-If you want to configure read timeout, this can be done by using the property `spring.cloud.config.requestReadTimeout`.
+If you want to configure read timeout, this can be done by using the property `spring.cloud.config.request-read-timeout`.
 
 === Security
 
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -209,7 +209,7 @@ public class ConfigServicePropertySourceLocator implements PropertySourceLocator
 				if (StringUtils.hasText(token)) {
 					headers.add(TOKEN_HEADER, token);
 				}
-				if (StringUtils.hasText(state)) { // TODO: opt in to sending state?
+				if (StringUtils.hasText(state) && properties.isSendState()) {
 					headers.add(STATE_HEADER, state);
 				}
 
@@ -248,8 +248,10 @@ public class ConfigServicePropertySourceLocator implements PropertySourceLocator
 
 	private RestTemplate getSecureRestTemplate(ConfigClientProperties client) {
 		SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
-		requestFactory.setReadTimeout((60 * 1000 * 3) + 5000); // TODO 3m5s, make
-																// configurable?
+		if (client.getRequestReadTimeout() < 0) {
+			throw new IllegalStateException("Invalid Value for Read Timeout set.");
+		}
+		requestFactory.setReadTimeout(client.getRequestReadTimeout());
 		RestTemplate template = new RestTemplate(requestFactory);
 		Map<String, String> headers = new HashMap<>(client.getHeaders());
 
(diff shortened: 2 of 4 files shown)


indraneelb1903:
Also authorization property is deprecated. Is it a good time to remove it?

spencergibb:
Yes, it's a good time to remove authorization.

indraneelb1903:
Let me know , if we need to add any documentation for the new properties.

ryanjbaxter:
Yes please add documentation for the new properties

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -289,6 +294,10 @@ public GenericRequestHeaderInterceptor(Map<String, String> headers) {
 			this.headers = headers;
 		}
 
+		public Map<String, String> getHeaders() {
```
ryanjbaxter: Since this is only used for testing purposes can you just extend the class in the test and add the method there?

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -1119,6 +1119,28 @@ To ensure high availability when you have multiple instances of Config Server de
 
 If you use HTTP basic security on your Config Server, it is currently possible to support per-Config Server auth credentials only if you embed the credentials in each URL you specify under the `spring.cloud.config.uri` property. If you use any other kind of security mechanism, you cannot (currently) support per-Config Server authentication and authorization.
 
+=== Config Client Properties
+
+The `spring.cloud.config.authorization` property has now been removed. It can now be provided through the additional headers used to create the client request - `spring.cloud.config.headers.authorization`. Also , two new properties have been added - `spring.cloud.config.requestReadTimeout` and `spring.cloud.config.sendState`. Please find below the complete list of client properties along with their description.
```
ryanjbaxter: I dont think we need all of this.  It is a enough to add a section on configuring read timeouts.  The properties will automatically be updated using a script when we release Finchley.

indraneelb1903: So,If I understand correctly , no need for complete list of properties? And add a small section on configuring read timeouts? Is that right?

ryanjbaxter: yup

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
indraneelb1903: Are you sure downcasting like this will work , as we are creating an instance of GenericRequestHeaderInterceptor in getSecureRestTemplate method, even if we create a subclass of GenericRequestHeaderInterceptor in the test for the getter method , casting to that subclass will fail.

indraneelb1903: In Object oriented terms, suppose the subclass is called TestRequestHeaderInterceptor then , a genericRequestHeaderInterceptor cannot be a TestRequestHeaderInterceptor but vice-versa is true.

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -1119,6 +1119,10 @@ To ensure high availability when you have multiple instances of Config Server de
 
 If you use HTTP basic security on your Config Server, it is currently possible to support per-Config Server auth credentials only if you embed the credentials in each URL you specify under the `spring.cloud.config.uri` property. If you use any other kind of security mechanism, you cannot (currently) support per-Config Server authentication and authorization.
 
+=== Configuring Read Timeouts
+
+If you want to configure read timeout , this can be done by using the property `spring.cloud.config.requestReadTimeout`.
```
ryanjbaxter: remove the space after timeout

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
ryanjbaxter: I was talking about creating a subclass of `ConfigServicePropertySourceLocator`

indraneelb1903: I could be missing something, but I don't see how that would help. You will need to extend `GenericRequestHeaderInterceptor` to add the getter for the headers , as it is in the inner class. Feel free to correct me if I am wrong.

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -1119,6 +1119,10 @@ To ensure high availability when you have multiple instances of Config Server de
 
 If you use HTTP basic security on your Config Server, it is currently possible to support per-Config Server auth credentials only if you embed the credentials in each URL you specify under the `spring.cloud.config.uri` property. If you use any other kind of security mechanism, you cannot (currently) support per-Config Server authentication and authorization.
 
+=== Configuring Read Timeouts
+
+If you want to configure read timeout, this can be done by using the property `spring.cloud.config.requestReadTimeout`.
```
spencergibb: We should reference the canonical format, so `request-read-timeout`.

indraneelb1903: Sure, will make the change in the evening.

indraneelb1903:
@spencergibb are the changes alright?

Issue #1032, opened 2018-05-30
NPE on Windows due to improper use of lastIndexOf(File.separator) (Finchley.RC2)


SladeRun14:
org.springframework.cloud.config.server.environment.HttpClientConfigurableHttpConnectionFactory.lookupHttpClientBuilder(URL url)

Line 68:

`spec = spec.substring(0, spec.lastIndexOf(File.separator));`

Issue:
The statement above results in an NPE on Windows since a `File.separator` on Windows is a backslash and therefore will never be found in a URL String (which only ever contains forward slashes). This code will work fine on operating systems having forward slashes as file separators, but makes Finchley.RC2 unusable in a Windows environment.

Context:
```
private HttpClientBuilder lookupHttpClientBuilder(URL url) throws MalformedURLException {
  String spec = url.toString();
  HttpClientBuilder builder = httpClientsByUri.get(spec);
  while (builder == null) { 
    spec = spec.substring(0, spec.lastIndexOf(File.separator));
    builder = httpClientsByUri.get(spec); 
  }
  return builder;
}
```

ryanjbaxter:
@pivotal-dylan-roberts ^^^

pivotal-dylan-roberts:
working on this now

fahimfarookme:
Faced the same issue. This worked for me.

```
private HttpClientBuilder lookupHttpClientBuilder(URL url) throws MalformedURLException {   	
        final Optional<HttpClientBuilder> cb = httpClientsByUri.entrySet()
                .stream()
                .filter(entry -> url.toString().startsWith(entry.getKey()))
                .map(Map.Entry::getValue)
                .findFirst();
        return cb.get();
}
```

bingbingll:
Currently, there is no good solution but to reduce the version, using 1.5.14.

fahimfarookme:
Until #1037 is released, you can define a custom [ConfigurableHttpConnectionFactory](https://github.com/fahimfarookme/config-server-stuck-thread-issue/blob/master/config-server-fix/src/main/java/me/fahimfarook/config/server/TimeoutHttpConnectionFactory.java) class and a [bean](https://github.com/fahimfarookme/config-server-stuck-thread-issue/blob/master/config-server-fix/src/main/java/me/fahimfarook/config/server/ConfigServerApplication.java#L22)

Issue #1028, opened 2018-05-29
Support for application placeholders in Git URIs is broken


georgeharley:
When a simple config server application is built using Spring Boot 2.0.2.RELEASE and Spring Cloud Finchley.RC2 it fails to return configuration from Git repositories that are declared with an application placeholder. This was previously working with Finchley.RC1.  
  
In more detail, when a request is made to the running server for configuration an exception is thrown...
```
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.substring(String.java:1967) ~[na:1.8.0_121]
	at org.springframework.cloud.config.server.environment.HttpClientConfigurableHttpConnectionFactory.lookupHttpClientBuilder(HttpClientConfigurableHttpConnectionFactory.java:69) ~[spring-cloud-config-server-2.0.0.RC2.jar:2.0.0.RC2]
	at org.springframework.cloud.config.server.environment.HttpClientConfigurableHttpConnectionFactory.create(HttpClientConfigurableHttpConnectionFactory.java:56) ~[spring-cloud-config-server-2.0.0.RC2.jar:2.0.0.RC2]
	at org.eclipse.jgit.transport.TransportHttp.httpOpen(TransportHttp.java:830) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.transport.TransportHttp.connect(TransportHttp.java:491) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.transport.TransportHttp.openFetch(TransportHttp.java:345) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.transport.FetchProcess.executeImp(FetchProcess.java:137) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.transport.FetchProcess.execute(FetchProcess.java:123) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.transport.Transport.fetch(Transport.java:1269) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
[...]

PR #1037, opened 2018-05-31
Update custom jgit connection factory for uri placeholders

pivotal-dylan-roberts:
- Due to jgit architecture, which uses static variables for http connection config, need to look up proxy/skip-ssl-validation config by url at time of call
 - If this config is associated with a template url containing placeholders, looking up by a fully realized url becomes tricky
 - This implementation attempts to reverse engineer the template url from the fully realized one and does the lookup after that
 - There is one case that doesn't work here, that is when two template URLs are identical other than the placeholder names. In this case we'll find >1 configurations which are not safe to randomly choose between, presumably; so we log an error and use default config (no proxy or skip-ssl-validation applied).

gh-1028
gh-1032

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
@@ -15,27 +15,41 @@
  */
 package org.springframework.cloud.config.server.environment;
 
-import java.io.File;
 import java.io.IOException;
-import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
 import java.security.GeneralSecurityException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.impl.client.HttpClients;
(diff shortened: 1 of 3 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
```diff
@@ -53,22 +67,81 @@ public HttpConnection create(URL url) throws IOException {
 
     @Override
     public HttpConnection create(URL url, Proxy proxy) throws IOException {
```
pivotal-dylan-roberts: @spencergibb `create()` here is called when jgit is about to connect to a repo. This is when we need to lookup the appropriately configured `HttpClientBuilder`. As you can see, all we have to go on is the URL. This URL param will have any placeholders filled in with actual values. So the trick here is looking up the correct `HttpClientBuilder` with nothing but this URL.

ryanjbaxter:
@pivotal-dylan-roberts could you give an example of what the configuration would look like for this

> There is one case that doesn't work here, that is when two template URLs are identical other than the placeholder names. In this case we'll find >1 configurations which are not safe to randomly choose between, presumably; so we log an error and use default config (no proxy or skip-ssl-validation applied).

pivotal-dylan-roberts:
@ryanjbaxter config for the unsupported case would look like this:

...
composite:
- type: git
  uri: http://host/{application}-repo.git
  skip-ssl-validation: true
  ...
- type: git
  uri: http://host/{another-property-placeholder}-repo.git
  skip-ssl-validation: false
  ...

ryanjbaxter:
Does it have anything to do with `skip-ssl-validation`?  How would we decide which repo to use in general, they would both match in this case, right?

pivotal-dylan-roberts:
@ryanjbaxter Yes, it affects skip-ssl-validation as well as proxy. When they both match, there's no sane way to choose, I believe. This PR is presently logging an error and using default settings in this case. So, no skip-ssl-validation or proxy.

PR #1041, opened 2018-06-01
Update docs on composite environment repositories

### LINKED ISSUE #116: spring-cloud-config-client should not depend on spring-boot-starter
As `spring-cloud-config-client` is first and foremost a library and not an application, it should not depend on `spring-boot-starter` which in turn pulls in `spring-boot-starter-logging` which in turn pulls in `logback-classic` which most likely is not needed by `spring-cloud-config-client` at all.

IMHO libraries should carefully define which libraries they need and only depend on those. Applications/microservices on the other hand can use the Spring Boot Starter POMs to get a consistent set of dependencies.

I would guess `spring-cloud-config-client` only needs `spring-boot`, `spring-boot-autoconfigure` and `slf4j-api` (and `spring-security-crypto`, which is managed by its own anyway).

### LINKED ISSUE #933: Document new composite yml config
See #616 
@pivotal-dylan-roberts

/cc @csterwa

fifthposition:
Fixes #933

Note that there's no mention of deprecation regarding the existing composite method (using multiple profiles for distinct backend types). If we're deprecating that, we may want to include a notice in the docs here in addition to the Finchley release notes.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -659,34 +659,61 @@ Property values behave in the same way as they would if they came from Spring Bo
 ==== Composite Environment Repositories
 
 In some scenarios, you may wish to pull configuration data from multiple environment repositories.
-To do so, you can enable multiple profiles in your config server's application properties or YAML file.
-If, for example, you want to pull configuration data from a Git repository as well as an SVN repository, you can set the following properties for your configuration server:
+To do so, you can enable the `composite` profile in your configuration server's application properties or YAML file.
+If, for example, you want to pull configuration data from a Subversion repository as well as two Git repositories, you can set the following properties for your configuration server:
 
 [source,yaml]
 ----
 spring:
   profiles:
-    active: git, svn
+    active: composite
   cloud:
     config:
       server:
-        svn:
+        composite:
+        -
+          type: svn
           uri: file:///path/to/svn/repo
-          order: 2
+        -
+          type: git
+          uri: file:///path/to/rex/git/repo
+        -
+          type: git
+          uri: file:///path/to/walter/git/repo
+----
+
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -659,34 +659,61 @@ Property values behave in the same way as they would if they came from Spring Bo
...
+Using this configuration, precedence can be determined by an `order` property.
+You can use the `order` property to specify the priority order for all your repositories.
 The lower the numerical value of the `order` property, the higher priority it has.
 The priority order of a repository helps resolve any potential conflicts between repositories that contain values for the same properties.
 
 NOTE: Any type of failure when retrieving values from an environment repository results in a failure for the entire composite environment.
 
 NOTE: When using a composite environment, it is important that all repositories contain the same labels.
-If you have an environment similar to the one in the preceding example and you request configuration data with the `master` label but the SVN repository does not contain a branch called `master`, the entire request fails.
+If you have an environment similar to the ones in the preceding examples and you request configuration data with the `master` label but the Subversion repository does not contain a branch called `master`, the entire request fails.
```
Buzzardo: I always avoid pluralizing "one". I'd replace it with "those".

fifthposition: 👍 Thanks, Jay!

ryanjbaxter:
@spencergibb what do you think about adding a note about deprecating the old way of doing composite repos?

spencergibb:
Are we deprecating it? It won't work in some future release?

ryanjbaxter:
It will works but the use case of having a composite repo composed of 2 different repo types is now covered by the new composite functionality so there is no need to have both

ryanjbaxter:
@fifthposition I talked to @spencergibb and I dont think we want to deprecate the "old" method right now, so what you have looks good.

PR #1054, opened 2018-06-11
Adding ssh proxy capability

bisigc:
Adding the capability to configure a proxy host and port for ssh git
repo connections.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/SshUri.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/SshUri.java
@@ -38,7 +38,7 @@ public abstract class SshUri {
 	private boolean ignoreLocalSshSettings;
 	private boolean strictHostKeyChecking = true;
 	private String proxyHost;
-	private int proxyPort = 8080;
+	private Integer proxyPort;
 
 	public static SshUriPropertiesBuilder builder() {
 		return new SshUriPropertiesBuilder();
@@ -116,11 +116,11 @@ public abstract class SshUri {
 		this.proxyHost = proxyHost;
 	}
 
-	public int getProxyPort() {
+	public Integer getProxyPort() {
 		return proxyPort;
 	}
 
-	public void setProxyPort(int proxyPort) {
+	public void setProxyPort(Integer proxyPort) {
 		this.proxyPort = proxyPort;
 	}
 
@@ -147,7 +147,7 @@ public abstract class SshUri {
 		private boolean ignoreLocalSshSettings;
 		private boolean strictHostKeyChecking = true;
 		private String proxyHost;
-		private int proxyPort = 8080;
+		private Integer proxyPort;
 		private Map<String, SshUriNestedRepoProperties> repos = new LinkedHashMap<>();
 
 		SshUriPropertiesBuilder() {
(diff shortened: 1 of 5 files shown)


bisigc:
Hi @ryanjbaxter 
Thanks for your reply. At the moment it is implemented, that the proxy host is set for file and property based ssh transport configurations. Do you think it will be better to use the property porxyHost only when `ignoreLocalSshSettings: true`? 

I will try to add some tests.

Regards,
Chris

ryanjbaxter:
It seems like the rest of the ssh settings in configuration only take effect when `ignoreLocalSshSettings` is true, so I think that makes sense.
http://cloud.spring.io/spring-cloud-static/Finchley.RC2/single/spring-cloud.html#_git_ssh_configuration_using_properties

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/SshUri.java
```diff
@@ -37,6 +37,8 @@
 	private String preferredAuthentications;
 	private boolean ignoreLocalSshSettings;
 	private boolean strictHostKeyChecking = true;
+	private String proxyHost;
+	private int proxyPort = 8080;
```
spencergibb: Is this a good idea? Seems like this should be a required field, shouldn't it?

bisigc: Do you mean the proxyPort field?

spencergibb: Yes, the line I commented on.

bisigc: Ok, I could write a validator which verifies, that proxyPort has to be set in case proxyHost is set.

spencergibb: Just do it in `PropertyBasedSshSessionFactory.java` after the not null check for the host. Use spring `Assert` class.

bisigc:
@spencergibb Ok I have changed the field type of proxyPort from int to Integer and added a null check in case the proxyHost is set. I added also a negativ test.

bisigc:
Ok, I'll do that. Since the properties are optional, I will not add them to the Example. How about: 

*proxyHost*
Hostname for the ssh proxy connection. Is optional and used only when `ignoreLocalSshSettings` is true

*proxyPort*
Port for the ssh proxy connection. Must be set if `proxyHost` is also set

ryanjbaxter:
Sounds like a good start

bisigc:
Hi @ojhughes, in my case I required to have the proxy connection via http, since the proxy server did not support a SocksProxy connection. Its works fine. There is an example on the jsch page for doing a ssh session via http proxy: http://www.jcraft.com/jsch/examples/ViaHTTP.java.html . Maybe I can add the fact, that an http proxy connection is done to the desciption of the properties in the docs. But I would suggest to add additional properties (e.g. `socksProxyHost` and `socksProxyPort`) or a switch property (e.g. `proxyConnectionType` with the values `http` and `socks`) to create the connection either via http or a socket connection. What do you think?

bisigc:
Hi @spencergibb 
How about your change requests? Are they now fullfilled?
Thank you and regards,
Christian

Issue #1081, opened 2018-07-11
Need detailed error info when using spring cloud config client


sunnykaka:
I'm using spring cloud config and git for configuration management.

When i modified yaml file in git and make some mistake, like: `a:1`

I start my app and got error message like this(with failfast property):

```
java.lang.IllegalStateException: Could not locate PropertySource and the fail fast property is set, failing
...
Caused by: org.springframework.web.client.HttpServerErrorException: 500 Internal Server Error
...
```

No more detailed info in app console.

While in my config server i can find what happened:

```
org.yaml.snakeyaml.scanner.ScannerException: mapping values are not allowed here
 in 'reader', line 16, column 10:
      a:a1
```

It's sophisticated to dive into config server to find detail info.

If i can see this error info in my app, it will help a lot.

PR #1082, opened 2018-07-11
Log more detailed error info when failfast=true and error happened

sunnykaka:
FIxes [1081](https://github.com/spring-cloud/spring-cloud-config/issues/1081)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -134,8 +134,8 @@ public ConfigServicePropertySourceLocator(ConfigClientProperties defaultProperti
 		}
 		if (properties.isFailFast()) {
 			throw new IllegalStateException(
-					"Could not locate PropertySource and the fail fast property is set, failing",
-					error);
+					"Could not locate PropertySource and the fail fast property is set, failing" +
+						(errorBody == null ? " " : ": " + errorBody), error);
```
ryanjbaxter: Why the space if there is no `errorBody`?

sunnykaka: @ryanjbaxter 
That's right, space was deleted.

Issue #1094, opened 2018-07-26
Config Server unable to access shared vault when backend/root contains "/"


alfusinigoj:
We are facing a weird issue connecting vault with config server where the backend seems to something like this  “secret/<some-guid>“. Here by default config server is escaping the forward slash and causing issue when connecting to vault.

***** Vault configuration below…
```"credentials": {
           "vault": "https://vault......net",
           "vaults": [
             "https://vault.sys.......net"
           ],
           "token": "bf3588c2-0a72-4ff1-b8ef-6b13ab7a5212",
           "root": "secret/dab94b6f-0cea-49bb-abc3-4704c796fae1"
         }
```

***** Logs...
```
: Created GET request for "https://vault.sys.fog.....net:443/v1/secret%2Fdab94b6f-0cea-49bb-abc3-4704c796fae1/XXXX,Production"
```

As seen above the real backend "secret/dab94b6f-0cea-49bb-abc3-4704c796fae1" is encoded to "secret%2Fdab94b6f-0cea-49bb-abc3-4704c796fae1"

FYI -- This is moved from https://github.com/spring-cloud/spring-cloud-vault/issues/232 to here

ryanjbaxter:
The backend should not include a `/`, just leave it as `secret`.  The application name should include the guid for example `dab94b6f-0cea-49bb-abc3-4704c796fae1/foo`.  

When making the request to the config server it is important that you encode the `/` correctly in the application name parameter.

For example 
```
$ curl -X "GET" "http://localhost:8888/dab94b6f-0cea-49bb-abc3-4704c796fae1(_)foo/default" -H "X-Config-Token: token"
```

The documentation on encoding the `/` is actually included under the Git Backend section of our docs
http://cloud.spring.io/spring-cloud-static/Finchley.RELEASE/single/spring-cloud.html#_git_backend

dwilliss:
That wouldn't work in the case where the back end is a git repository (or a combined git/vault scenario).
The git repo would then have to have the guid in it, which it doesn't need as each config server can specify a different git location using whatever identifier it needs (preferably humanly-readable). The GUID is unique to vault.

Ideally, the backend value of the vault section of the config should be able to take the whole "root" from the vault config.  However, since it's a REST object, it wants to URL encode that.
One solution would be to provide another optional parameter for the GUID part of the root and insert that into the URI if it's not blank.

ryanjbaxter:
[...]

PR #1098, opened 2018-08-01
Omit secret backend URL encoding for Vault.

mp911de:
We now construct the URL template from the base URL, API version prefix and secret backend upfront and apply templating only for the secret key. This bypasses URL encoding for the secret backend which allows for special characters such as slashes to be passed thru directly.

Fixes gh-1094.


ryanjbaxter:
@spencergibb I dont think there needs to be a documentation change since this should have already been supported.

PR #1107, opened 2018-08-13
Read Gitee Push Hook contents as PropertyPathNotification

ian4hu:
Read Gitee Push Hook contents as PropertyPathNotification, instead of use the fixed `application.yml`

[Abount Gitee Web Hooks](http://git.mydoc.io/?t=154711)

### DIFF
--- a/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/GiteePropertyPathNotificationExtractor.java
+++ b/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/GiteePropertyPathNotificationExtractor.java
@@ -17,7 +17,9 @@
 package org.springframework.cloud.config.monitor;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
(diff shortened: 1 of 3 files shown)


ryanjbaxter:
Can you explain more about why you think this change should be made?

ian4hu:
Previous version always emit a fixed file `application.yml` for every push event, but the actual commits may contents various files changed. 

According to the implementation about Github & Gitlab, the modification tried to read the full commit changelist and emit the PropertyPathNotification.

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/GiteePropertyPathNotificationExtractor.java
```diff
@@ -39,11 +41,29 @@
...
+				@SuppressWarnings("unchecked")
+				Collection<Map<String, Object>> commits = (Collection<Map<String, Object>>) request
+						.get("commits");
+				for (Map<String, Object> commit : commits) {
+					addAllPaths(paths, commit, "added");
+					addAllPaths(paths, commit, "removed");
+					addAllPaths(paths, commit, "modified");
+				}
+				if (!paths.isEmpty()) {
+					return new PropertyPathNotification(paths.toArray(new String[0]));
```
ryanjbaxter: Shouldnt the size of the String array be the size of `paths`?

ian4hu:
@ryanjbaxter it would be better if the String array size is same as the origin list (By the way, if the array size is not enough, toArray will create a new array with the list size.). And I have fixed it with proper size. Thanks.

ryanjbaxter:
If you would like this to be included in the next FInchley release you will need to submit the PR against the 2.0.x branch.

ian4hu:
@ryanjbaxter OK, should I keep or close this PR, then create a new PR against the 2.0.x branch ?

ryanjbaxter:
No you should be able to target the 2.0.x branch instead of master by editing the PR

ian4hu:
@ryanjbaxter done.

ryanjbaxter:
Covered in https://github.com/spring-cloud/spring-cloud-config/pull/1120

PR #1171, opened 2018-10-23
#1169: add accept header to ensure client requests json

### LINKED ISSUE #1169: Set accept header in ConfigServicePropertySourceLocator
We currently face an issue which was already reported multiple times:

https://github.com/spring-cloud/spring-cloud-config/issues/306
https://github.com/spring-cloud/spring-cloud-config/issues/368

If you run a spring-cloud server and eureka-server in the same service, the output of the config-server does no longer default to JSON but XML.
This especially breaks the usage of lists in your configuration as they are not serialised correctly.

A workaround is to exclude the jackson-xml dependency in the eureka-server but this might cause other problems. 
As by now, I ask myself if it would not be the easiest/best solution to explicitly set the accept header in config-client so that it requests the correct serialised version of the configuration. 
This bypasses the serialisation into XML and communication between config-server / config-client works fine again.

deveth0:
Fixes #1169

This PR should serve the issue, that a config-client is served XML instead of JSON when the config-server is also running eureka (or has the jackson-dataformat-xml dependency included due to some other reason).

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.config.client;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


spencergibb:
Merged with a test via 7f9c9dd4ca2f571cba4d957049ee2c5b5bc9deba

PR #1180, opened 2018-11-07
Dont use parent of basedir when using a parameterized url

### LINKED ISSUE #1175: Spring Cloud Config Git Basedir
I am testing out spring-cloud-config-server and getting an odd situation where repo directories are not created as expected.

Setting spring.cloud.config.server.git.basedir to a given path doesn't create config-repo-* directories below that path. E.g.

```
spring:
      cloud:
          config:
              server:
                  git:
                      basedir: "/config-server/repos"
```

The above will be created but the repo directories will be stored in /config-server instead of /config-server/repos I am using 2.0.2.RELEASE of spring-cloud-config-server.

ryanjbaxter:
Fixes gh-1175


spencergibb:
Hmm, seems that is how we meant it to be.

https://github.com/spring-cloud/spring-cloud-config/commit/f67bdba5ed418f9bc9c5f3c21107e7aab8d24954

ryanjbaxter:
> The basedir still has to be unique, so we now use the same parent directory
as the parent repo, and add the temp file name as a directory name. I.e.
if basedir=/foo/bar the pattern matched basedir will be something like
/foo/config-repo-23874691.

But why cant it be unique inside the `basedir` itself?  Why does it need to be the parent?

spencergibb:
I think this stems from pre multiple jgit environment repository, when there was only one, but it used patterns. There could be something checked out into basedir itself.

spencergibb:
I'm just worried about breaking backwards compatibility.

ryanjbaxter:
Yeah I understand the backwards compatibility concern, so that would rule out Finchley.  But looking at Greenwich, doesnt the random temp directory eliminate the possibility of overlapping names for multiple git repos?

spencergibb:
yeah, I can't see a downside.

ryanjbaxter:
So OK to merge?

@dsyer do you have any opinions?

PR #1185, opened 2018-11-12
Override native default label if config server default label is not null

### LINKED ISSUE #1184: Property spring.cloud.config.server.native.default-label is being ignored
The value of `spring.cloud.config.server.native.default-label`, which should default to `master` is always overwritten by the value of `spring.cloud.config.server.default-label`.

If `server.default-label` is non-null, this behaviour is correct and consistent with SCM-based repositories, but in case `server.default-label` is null, the native repository's setting should not be overwritten.

The Git and SVN repository factories check `server.default-label` and only set the default label on the created repository if it is non-null.

ryanjbaxter:
Fixes #1184

### DIFF
--- a/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepositoryFactoryTest.java
+++ b/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepositoryFactoryTest.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.config.server.environment;
+
+import org.junit.Test;
+import org.springframework.cloud.config.server.config.ConfigServerProperties;
+import org.springframework.core.env.StandardEnvironment;
+
+import static org.junit.Assert.assertEquals;
+
+
+/**
+ * @author Ryan Baxter
+ */
+public class NativeEnvironmentRepositoryFactoryTest {
+
+	@Test
+	public void testDefaultLabel() {
+		ConfigServerProperties props = new ConfigServerProperties();
+		props.setDefaultLabel("mylabel");
+		NativeEnvironmentRepositoryFactory factory = new NativeEnvironmentRepositoryFactory(new StandardEnvironment(), props);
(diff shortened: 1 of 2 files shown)


spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepositoryFactoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+		NativeEnvironmentRepositoryFactory factory = new NativeEnvironmentRepositoryFactory(new StandardEnvironment(), props);
+		NativeEnvironmentProperties environmentProperties = new NativeEnvironmentProperties();
+		NativeEnvironmentRepository repo = factory.build(environmentProperties);
+		assertEquals("mylabel", repo.getDefaultLabel());
+
+		factory = new NativeEnvironmentRepositoryFactory(new StandardEnvironment(), props);
+		environmentProperties = new NativeEnvironmentProperties();
+		environmentProperties.setDefaultLabel("mynewlabel");
+		repo = factory.build(environmentProperties);
+		assertEquals("mylabel", repo.getDefaultLabel());
```
spencergibb: Shouldn't this be "mynewlabel"?

spencergibb: I guess I'm confused as server default label has precedence over the repo specific one.

ryanjbaxter: I thought so too ;)  But the way SVN and GIT repos work today is the config server default label property overrides the environment repo default label property as long as its not `null`.

https://github.com/spring-cloud/spring-cloud-config/blob/2.0.x/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/SvnEnvironmentRepositoryFactory.java

spencergibb: :-) but why?

ryanjbaxter: its a good question, i dont see where we have documented the correct behavior, my worry about changing it now would be breaking people relying on it.

spencergibb: oh well, best make it consistent.

sunruh: > :-) but why?

@spencergibb They (SVN & Git) only let the config server default override the repositories default if it's non-null as SVN defaults to `trunk` and Git to `master`. So when someone is not using labels at all (and not setting the config server default), they behave correctly and use their implementation-specific default or the default set in their specific configuration. That way one can set the default globally (taking precedence) or on a per-repository basis.

spencergibb: I'm not questioning consistency. I'm questioning why the global takes precedence over the specific. Seems backward to me, but that's for another issue.

PR #1198, opened 2018-11-22
Fix error in fail-fast doc

BartoszMiller:
I'm not familiar with the code base, but I believe that the value of fail-fast flag is documented incorrectly.
My understanding is that:
**fail-fast: true** - no retry
**fail-fast: false** - retry


spencergibb:
This was wrong, failFast needs to be true, otherwise no exception is thrown and retry never happens.

ryanjbaxter:
Reverted here https://github.com/spring-cloud/spring-cloud-config/commit/4f6ae7231b7b7deed6ef4d0bbae8259340db3945

PR #1211, opened 2018-12-07
Adding support for CredHub as a backend

AlbertoImpl:
Adding support for CredHub as a backend.
Added optional dependency to `spring-credhub-core`

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.config.server.environment;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.cloud.config.environment.Environment;
+import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.credhub.core.CredHubOperations;
+import org.springframework.credhub.support.CredentialDetails;
+import org.springframework.credhub.support.SimpleCredentialName;
+import org.springframework.credhub.support.json.JsonCredential;
+import org.springframework.util.StringUtils;
+
+import static java.util.stream.Collectors.toMap;
+
+/**
+ * @author Alberto C. Ríos
+ */
+public class CredhubEnvironmentRepository implements EnvironmentRepository {
+
+	private CredHubOperations credHubOperations;
+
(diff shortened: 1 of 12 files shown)


AlbertoImpl:
Note that `spring-credhub-core` hasn't been released yet, but I wanted to get your feedback before that happened.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java
```diff
@@ -0,0 +1,63 @@
...
+		this.credHubOperations = credHubOperations;
+	}
+
+	@Override
+	public Environment findOne(String application, String profilesList, String label) {
+		if (StringUtils.isEmpty(profilesList)) {
+			profilesList = "default";
+		}
+		if (StringUtils.isEmpty(label)) {
+			label = "master";
```
spencergibb: Does master make sense here?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java
```diff
@@ -0,0 +1,63 @@
...
+		environment.add(new PropertySource("credhub-" + application, properties));
+
+		return environment;
+	}
+
+	private Map<Object, Object> findProperties(String application, String profile, String label) {
+		String path = "/" + application + "/" + profile + "/" + label;
+
+		return credHubOperations.credentials().findByPath(path)
+								.stream()
```
spencergibb: Too far indentation

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java
AlbertoImpl: I don't think it does make a lot of sense, however, a lot of the repositories, including `Jdbc` are using `master` as a default label.
I am happy to remove the default and just do not concatenate any label if nothing was specified, but for the sake of consistency, I decided to add that there. What do you think?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java
AlbertoImpl: Fixed, thanks!

AlbertoImpl:
Just added the missing license headers, thanks for the catch @ryanjbaxter!

AlbertoImpl:
Once `spring-credhub-core` gets released. Where should I add the version @spencergibb?

spencergibb:
What version are you asking about?

AlbertoImpl:
At the moment we are building with `spring-credhub-core` with `2.0.0.BUILD-SNAPSHOT` but a `2.0.0.RELEASE` will come out soon.
For the test to run I hardcoded the version here: https://github.com/spring-cloud/spring-cloud-config/pull/1211/files#diff-27baec15a3f2afe27aa44caf3c8e2e4dR53 but there must be a place where I can set it to `2.0.0.RELEASE` outside of that pom. Am I right?

spencergibb:
In the dependencies pom. You can do that now. Any idea when it will be released?

scottfrederick:
@AlbertoImpl This PR can be updated to use `spring-credhub-core` version `2.0.0.RC1` from the Spring milestone repository.

AlbertoImpl:
Just force-pushed the changes with the new version and moved it to `spring-cloud-config-dependencies/pom.xml`.
Let me know if I need to do something else or I did something wrong @spencergibb 

Thanks!

spencergibb:
@scottfrederick just a reminder we will need a spring-credhub release prior to spring cloud Greenwich GA (mid-Jan).

PR #1216, opened 2018-12-27
pom.xml file cleanup

Buzzardo:
I started by removing the docbkx plugin. Marcin also pointed out that I could remove the mojo plugin and the inherited=false lines, so I did that.

NOTE: This PR depends on https://github.com/spring-cloud/spring-cloud-build/pull/84

DO NOT merge this one until that one is merged.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -33,31 +33,20 @@
 		<profile>
 			<id>docs</id>
 			<build>
-			   <plugins>
-								<plugin>
-									<groupId>org.apache.maven.plugins</groupId>
-									<artifactId>maven-dependency-plugin</artifactId>
-								</plugin>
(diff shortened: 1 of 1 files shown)


karateechop:
Yeah

spencergibb:
@Buzzardo is this still an issue? If so, we'd need to target the 2.2.x branch. Feel free to open a separate PR

Buzzardo:
Yes. We should make this change to keep the projects consistent with one another.

spencergibb:
Things have changed enough that you'll need to see if what we have is good enough on 2.2.x or not.

Buzzardo:
When I go look at the docs pom in 2.2.x, I see that we're OK, so we can let this one go. Thanks.

Issue #1228, opened 2019-01-17
Support lastModified query for file resources?


PeterTeunissen:
We use the Spring Cloud Config Server to serve up some definition files to our application. The round trip from application to config server is quiet large so we want to cache these files inside our application and only refresh the cache if the file on the config server changes. Looking at the code in org.springframework.cloud.config.server.resource.ResourceController it seems it just returns the file content and not much more.  I made a copy of this file and added a query parameter so that we can ask for the lastModified of the resource by adding ?lastModified=true. Is this something that is interesting enough for others ? Does this warrant a PR?

(Attached a crude version of this)
[ResourceController.zip](https://github.com/spring-cloud/spring-cloud-config/files/2769372/ResourceController.zip)

spencergibb:
A PR would be better than an attachment.

rgoers:
I should note that I am modifying Log4j 2 to take advantage of this feature.

PR #1235, opened 2019-01-25
Support If-Modified-Since

rgoers:
Supports If-Modified-Since header and returns Last-Modified header

fixes gh-1228

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/resource/ResourceController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/resource/ResourceController.java
@@ -49,7 +49,6 @@ import static org.springframework.cloud.config.server.support.EnvironmentPropert
  *
  * @author Dave Syer
  * @author Daniel Lavoie
- *
  */
 @RestController
 @RequestMapping(method = RequestMethod.GET, path = "${spring.cloud.config.server.prefix:}")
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
Please see the feedback in #1228 and apply it to this PR

rgoers:
The files have been updated to match the original format.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/resource/ResourceController.java
```diff
@@ -71,87 +70,119 @@ public ResourceController(ResourceRepository resourceRepository,
 
 	@RequestMapping("/{name}/{profile}/{label}/**")
 	public String retrieve(@PathVariable String name, @PathVariable String profile,
-			@PathVariable String label, HttpServletRequest request,
+        	@PathVariable String label, ServletWebRequest request,
```
spencergibb: I wonder about changing the signature on all of these public methods.

AFAICS you are using it for
```
request.checkNotModified(resource.lastModified());
```

Can you do 
```
new ServletWebRequest(HttpServletRequest);
```

in `checkNotModified()`?

It would simplify things and keep signatures.

rgoers: What is the advantage of that? Although the methods are public the only caller of them will be Spring MVC and the unit tests and Spring handles this automatically. As for simpler, I would argue that the current code is simpler because it doesn't have to have the new ServletWebRequest in every endpoint method. In addition, the only place the HttpServletRequest is used is in getFilePath(). So the real argument for doing what you are suggesting is to keep it backward compatible even though that doesn't really matter.

spencergibb: We are very particular about any public api we publish. This being one of them.

rgoers: If Imake the requested change it will still cause the method signatures to be modified. ServletWebRequest must be constructed with the HttpServletRequest and HttpServletResponse. Currently the HttpServletResponse is not being passed to the methods. Without the HttpServletResponse ServletWebRequest can't set the HTTP status code that needs to be returned.

spencergibb: There is a request only constructor

rgoers: Yup, and it is exactly equivalent to calling new ServletWebRequest(request, null) which won't work.

spencergibb: I didn't dig into `checkNotModified()` to see that it required the response.

spencergibb: All right, my team convinced me otherwise. You can remove the backwards compatible methods. If you don't have time, we can do it on merge.

rgoers: OK. I removed them.

Issue #1239, opened 2019-01-28
Composite Repository Definition and spring-cloud-config-monitor fails to start


y24jds:
I wish to use the composite repository feature and the spring-cloud-config-monitor feature at the same time but it appears that spring-cloud-config-monitor does not support multiple repositories as they are created as composite repositories.  

The steps are very easy to reproduce if you want to follow the steps below or I have created a demo repository here:

https://github.com/y24jds/test-configserver

# Steps to reproduce issue of using composite repositories and spring-cloud-config-monitor dependency

1. Used spring Initalizr to create new project with Spring Boot 2.1.2, Config Server, and Cloud Bus.
https://start.spring.io/
Using the Greenwich.RELEASE version of spring-cloud
2. Also added the spring-cloud-config-monitor dependency to pom.xml
```xml<dependency>
<groupId>org.springframework.cloud</groupId>
  <artifactId>spring-cloud-config-monitor</artifactId>
</dependency>
```
3. Also added the AMQP dependency to pom.xml
```xml<dependency>
<groupId>org.springframework.cloud</groupId>
  <artifactId>spring-cloud-starter-bus-amqp</artifactId>
</dependency>
```    
4. Make sure that the main application class is annotated with: @SpringBootApplication, @EnableConfigServer, @Configuration

5. Add the following to the application.properties file (note, it doesnt really matter if the git repo's point to a valid repo or not, just the fact of using a composite repository definition and adding at least 2 repo's breaks spring-cloud-config-monitor):

spring.profiles.active=composite
spring.cloud.config.server.composite[0].type=git
spring.cloud.config.server.composite[0].uri=http://github.com/y24jds/common.git
spring.cloud.config.server.composite[1].type=git
spring.cloud.config.server.composite[1].uri=http://github.com/y24jds/{application}.git

6.  Start the application

# Issue:
Observe that the following is logged:
***************************
APPLICATION FAILED TO START
***************************

Description:

Field scmRepository in org.springframework.cloud.config.monitor.FileMonitorConfiguration required a single bean, but 2 were found:
	- git-env-repo0: defined by method 'build' in null
	- git-env-repo1: defined by method 'build' in null

Action:

Consider marking one of the beans as @Primary, updating the consumer to accept multiple beans, or using @Qualifier to identify the bean that should be consumed

# The Problem:
[...]

PR #1247, opened 2019-02-07
#1239 allow multiple repos in FileMonitorConfiguration.

ask4gilles:
Fixes gh-1239

### DIFF
--- a/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/FileMonitorConfiguration.java
+++ b/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/FileMonitorConfiguration.java
@@ -31,6 +31,7 @@ import java.nio.file.WatchService;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/FileMonitorConfiguration.java
```diff
@@ -185,17 +186,19 @@ public void poll() {
...
-
-				Resource resource = this.resourceLoader
-						.getResource(this.scmRepository.getUri());
-				if (resource instanceof FileSystemResource) {
-					return Collections.singleton(Paths.get(resource.getURI()));
+				for (AbstractScmEnvironmentRepository repository : scmRepositories) {
+					repositoryUri = repository.getUri();
+					Resource resource = this.resourceLoader.getResource(repositoryUri);
+					if (resource instanceof FileSystemResource) {
+						return Collections.singleton(Paths.get(resource.getURI()));
```
spencergibb: I guess you only watch the first one? What if there are multiple `file://` git repos?

ask4gilles:
Hi @spencergibb ,
Thanks for looking at this.
I've corrected the behaviour in case there are multiple paths and I've also added a unit test.
I don't see any IT tests in this module though...

Issue #1260, opened 2019-02-27
Support Google Cloud Source out of the box


ewirch:
`spring-cloud-config` adds extra support for AWS Code Commit repositories (`org.springframework.cloud.config.server.support.AwsCodeCommitCredentialProvider`). Why not also for Google Cloud Source?

Add the following configuration class to the Spring context and this dependency:
```
com.google.auth:google-auth-library-oauth2-http:0.12.0
```
and JGit will automatically pick up the transport configurer. Google auth library will handle the complex part of background authentication (using [application default credentials](https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login)). Works locally for user accounts and on server with service accounts.

```
import static java.util.stream.Collectors.toMap;

import com.google.auth.oauth2.GoogleCredentials;
import org.eclipse.jgit.api.TransportConfigCallback;
import org.eclipse.jgit.transport.Transport;
import org.eclipse.jgit.transport.TransportHttp;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;

import java.io.IOException;
import java.util.List;
import java.util.Map;
import java.util.Map.Entry;
import java.util.Objects;

@Configuration
public class JgitGoogleCloudSourceSupportConfiguration {

    @Bean
    public TransportConfigCallback createGoogleCloudSourceTransportConfigurer() {
        return new InjectAuthorizationTransportConfigurer();
    }

    public static class InjectAuthorizationTransportConfigurer implements TransportConfigCallback {

        private static final String GOOGLE_CLOUD_SOURCE_DOMAIN = "source.developers.google.com";

        @Override
        public void configure(Transport transport) {
            if (transport instanceof TransportHttp) {
                if (isGoogleCloudSourceUri((TransportHttp) transport)) {
                    addHeaders((TransportHttp) transport, getAuthorizationHeaders());
                }
            }
        }

        private boolean isGoogleCloudSourceUri(TransportHttp transport) {
            return Objects.equals(transport.getURI().getHost(), GOOGLE_CLOUD_SOURCE_DOMAIN);
        }

        private void addHeaders(TransportHttp transport, Map<String, String> headers) {
            transport.setAdditionalHeaders(headers);
        }

        private Map<String, String> getAuthorizationHeaders() {
            try {
                return GoogleCredentials.getApplicationDefault()
                    .getRequestMetadata()
                    .entrySet()
                    .stream()
[...]

PR #1264, opened 2019-02-27
Authenticate against Google Cloud Source repositories

ewirch:
Fixes gh-1260

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -338,7 +338,7 @@ If the `aws-java-sdk-core` jar is not on your classpath, the AWS Code Commit cre
 
 Spring Cloud Config Server also supports authenticating against https://cloud.google.com/source-repositories/[Google Cloud Source] repositories.
 
-If your Git URI uses the `http` or `https` protocol and the domain name is `source.developers.google.com`, the Google Cloud Source credentials provider will be used. A Google Cloud Source repository URI has the format `https://source.developers.google.com/p/${GCP_PROJECT}/r/${REPO}`. To obtain the URI for your repository, click on "Clone" in the Google Cloud Source ui, and select "Manually generated credentials". Do not generate any credentials, simply copy the displayed URI.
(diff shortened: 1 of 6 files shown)


docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -334,6 +334,17 @@ AWS EC2 instances may use http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam
 NOTE: The `aws-java-sdk-core` jar is an optional dependency.
 If the `aws-java-sdk-core` jar is not on your classpath, the AWS Code Commit credential provider is not created, regardless of the git server URI.
 
+===== Authentication with Google Cloud Source
+
+Spring Cloud Config Server also supports authenticating against https://cloud.google.com/source-repositories/[Google Cloud Source] repositories.
+
+If your Git URI uses the `http` or `https` protocol and the domain name is `source.developers.google.com`, the Google Cloud Source credentials provider will be used. A Google Cloud Source repository URI has the format `https://source.developers.google.com/p/${GCP_PROJECT}/r/${REPO}`. To obtain the URI for your repository, click on "Clone" in the Google Cloud Source ui, and select "Manually generated credentials". Do not generate any credentials, simply copy the displayed URI.
```
ryanjbaxter: capitalize UI

ewirch:
> It would be good if you could add some tests

I did not find testable and useful test cases. I could test if response from `getRequestMetadata()` (`Map<String, List<String>>`) is properly converted to `Map<String, String>`. But I'd need to mock a static method to do so. And the conversion code is simple. So cost and benefit are not balanced.

It would be useful to test if JGit can actually authenticate against a Google Cloud Source repository. But I see no practical way to do this. The Google OAuth2 lib talks directly to Google servers when generating auth headers. I'd have to use an actual Google user for this to succeed. I have no knowledge of a Google Cloud Source repository stub implementation. I'd have to use a real Google Cloud Source repo.

ewirch:
The failing tests show, that my PR was premature. Will close it for improvement.

PR #1272, opened 2019-03-01
Prevent JGit from loading local configuration in tests

ewirch:
Without any preparations JGit defaults to read system / user / project
configuration from local file system. Users running tests locally
might see test failures due to their config diverging from the test
author config.

For me JGitEnvironmentRepositoryIntegrationTests.pullDirtyRepo() was
constantly failing, because JGit calculated a different hash for the
created commit than on the build server.


ewirch:
Does anything prevent the merge?

spencergibb:
Nope. It will happen in a few days

ewirch:
Rebased on top of master.

ryanjbaxter:
@ewirch would you mind submitting this against the 2.0.x branch and we will take care of merging it forward?  I think it would be useful across all the active branches.

ewirch:
Sure thing. Rebased.

Issue #1270, opened 2019-02-28
Connect to Multiple Vault Secrets Engines


kevin-thompson:
Since Vault allows for multiple secrets engines, it seems reasonable to expect that one might use multiple engines for the same project. For example, Vault supports database secrets engines that manage, well, database secrets, but many other properties will just be key-values. It would be nice to pass a list of secrets engines into the "backend" property rather than it just accepting a single secrets engine.

kamalakarp:
@kevin-thompson - This week I added the vault  namespace changes and it'll be good to add other secret engines as well. I can contribute the related code changes and please let me know what help is needed for me.

atdiff:
What's the status on this? Would be a useful feature

spencergibb:
Were waiting for votes before this gets prioritized.

luroto:
Hi there! Just out of curiosity, is there any updates related to this issue? Or is there any workaround? I've been searching for some time and I didn't get a clue related to this, besides to use different paths inside the very unique secret engine. Thanks in advance!

PR #1273, opened 2019-03-01
add support for multiple vault secrets engines

kevin-thompson:
To address the issue #1270 

Changes the property backend to backends and makes it a string array. Leaves "secret" as the default.
Updates the name for the property source to be "vault/{backend}:{key}" instead of just "vault:{key}"

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.DeprecatedConfigurationProperty;
 import org.springframework.cloud.config.server.proxy.ProxyHostProperties;
 import org.springframework.cloud.config.server.support.HttpEnvironmentRepositoryProperties;
 import org.springframework.core.Ordered;
@@ -43,7 +44,7 @@ public class VaultEnvironmentProperties implements HttpEnvironmentRepositoryProp
 	/** Timeout (in seconds) for obtaining HTTP connection, defaults to 5 seconds. */
 	private int timeout = 5;
 
-	/** Vault backend. Defaults to secret. */
+	/** Vault backend. Defaults to a single backend named secret. */
 	private String[] backends = new String[] { "secret" };
 
 	/**
@@ -101,6 +102,18 @@ public class VaultEnvironmentProperties implements HttpEnvironmentRepositoryProp
 
 	public void setScheme(String scheme) {
 		this.scheme = scheme;
+	}
+
+	@DeprecatedConfigurationProperty(replacement = "spring.cloud.config.server.vault.backends")
+	@Deprecated
+	public String getBackend() {
+		return this.backends != null && this.backends.length > 0 ? this.backends[0]
+				: null;
+	}
+
+	@Deprecated
+	public void setBackend(String backend) {
+		setBackends(new String[] { backend });
 	}
 
 	public String[] getBackends() {
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
@@ -44,7 +44,7 @@ public class VaultEnvironmentProperties implements HttpEnvironmentRepositoryProp
 	private int timeout = 5;
 
 	/** Vault backend. Defaults to secret. */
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@kevin-thompson Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1273)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1273) to manually synchronize the status of this Pull Request.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
```diff
@@ -102,12 +103,12 @@ public void setScheme(String scheme) {
 		this.scheme = scheme;
 	}
 
-	public String getBackend() {
```
spencergibb: There is a process to deprecate properties that spring boot has. We'll need to use it here. I'm on mobile, but if you haven't found it by Monday, ping me and I'll find it.

pivotal-issuemaster:
@kevin-thompson Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1273)!

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
kevin-thompson: I'll look into it, thanks.

kevin-thompson:
> 
> 
> So you'll need to keep the old tests and add new as well to cover the deprecation

I didn't remove any test. I added a single test and cleaned up the mocking of the responses.

I did update the test that was using setBackend() to use setBackends(). Are you saying I should duplicate that with the one and continue to use the deprecated method in one place and use the new method in the other?

spencergibb:
Yes, are set back end if restored there needs to be tests that verify backwards compatibility

kamalakarp:
@kevin-thompson - the changes you did will only work for multiple KV(Key/Value) secret backends and it doesn't support database or other secret engines because the path is different.

KV secret engine: **vault read secret/configclient** which makes the URL : http://127.0.01:8200/v1/secret/foo

Database secret engine : **vault read database/creds/test-mongo-db-role** which makes the URL : http://127.0.01:8200/v1/database/creds/test-mongo-db-role

If your intention is to support the multiple secret backends like KV, database, and transit - then paths need to be adjusted properly and read the data from Vault

image

In the picture, config, cubbyhole, kv, and secret are variants of KV secret engines where as database and transit secret engines have different pathing.

kevin-thompson:
@kamalakarp 
Well, you can massage the correct path into place by setting a default key, but I agree that's not ideal.
The bigger problem with database backends is that they require a completely different response wrapper. Additionally, they necessitate (not always, but practically), the need to handle the lease information. That brings up a whole different issue because it's not the config server that's using the lease information, it's the consuming app. We could take care of the leasing on the config server side, but that would require a similar implementation of push notifications that the Git implementations use. I'm not prepared to tackle that here.

You bring up a great point about the kvVersion possibly applying differently, and I would not be opposed to reworking the code to turning backends into a list of objects rather than strings. This would better support the (hopefully) eventual support of the database backends (it would need additional properties such as the names of the properties into which the username/password should be dumped).

Also, you bring up the transit backend. I'm not sure that really belongs in the config server as it's just vault encrypting passed information, not setting up relatively static properties.

kevin-thompson:
@spencergibb I did not find any documentation describing how to deprecate properties, but I did find the annotation that I applied (@DeprecatedConfigurationProperty). That seems like what you were talking about.
@ryanjbaxter I updated the documentation. Let me know if I missed anything.

I'd like to hear back from @kamalakarp to get a response on the string list vs. object list thing.

kamalakarp:
@kevin-thompson - it would be better if we can make backend as a list of objects. 
With these changes, we will have to deprecate backend, defaultKey and kv.Version and make them as an object.

naxhh:
I've just started familiarizing myself with the project so dont take me so serious.

I agree with @kamalakarp in the fact that would be better to manage vault backends as a list of objects.
The current approach doesn't allow you to have KV backends with v1 and v2 at the same time.

Also the backends can be under any path, so that's also it will need to be taken care of.

Another thing I was thinking about is that some backends (like db, ones) have a ttl for the credentials, and you may need to renew those, I don't see how this is managed here, it's because the code will do polling of the resources? again not very familiar with the code yet.

The `.backend` `.kvVersion` and other configurations could be used as the first backend of the list if exist

spencergibb:
It feels like this is on the right path, but not supporting other backend types seems like missing something basic. I'm going to close this and we can revisit if there is any interest in picking this up.

PR #1278, opened 2019-03-07
Support Google Cloud Source repositories

ewirch:
Fixes gh-1260

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
@@ -64,6 +64,7 @@ import org.springframework.cloud.config.server.environment.SvnKitEnvironmentRepo
 import org.springframework.cloud.config.server.environment.VaultEnvironmentProperties;
 import org.springframework.cloud.config.server.environment.VaultEnvironmentRepository;
 import org.springframework.cloud.config.server.environment.VaultEnvironmentRepositoryFactory;
+import org.springframework.cloud.config.server.support.TransportConfigCallbackFactory;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
@@ -138,8 +139,10 @@ public class EnvironmentRepositoryConfiguration {
 				ConfigurableEnvironment environment, ConfigServerProperties server,
 				Optional<ConfigurableHttpConnectionFactory> jgitHttpConnectionFactory,
 				Optional<TransportConfigCallback> customTransportConfigCallback) {
+			final TransportConfigCallbackFactory transportConfigCallbackFactory = new TransportConfigCallbackFactory(
+					customTransportConfigCallback.orElse(null));
 			return new MultipleJGitEnvironmentRepositoryFactory(environment, server,
-					jgitHttpConnectionFactory, customTransportConfigCallback);
+					jgitHttpConnectionFactory, transportConfigCallbackFactory);
 		}
 
 	}
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
@@ -64,6 +64,7 @@ import org.springframework.cloud.config.server.environment.SvnKitEnvironmentRepo
 import org.springframework.cloud.config.server.environment.VaultEnvironmentProperties;
 import org.springframework.cloud.config.server.environment.VaultEnvironmentRepository;
 import org.springframework.cloud.config.server.environment.VaultEnvironmentRepositoryFactory;
+import org.springframework.cloud.config.server.support.GoogleCloudSourceSupport;
 import org.springframework.cloud.config.server.support.TransportConfigCallbackFactory;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -92,7 +93,8 @@ import org.springframework.jdbc.core.JdbcTemplate;
 		VaultRepositoryConfiguration.class, CredhubConfiguration.class,
 		CredhubRepositoryConfiguration.class, SvnRepositoryConfiguration.class,
 		NativeRepositoryConfiguration.class, GitRepositoryConfiguration.class,
-		RedisRepositoryConfiguration.class, DefaultRepositoryConfiguration.class })
+		RedisRepositoryConfiguration.class, GoogleCloudSourceConfiguration.class,
+		DefaultRepositoryConfiguration.class })
 public class EnvironmentRepositoryConfiguration {
 
 	@Bean
@@ -138,9 +140,11 @@ public class EnvironmentRepositoryConfiguration {
 		public MultipleJGitEnvironmentRepositoryFactory gitEnvironmentRepositoryFactory(
 				ConfigurableEnvironment environment, ConfigServerProperties server,
 				Optional<ConfigurableHttpConnectionFactory> jgitHttpConnectionFactory,
-				Optional<TransportConfigCallback> customTransportConfigCallback) {
+				Optional<TransportConfigCallback> customTransportConfigCallback,
+				Optional<GoogleCloudSourceSupport> googleCloudSourceSupport) {
 			final TransportConfigCallbackFactory transportConfigCallbackFactory = new TransportConfigCallbackFactory(
-					customTransportConfigCallback.orElse(null));
+					customTransportConfigCallback.orElse(null),
+					googleCloudSourceSupport.orElse(null));
 			return new MultipleJGitEnvironmentRepositoryFactory(environment, server,
 					jgitHttpConnectionFactory, transportConfigCallbackFactory);
 		}
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/MultipleJGitEnvironmentRepositoryFactory.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/MultipleJGitEnvironmentRepositoryFactory.java
@@ -18,12 +18,10 @@ package org.springframework.cloud.config.server.environment;
 
 import java.util.Optional;
 
-import org.eclipse.jgit.api.TransportConfigCallback;
 import org.eclipse.jgit.transport.HttpTransport;
 
 import org.springframework.cloud.config.server.config.ConfigServerProperties;
-import org.springframework.cloud.config.server.ssh.FileBasedSshTransportConfigCallback;
-import org.springframework.cloud.config.server.ssh.PropertiesBasedSshTransportConfigCallback;
+import org.springframework.cloud.config.server.support.TransportConfigCallbackFactory;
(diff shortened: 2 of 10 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/MultipleJGitEnvironmentRepositoryFactory.java
```diff
@@ -38,23 +36,23 @@
...
 	@Deprecated
 	public MultipleJGitEnvironmentRepositoryFactory(ConfigurableEnvironment environment,
 			ConfigServerProperties server,
-			Optional<TransportConfigCallback> customTransportConfigCallback) {
-		this(environment, server, Optional.empty(), customTransportConfigCallback);
+			TransportConfigCallbackFactory transportConfigCallbackFactory) {
+		this(environment, server, Optional.empty(), transportConfigCallbackFactory);
 	}
 
 	public MultipleJGitEnvironmentRepositoryFactory(ConfigurableEnvironment environment,
```
ryanjbaxter: For backwards compatibility we should add a new constructor

ewirch: I moved the logic for creation of ssh transport config callbacks to `TransportConfigCallbackFactory`. If I have to preserve the `Optional<TransportConfigCallback>` constructor, either
- `MultipleJGitEnvironmentRepositoryFactory` would have to create `TransportConfigCallbackFactory`, which 1. is mixing concerns and 2. not possible, because `TransportConfigCallbackFactory` additionally has the `GoogleCloudSourceSupport` dependency.
- or: I would have to duplicate the ssh transport config callback creation code, which is 1. bad code duplication, 2. would lead to inconsistent behavior, because `GoogleCloudSourceSupport` would not be handled, when calling old `Optional<TransportConfigCallback>` constructor.
- or: `Optional<TransportConfigCallback>` constructor would only set the transport configuration callback to the `customTransportConfigCallback` (if it is present), ignoring ssh transport config callback and the google cloud source transport config callbacks, which would be - again - inconsistent to the `TransportConfigCallbackFactory` constructor.

My opinion is, that none of these solutions is desirable. But you are the one with more experience. Your call.

ewirch: Ryan? Which approach do you prefer?

ryanjbaxter: Why cant you just keep the current constructor marking it as `@Deprecated` and add a new one with the new parameter?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
```diff
@@ -86,7 +88,7 @@
 		VaultRepositoryConfiguration.class, CredhubConfiguration.class,
 		CredhubRepositoryConfiguration.class, SvnRepositoryConfiguration.class,
 		NativeRepositoryConfiguration.class, GitRepositoryConfiguration.class,
-		DefaultRepositoryConfiguration.class })
+		DefaultRepositoryConfiguration.class, GoogleCloudSourceConfiguration.class })
```
spencergibb: Default needs to be last.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/MultipleJGitEnvironmentRepositoryFactory.java
spencergibb: Spring supports factories as beans rather than the actual bean. Invert control. Leave the constructor as is and create a factory bean rather than injecting the factory.

ewirch: @ryanjbaxter

> Why cant you just keep the current constructor marking it as @Deprecated and add a new one with the new parameter?

The logic before my change was:
 - if there is a bean of type `TransportConfigCallback` in Spring context, it will become the `customTransportConfigCallback`.
 - if there is a `customTransportConfigCallback`, use this bean as transport config callback.
 - otherwise: for all repositories, use `PropertiesBasedSshTransportConfigCallback` or `FileBasedSshTransportConfigCallback`, based on `MultipleJGitEnvironmentProperties.isIgnoreLocalSshSettings()`.

I have to add `GCSTransportConfigCallback` in to the mix without breaking backwards compatibility. Additionally you want me to keep the previous constructors, and add a third one.


**Approach 1**: not extracting transport callback construction logic into a separate class.

I add this constructor
```
public MultipleJGitEnvironmentRepositoryFactory(
    ConfigurableEnvironment environment,
    ConfigServerProperties server,
    Optional<ConfigurableHttpConnectionFactory> connectionFactory,
    Optional<TransportConfigCallback> customTransportConfigCallback,
    Optional<GoogleCloudSourceSupport> googleCloudSourceSupport)
```

Drawbacks:
- Since the other constructors do not receive an instance of `GoogleCloudSourceSupport`, `MultipleJGitEnvironmentRepositoryFactory` will work inconsistently, depending on which constructor you call. If you call one of the old constructors, Google Cloud Source repositories will just not work, even if all other conditions are met.
- `MultipleJGitEnvironmentRepositoryFactory` is polluted with details about repository vendors.
- The logic about choosing the proper transport config callback complicates `MultipleJGitEnvironmentRepositoryFactory`. But `MultipleJGitEnvironmentRepositoryFactory` should not be concerned with this.


**Approach 2**: extracting transport config construction logic into `TransportConfigCallbackFactory`.

I add this constructor
```
public MultipleJGitEnvironmentRepositoryFactory(
    ConfigurableEnvironment environment,
    ConfigServerProperties server,
    Optional<ConfigurableHttpConnectionFactory> connectionFactory,
    TransportConfigCallbackFactory transportConfigCallbackFactory) 
```
and change one of the old constructors to (the other delegates to this one):
```
@Deprecated
public MultipleJGitEnvironmentRepositoryFactory(ConfigurableEnvironment environment,
    ConfigServerProperties server,
    Optional<ConfigurableHttpConnectionFactory> connectionFactory,
    Optional<TransportConfigCallback> customTransportConfigCallback) {

  this(
      environment,
      server,
      connectionFactory,
      new TransportConfigCallbackFactory(customTransportConfigCallback.orElse(null), null)
  );
}
```


Some drawbacks of approach 1 are resolved, but others are added.

Drawbacks:
- `MultipleJGitEnvironmentRepositoryFactory` has to know the exact implementation of `TransportConfigCallbackFactory`, because it has to create an instance of it.
- this remains: Since the other constructors do not receive an instance of `GoogleCloudSourceSupport`, `MultipleJGitEnvironmentRepositoryFactory` will work inconsistently, depending on which constructor you call.


**Approach 3**: copy  transport config construction logic into `TransportConfigCallbackFactory`

I add this constructor
```
public MultipleJGitEnvironmentRepositoryFactory(
    ConfigurableEnvironment environment,
    ConfigServerProperties server,
    Optional<ConfigurableHttpConnectionFactory> connectionFactory,
    TransportConfigCallbackFactory transportConfigCallbackFactory) 
```
keep the other constructors as they are, and also keep the transport config creation logic in `MultipleJGitEnvironmentRepositoryFactory`. It will be used if one of the old constructors is called:

Drawbacks:
- I have to duplicate the code for the transport config creation logic.
- this remains: Since the other constructors do not receive an instance of `GoogleCloudSourceSupport`, `MultipleJGitEnvironmentRepositoryFactory` will work inconsistenly, depending on which constructor you call.


**Solution 4**: instantiate `GCSTransportConfigCallback` in `MultipleJGitEnvironmentRepositoryFactory` (as done for `FileBasedSshTransportConfigCallback` and `PropertiesBasedSshTransportConfigCallback`)

This is not a solution at all. Because dependency to Google library is optional, I cannot hard wire the Google Cloud Source implementation to `MultipleJGitEnvironmentRepositoryFactory`.


From my point of view there is no way to keep the old constructors **and** make `MultipleJGitEnvironmentRepositoryFactory` behave consistently, independent of the constructor you use. Am I missing something? Or do you prefer inconsistent behavior in favor of compile backwards compatibility?


@spencergibb

> create a factory bean rather than injecting the factory

I struggle to see how a factory bean will help me in this situation. 

I could create a factory bean, which will create a Google Cloud Source `TransportConfigCallback`. But this will mess up the legacy logic around `customTransportConfigCallback` (see above).

I could create a factory bean, which will create a `TransportConfigCallbackFactory`. This way `EnvironmentRepositoryConfiguration` will not have to instantiate `TransportConfigCallbackFactory` directly. But since I need a configuration class to tell Spring, that such a factory bean exists, I would just move the `new TransportConfigCallbackFactory()` code from one `@Bean` method to an other `@Bean` method. Not much is won. But on the other hand we well have a factory bean creating a factory, which is weird.

Am I missing your point, @spencergibb?

spencergibb:
See [FactoryBean java doc](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/beans/factory/FactoryBean.html) or [this tutorial](https://www.baeldung.com/spring-factorybean).

Make `TransportConfigCallbackFactory` implement `FactoryBean<TransportConfigCallback>` and restore the original constructor.

ewirch:
@spencergibb 

> Make TransportConfigCallbackFactory implement FactoryBean<TransportConfigCallback>

As soon as there is anything which will make Spring satisfy a bean dependency of `TransportConfigCallback` (be it a bean of type `TransportConfigCallback`, a `javax.inject.Provider<TransportConfigCallback>` or a `FactoryBean<TransportConfigCallback>`) Spring will happily create a bean of this type and populate `Optional<TransportConfigCallback> customTransportConfigCallback` of `EnvironmentRepositoryConfiguration.JGitFactoryConfig.gitEnvironmentRepositoryFactory()`. This will break what I understand as a extension point for library users. 

When google auth library and JGit are on the class path, Spring will populate `customTransportConfigCallback` with the `GCSTransportConfigCallback` (Google Cloud Source), preventing any other customization. If a library user declared a bean of `TransportConfigCallback` on his own, Spring will fail to start, because there are two beans of the same type.

You see, the problem is not getting the Google Cloud Source `TransportConfigCallback` into `MultipleJGitEnvironmentRepositoryFactory`, but also
1. maintain the `customTransportConfigCallback` customization possibility
1. keep all old constructors of `MultipleJGitEnvironmentRepositoryFactory`
1. keep consistent behavior independent of which constructor of `MultipleJGitEnvironmentRepositoryFactory` was called.

I elaborated deeply on this in my long comment above.

I see no way to meet all above requirements at the same time.

spencergibb:
I got to the point where I could make it work with a delegate pattern. The trouble is, the properties are only available at `EnvironmentRepositoryFactory.build()`. I'm also very sorry we introduced `Optional` as parameters :-(

Being that this will only go in a new version, I think I'm ok with this. It will just require some release notes.

I've mentioned a few times that `DefaultRepositoryConfiguration.class` needs to be last.

ewirch:
> I've mentioned a few times that DefaultRepositoryConfiguration.class needs to be last.

I fixed that already. Just didn't push yet, because this would have made no sense as long as the constructor topic was not resolved.

ewirch:
Pushed the `DefaultRepositoryConfiguration` fix. Rebased on top of master.

PR #1298, opened 2019-03-19
Support explicit file name definition for app properties

moudatsos:
spring-cloud-config-server now supports explicit definition of additional
file names that contain properties to be served, in either native or scm
operation mode. The file names are defined by means of the
"spring.cloud.config.name" environment property. Implementation is
applied on NativeEnvironmentRepository (native profile), which is also
used by AbstractScmEnvironmentRepository; it essentially appends the
file names, provided in the "spring.cloud.config.name" environment
property to those already computed by NativeEnvironmentRepository.
(Note that the latter leverages the ConfigFileApplicationListener
functionality by booting a mini app and passing these computed file names
as a value of the "spring.config.name" environment property, in order to
collect their properties).

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
@@ -264,7 +264,7 @@ public class NativeEnvironmentRepository
 	private String[] getArgs(String application, String profile, String label) {
 		List<String> list = new ArrayList<String>();
 		String config = application;
-		String searchNames = this.environment.getProperty("spring.cloud.config.name");
+		String searchNames = this.environment.getProperty("spring.cloud.config.server.searchNames");
 		if (searchNames != null && !searchNames.isEmpty()) {
 			config += "," + searchNames;
 		}
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
@@ -264,6 +264,10 @@ public class NativeEnvironmentRepository
 	private String[] getArgs(String application, String profile, String label) {
 		List<String> list = new ArrayList<String>();
 		String config = application;
+		String searchNames = this.environment.getProperty("spring.cloud.config.name");
+		if (searchNames != null && !searchNames.isEmpty()) {
+			config += "," + searchNames;
+		}
 		if (!config.startsWith("application")) {
 			config = "application," + config;
 		}
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
@@ -264,9 +264,15 @@ public class NativeEnvironmentRepository
 	private String[] getArgs(String application, String profile, String label) {
 		List<String> list = new ArrayList<String>();
 		String config = application;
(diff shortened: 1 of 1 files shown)


spencergibb:
Not sure this makes sense as that is a client side property, not server side.

moudatsos:
> Not sure this makes sense as that is a client side property, not server side.

Does this refer to the name of the property or the provided functionality?

Apparently, the name of the property can change if we feel it might confuse people using this feature (or if a configserver could be itself a config client of another configserver - I admit I haven't checked this out - but I thought that in 1.4.x version this property did not exist).

As far as the functionality is concerned, I would be glad to discuss the added value, in depth & detail. For brevity, I would just recall the all the rellevant and applicable reasons that hold for the implementation of this feature for individual springboot app property sources (via "spring.config.name" environment property)

spencergibb:
I'm unsure of the existing feature you are referring to, can you point me at it?

moudatsos:
> I'm unsure of the existing feature you are referring to, can you point me at it?

Assuming you 're not referring to the "spring.cloud.config.name" case, since you already acknowledged its existence in spring-cloud-config-client project, the "spring.config.name" environment property (defined as `org.springframework.boot.context.config.ConfigFileApplicationListener#CONFIG_NAME_PROPERTY` of spring-boot-1.5.16.RELEASE) is leveraged by `org.springframework.boot.context.config.ConfigFileApplicationListener.Loader#getSearchNames` in order to fetch all the defined application property file names to be used as property sources, in `org.springframework.boot.context.config.ConfigFileApplicationListener.Loader#load()`, called by `org.springframework.boot.context.config.ConfigFileApplicationListener#addPropertySources`, which is invoked on `org.springframework.boot.context.config.ConfigFileApplicationListener#postProcessEnvironment`

Feature is documented at spring boot app **externalized configuration** docs, in [application property files paragraph](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-application-property-files)

One could point out the subtle analogy between the spring boot app "spring.config.name" & "spring.config.location" environment properties use case and the configserver "spring.cloud.config.name" (provided by this patch) & "spring.cloud.config.server.git.searchPaths" or "spring.cloud.config.server.native.searchLocations" (native profile) environment properties use case.

Noteworthy point # 2: `org.springframework.cloud.config.server.environment.NativeEnvironmentRepository` uses this very feature in order to compute its property sources, as I point out in my commit message.

(Apologies if I 've been too verbose, I'm not sure whether you wanted a technical or documentation point-reference - I 'm trying to save you time by providing as much as I see fit)

moudatsos:
It seems that the environment property should be changed to something unique. 

In case both the config client app and the configserver app, are deployed as JEE apps in the same app server, the property scemantics will clash, since the client will use the value as its name when asking for properties from the config server, which will result in wrong properties being served by the server, in case one of the file names coincides with the name of another app, in the configserver data backend (either native or scm).

Since this is a config server side property, we could put it in the same property prefix, as long as it is not causing any problems, e.g.: **spring.cloud.config.server.searchNames** (inspired by `org.springframework.boot.context.config.ConfigFileApplicationListener.Loader#getSearchNames`, which is actually where the defined values are going to be used)

Should this property be included as a corresponding field in `org.springframework.cloud.config.server.config.ConfigServerProperties`?

And last, should I close this pull request, push the property name change to my github repo and open a new pull request for the latest revision, referencing this one? Or should someone else [do something first and then] close this pull request?

moudatsos:
Is there anything I could do to help moving this issue forward?

moudatsos:
I've read on the relative instructions that we should avoid creating both a ticket and a pull request, for the same conceptual issue, since github treats them as the same, but could it be the case that I need to create a ticket for this issue so that it would put it in an active review / processing flow?

Thank you in advance for your answer

spencergibb:
The 1.4.x line is no longer maintained.

spencergibb:

amargJava:
Hey Guys,
Any solution, did you find for the same?

Issue #1283, opened 2019-03-13
 Could not locate PropertySource message not indicate the real issue when a invalid file is pushed to config repository


eldontc:
I've experienced a problem with spring cloud config client when unfortunately pushed an invalid yaml file to my git repository. The message that I've got is: 

``
o.s.c.c.c.ConfigServicePropertySourceLocator - Fetching config from server at: http://my-config-server:8888
ERROR [] o.s.boot.SpringApplication - Application startup failedjava.lang.IllegalStateException: Could not locate PropertySource and the fail fast property is set, failing
``

With this message, I was investigating my problem in the wrong way. Actually, the problem was not the absence of PropertySource. I spent my time to realize that the real problem was an invalid yaml file with the wrong syntax. 

I am using the spring-cloud-config-client:2.0.0.M9

Enhancement
I suggest that the above error message could be more specific, like "Could not load a file on config repository. The file **name-of-file** is invalid". When _name-of-file_ could be the equivalent to spring.application.name + profile. 

A debug flag on configurantion inside the bootstrap.properties could be valueable. Something like:

> spring.cloud.config.debug=true

With default to false

Whe debug is flagged to true, some precious messages could be printed on console, like the files that client is trying to get,

pmerson:
+1 for fixing this issue. Just spent a lot of time to find out that my yml properties file had a duplicate key. The error message from the spring cloud server has no useful information.

nhomble:
This burned me for a bit as well so I'd be interested in helping out with this. Is there any reason not to wrap the ```loader.load``` bit in ```ConfigFileApplicationListener::592``` in some custom exception like ```FailedToLoadResourceException``` that could be caught in a controller advice?

ryanjbaxter:
@nhomble if you would like to help, submit a PR with your proposed solution and we can review it

spencergibb:
`ConfigFileApplicationListener` is part of spring boot, not spring cloud.

nhomble:
Oh good catch I was lazily looking at the stacktrace from a repro I got on my branch. 

I'll happily follow up with a proposal!

nhomble:
https://github.com/nhomble/spring-cloud-config/tree/feature/gh-1283

[...]

PR #1341, opened 2019-04-02
Better error messages

nhomble:
work in process to resolve GH-1283


nhomble:
@ryanjbaxter I just wanted to quickly propose something in case there were any concerns with where and how I am catching the exception. I will polish the branch asap.

PR #1350, opened 2019-04-09
code changes for Vault role id and secret id

kamalakarp:
code changes for Vault role id and secret id

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
@@ -47,6 +47,16 @@ public class ConfigClientProperties {
 	public static final String TOKEN_HEADER = "X-Config-Token";
 
 	/**
+	 * Role Id header name.
+	 */
+	public static final String APP_ROLE_ID_HEADER = "X-Config-Role-Id";
+
+	/**
+	 * Secret id header name.
+	 */
+	public static final String APP_SECRET_ID_HEADER = "X-Config-Secret-Id";
+
+	/**
 	 * State header name.
 	 */
 	public static final String STATE_HEADER = "X-Config-State";
@@ -118,6 +128,10 @@ public class ConfigClientProperties {
 	 * Flag to indicate whether to send state. Default true.
 	 */
 	private boolean sendState = true;
+
+	private String roleId;
+
+	private String secretId;
 
 	/**
 	 * Additional headers used to create the client request.
@@ -241,6 +255,22 @@ public class ConfigClientProperties {
 
 	public void setHeaders(Map<String, String> headers) {
 		this.headers = headers;
+	}
+
+	public String getRoleId() {
+		return roleId;
+	}
+
+	public void setRoleId(String roleId) {
+		this.roleId = roleId;
+	}
+
+	public String getSecretId() {
+		return secretId;
+	}
+
+	public void setSecretId(String secretId) {
+		this.secretId = secretId;
 	}
 
 	private Credentials extractCredentials(int index) {
--- a/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/BaseDiscoveryClientConfigServiceBootstrapConfigurationTests.java
+++ b/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/BaseDiscoveryClientConfigServiceBootstrapConfigurationTests.java
@@ -41,6 +41,10 @@ import static org.springframework.cloud.config.client.ConfigClientProperties.Dis
 
 public abstract class BaseDiscoveryClientConfigServiceBootstrapConfigurationTests {
 
+	/**
+	 * @author ubuntu
+	 *
+	 */
 	@Rule
 	public ExpectedException expectedException = ExpectedException.none();
 
--- a/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientPropertiesTests.java
+++ b/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientPropertiesTests.java
@@ -34,6 +34,9 @@ import static org.assertj.core.api.Assertions.assertThat;
  */
 public class ConfigClientPropertiesTests {
 
+	/**
+	 * @author Kamalakar Ponaka
+	 */
 	@Rule
 	public ExpectedException expected = ExpectedException.none();
 
--- a/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocatorTests.java
+++ b/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocatorTests.java
@@ -57,6 +57,10 @@ import static org.springframework.cloud.config.client.ConfigClientProperties.AUT
 
 public class ConfigServicePropertySourceLocatorTests {
 
+	/**
+	 * @author ubuntu
+	 *
+	 */
 	@Rule
 	public ExpectedException expected = ExpectedException.none();
 
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -53,13 +53,10 @@ import org.springframework.web.client.HttpServerErrorException;
 import org.springframework.web.client.ResourceAccessException;
 import org.springframework.web.client.RestTemplate;
 
-import static org.springframework.cloud.config.client.ConfigClientProperties.AUTHORIZATION;
-import static org.springframework.cloud.config.client.ConfigClientProperties.STATE_HEADER;
-import static org.springframework.cloud.config.client.ConfigClientProperties.TOKEN_HEADER;
-
 /**
  * @author Dave Syer
  * @author Mathieu Ouellet
+ * @author Kamalakar Ponaka
  *
  */
 @Order(0)
@@ -183,6 +180,8 @@ public class ConfigServicePropertySourceLocator implements PropertySourceLocator
 		String name = properties.getName();
 		String profile = properties.getProfile();
 		String token = properties.getToken();
+		String roleId = properties.getRoleId();
+		String secretId = properties.getSecretId();
 		int noOfUrls = properties.getUri().length;
 		if (noOfUrls > 1) {
 			logger.info("Multiple Config Server Urls found listed.");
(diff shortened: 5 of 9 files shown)


spencergibb:
Can you describe the changes beyond "code changes"?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
```diff
@@ -46,6 +46,16 @@
 	 */
 	public static final String TOKEN_HEADER = "X-Config-Token";
 
+	/**
+	 * Role Id header name.
+	 */
+	public static final String APP_ROLE_ID_HEADER = "X-Config-Role-Id";
```
spencergibb: `X-Config-Token` was opaque and can be used for other back ends. Adding vault specific tokens in the client is not a good idea.

spencergibb:
I'd rather you stop working on this to discuss your needs. Vault details can't leak into the config client.

kamalakarp:
Spencer, Here is the Change request details.

Current Config Server implementation depends upon the Vault token to communicate with HashiCorp. The token approach doesn't work across the multi cluster environment and move towards the AppRole - Role ID and Secret ID which gets replicated across multiple Vault Clusters.

spencergibb:
I understand. The token head in the client is NOT specifically the vault token. The vault env repository is currently the only implementation to use it. I won't leak vault specific properties into the config client. We have to come up with a better way.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
spencergibb: There is already a `headers` field. Why not make use of that?

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/BaseDiscoveryClientConfigServiceBootstrapConfigurationTests.java
```diff
@@ -41,6 +41,10 @@
 
 public abstract class BaseDiscoveryClientConfigServiceBootstrapConfigurationTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocatorTests.java
```diff
@@ -57,6 +57,10 @@
 
 public class ConfigServicePropertySourceLocatorTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientPropertiesTests.java
```diff
@@ -34,6 +34,9 @@
  */
 public class ConfigClientPropertiesTests {
 
+	/**
+	 * @author Kamalakar Ponaka
```
spencergibb: Remove

kamalakarp:
I'm open to ideas if we don't want to leak Vault specific properties. if there is a better way to pass AppRole - Role ID and Secret ID from config client to config server , i can modify the code

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/NativeBootstrapFailureAnalyzerTests.java
```diff
@@ -30,8 +30,13 @@
 /**
  * @author Ryan Baxter
  */
+
 public class NativeBootstrapFailureAnalyzerTests {
 
+	/**
+	 * @author Ryan Baxter
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/composite/CompositUtilsTests.java
```diff
@@ -29,6 +29,10 @@
 
 public class CompositUtilsTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/config/HttpClientVaultRestTemplateFactoryTest.java
```diff
@@ -72,6 +72,10 @@
 		HTTPS_PROXY.setPort(8081);
 	}
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/ConfigurableHttpConnectionFactoryIntegrationTests.java
```diff
@@ -93,7 +93,10 @@
 		HTTPS_PROXY.setHost("http://https.proxy");
 		HTTPS_PROXY.setPort(8081);
 	}
-
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/EnvironmentControllerTests.java
```diff
@@ -47,6 +47,10 @@
  */
 public class EnvironmentControllerTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/EnvironmentEncryptorEnvironmentRepositoryTests.java
```diff
@@ -37,6 +37,10 @@
  */
 public class EnvironmentEncryptorEnvironmentRepositoryTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepositoryTests.java
```diff
@@ -89,6 +89,10 @@
  */
 public class JGitEnvironmentRepositoryTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/MultipleJGitEnvironmentRepositoryIntegrationTests.java
```diff
@@ -52,6 +52,10 @@
  */
 public class MultipleJGitEnvironmentRepositoryIntegrationTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/MultipleJGitEnvironmentRepositoryTests.java
```diff
@@ -47,6 +47,10 @@
  */
 public class MultipleJGitEnvironmentRepositoryTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryIntegrationTests.java
```diff
@@ -50,6 +50,10 @@
 		"server.key-alias=ssl-test" })
 public class VaultEnvironmentRepositoryIntegrationTests {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/support/HttpClientSupportTest.java
```diff
@@ -45,6 +45,10 @@
 @SpringBootTest(classes = HttpClientSupportTest.TestConfiguration.class, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 public class HttpClientSupportTest {
 
+	/**
+	 * @author ubuntu
```
spencergibb: Remove

kamalakarp:
@spencergibb - maven-checkstyle-plugin fails the build of i don't add the comment in the test cases.
[ERROR] /home/ubuntu/spring-cloud-config/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryIntegrationTests.java:53:2: Missing a Javadoc comment. [JavadocVariable]

spencergibb:
Run with `-Ddisable.checks=true`. It makes reviewing a PR difficult.

kamalakarp:
@spencergibb : Will this yml format work for secret-id and role-id - if i use headers.

spring:
  application:
    name: configclient
  cloud:
    config:
      token: myroot
      uri: https://localhost:8888
      headers: 
        role-id: 12345
        secret-id: abc

spencergibb:
It will, the headers are taken directly from there. So if you want to match the server side, you'll need `X-Config-Role-Id`, etc... otherwise it would just be `role-id`.

kamalakarp:
let me work on these changes and will keep you posted

spencergibb:
Looks like you changed the line endings for a bunch of files. You need to change them back

kamalakarp:
I'm working on t . I'm somehow not able to retrieve the role id and secret id from headers.

mp911de:
What's the reason to not use [authentication method implementations](https://github.com/spring-projects/spring-vault/blob/master/spring-vault-core/src/main/java/org/springframework/vault/authentication/AppIdAuthentication.java) provided by Spring Vault? They are tested integratively against Vault. Spring Vault has already support for all suitable machine to machine auth methods so there's likely little sense in duplicating efforts.

Initially, we decided to not integrate with Spring Vault because pulling in a dependency to abstract `RestTemplate` away didn't make sense. With supporting various auth methods and Key-Value v1/v2 backends, we might want to reconsider that decision.

kamalakarp:
@mp911de Spring Vault only works for Java implementations. Config Server can be used in a language agnostic way. E.g. Config Server is available as a market place offering PCF (Pivotal Cloud Foundry) which can be bind to any app.

mp911de:
Aren't we discussing changes to Java code for the server part here, @kamalakarp?

kamalakarp:
@mp911de As far as i know, current spring vault doesn't support namespaces which are available in licensed version of Vault. https://learn.hashicorp.com/vault/operations/namespaces
If Config Server can be integrated with Spring Vault, i'm totally fine.

spencergibb:
@kamalakarp https://github.com/spring-projects/spring-vault/issues/346

kamalakarp:
> @kamalakarp [spring-projects/spring-vault#346](https://github.com/spring-projects/spring-vault/issues/346)

That's cool

kamalakarp:
i have one question though, currently with the config server, i use annotations with key names to get the value but does spring vault does the same way ?

Config Server :

    @Value("${client.pseudo.property}")
    private String pseudoProperty;

kamalakarp:
@spencergibb @csterwa - I have made all the changes needed for role_id and secret_id changes.
I have tested all the changes and they are working fine.

```
spring:
  profiles:
    active: vault
  application:
    name: configclient
  cloud:
    config:
#      token: s.sTZMTecnExpuplVggKKwvwyq
#      uri: https://my-config-server.ausmpc.pcf.x.com
      uri: http://localhost:8888
      headers:
        role_id: 4c0ec2ae-6467-6d48-e33b-318660655383
        secret_id: 60b07a6e-c124-33ed-148f-11394972c13f

```

kamalakarp:
> There are some big problems with this PR now. Looks like there are significant whitespaces changes in every file to the point where I can't see what has changed. It also needs to be rebased as there are many merge conflicts.

I'll work on the changes and keep you posted

PR #1385, opened 2019-05-16
Allow file:// URLs for file monitoring in git profile

### LINKED ISSUE #1374: Filemonitor does not work for git repositories with file:// url
I configured `spring.cloud.config.server.git.uri: file://c:/test/`, with C:/test/ being a local git repo.
I expected the file monitor to watch this git repo for changes.

In https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/FileMonitorConfiguration.java#L197 the repository URL gets loaded as a `Resource`. File watching will only be enabled when `resource instanceof FileSystemResource`, but the Default ResourceLoader will never return a FileSystemResource, only a FileUrlResource, so file monitoring does not work for a file:// url. As long as the default ResourceLoader is used, the instanceof check can never be true. In case of the native Repositories, the searchlocations URL(s) are loaded as resources but only checked if they exist (https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/FileMonitorConfiguration.java#L211).

I could only fix that by providing a ProtocolResolver that uses a FileSystemResourceLoader for URLs starting with file://, thus returning a FileSystemResource for such URLs as described here: https://stackoverflow.com/a/51420272/727373

I think, either the check on the SCM repository URLs should be made less restrictive (why is this check more strict as the check for the native repository URL?), or at least documentation should be improved to help providing the necessary ProtocolResolver.

dl1ely:
Fixes #1374

### DIFF
--- a/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/FileMonitorConfiguration.java
+++ b/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/FileMonitorConfiguration.java
@@ -195,7 +195,8 @@ public class FileMonitorConfiguration implements SmartLifecycle, ResourceLoaderA
 				for (AbstractScmEnvironmentRepository repository : scmRepositories) {
 					repositoryUri = repository.getUri();
 					Resource resource = this.resourceLoader.getResource(repositoryUri);
-					if (resource instanceof FileSystemResource || resource instanceof FileUrlResource) {
+					if (resource instanceof FileSystemResource
+							|| resource instanceof FileUrlResource) {
 						paths.add(Paths.get(resource.getURI()));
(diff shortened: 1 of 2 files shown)


dl1ely:
Will do.

dl1ely:
I added tests for the single SCM Repo case with a file:// URL and another test for the case with two SCM repos, one using a file path and the other using a file:// URL. Sorry for the delay, but i was on holidays and very busy in the mean time.

dl1ely:
Sorry, i do not know much about codecov, what exactly is the issue with the failing codecov/patch check? What should i do?

Issue #1378, opened 2019-05-10
ConfigClientWatcher with Git backend


dlsrb6342:
With git backend, as I know, Environment.getVersion() return HEAD Commit's checksum.
But ConfigClientWatcher just watch "config.client.state" value.

If ConfigClientWatcher watch "config.client.version", It would be works well with git without any config.
How do you think about this? I would like to make a implementation based on this.

ryanjbaxter:
What problem are you running into with the current implementation?
Can you provide a [complete, minimal, verifiable sample](https://stackoverflow.com/help/mcve) that reproduces the problem? It should be available as a GitHub (or similar) project or attached to this issue as a zip file.

dlsrb6342:
It doesn't have any other config or code. I just run one spring-cloud-config-server and one client.
my client project's bootstrap.yml 
```
spring:
  application:
    name: client
  profiles:
    active: local

  cloud:
    config:
      uri: http://localhost:8081
      fail-fast: true
      watch:
        enabled: true
```

and my server project's application.yml
```
server:
  port: 8081

spring:
  cloud:
    config:
      server:
        git:
          uri: https://github.com/dlsrb6342/configs.git
          clone-on-start: true
          username: USERNAME
          password: PASSWORD
```

spring.cloud.config.watch.enabled is set true for ConfigClientWatch.
ConfigClientWatch check state by below code.
```
String newState = this.environment.getProperty("config.client.state");
String oldState = ConfigClientStateHolder.getState();
```

With github as backend, spring-cloud-config-server doesn't send state. 
Because of this, newState and oldState are always NULL. so it never refresh.

But cloud-config-server with github, it send **version**.
**version** means git HEAD commit's checksum. So I can recognize whether the config repo is updated by compare newVersion and oldVersion.

So I think ClientConfigWatch should be watching **version**.

dlsrb6342:
And also, if config-server send **state**, there will be two refresh job.

If watch is enabled and refresh endpoint is enabled,
The first refresh job is caused by /actuator/refresh call. It refresh with new config.

And second one is caused by ConfigClientWatch. `String newState = this.environment.getProperty("config.client.state");` would be changed by first refresh job, so it cause `this.refresher.refresh()`.
https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientWatch.java#L62:L73

[...]

PR #1390, opened 2019-05-21
Support ConfigClientWatch for 'config.client.version' property

dlsrb6342:
Fixes gh-1378

- `ConfigClientWatch` is renamed as `ConfigClientStateWatch`.
- `ConfigClientWatch` is added as interface.
- `ConfigClientVersionWatch` is watching `config.clientversion` property. it works well with git backend.


spencergibb:
Thanks for the PR. In my comment https://github.com/spring-cloud/spring-cloud-config/issues/1378#issuecomment-492062463 I wasn't sure we wanted to support something like this. I'm going to close this and if we decide otherwise we can re-open.

PR #1398, opened 2019-05-30
Add abstraction for retrieving config token.

### LINKED ISSUE #1393: VaultEnvironmentRepository doesn't load with embedded config server
When embedding config server in a Spring Boot application as described in [the docs](https://cloud.spring.io/spring-cloud-static/Greenwich.SR1/single/spring-cloud.html#_embedding_the_config_server), and configuring a `vault` profile, the app will fail with an exception at startup: 

```
2019-05-22 09:10:36.949 ERROR 79613 --- [           main] o.s.boot.SpringApplication               : Application run failed

java.lang.IllegalStateException: No HttpServletRequest available
	at org.springframework.cloud.config.server.environment.VaultEnvironmentRepository.findOne(VaultEnvironmentRepository.java:129) ~[spring-cloud-config-server-2.1.1.RELEASE.jar:2.1.1.RELEASE]
	at org.springframework.cloud.config.server.environment.CompositeEnvironmentRepository.findOne(CompositeEnvironmentRepository.java:61) ~[spring-cloud-config-server-2.1.1.RELEASE.jar:2.1.1.RELEASE]
	at org.springframework.cloud.config.server.environment.EnvironmentRepositoryPropertySourceLocator.locate(EnvironmentRepositoryPropertySourceLocator.java:55) ~[spring-cloud-config-server-2.1.1.RELEASE.jar:2.1.1.RELEASE]
	at org.springframework.cloud.bootstrap.config.PropertySourceBootstrapConfiguration.initialize(PropertySourceBootstrapConfiguration.java:97) ~[spring-cloud-context-2.1.1.RELEASE.jar:2.1.1.RELEASE]
	at org.springframework.boot.SpringApplication.applyInitializers(SpringApplication.java:649) ~[spring-boot-2.1.4.RELEASE.jar:2.1.4.RELEASE]
	at org.springframework.boot.SpringApplication.prepareContext(SpringApplication.java:373) ~[spring-boot-2.1.4.RELEASE.jar:2.1.4.RELEASE]
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:314) ~[spring-boot-2.1.4.RELEASE.jar:2.1.4.RELEASE]
	at org.springframework.boot.builder.SpringApplicationBuilder.run(SpringApplicationBuilder.java:139) [spring-boot-2.1.4.RELEASE.jar:2.1.4.RELEASE]
	at org.springframework.boot.builder.SpringApplicationBuilder$run$0.call(Unknown Source) [spring-boot-2.1.4.RELEASE.jar:2.1.4.RELEASE]
```

This is apparently due to the fact that the `VaultEnvironmentRepository` requires an active `HttpServletRequest` so that it can extract the `X-Config-State` and `X-Config-Token` headers from the HTTP request sent to config server, and there is no HTTP request active when the app bootstraps configuration in the embedded mode.

scottfrederick:
The VaultEnvironmentRepository requires a token to be passed in a X-Config-Token header in an HttpServletRequest. This doesn't work well when config server is embedded in an application, as an HttpServletRequest is not available in that case.

These changes allows the Vault token to be provided in the HttpServletRequest header (the default behavior) or as a configuration property.

Fixes #1393

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepository.java
@@ -39,13 +39,13 @@ import org.springframework.validation.annotation.Validated;
 import org.springframework.web.client.RestTemplate;
 
 import static org.springframework.cloud.config.client.ConfigClientProperties.STATE_HEADER;
-import static org.springframework.cloud.config.client.ConfigClientProperties.TOKEN_HEADER;
 
 /**
  * @author Spencer Gibb
  * @author Mark Paluch
  * @author Haroun Pacquee
  * @author Haytham Mohamed
+ * @author Scott Frederick
  */
 @Validated
 public class VaultEnvironmentRepository implements EnvironmentRepository, Ordered {
@@ -53,12 +53,12 @@ public class VaultEnvironmentRepository implements EnvironmentRepository, Ordere
 	/**
 	 * Vault token header name.
 	 */
-	public static final String VAULT_TOKEN = "X-Vault-Token";
+	private static final String VAULT_TOKEN = "X-Vault-Token";
 
 	/**
 	 * Vault namespace header name.
 	 */
-	public static final String VAULT_NAMESPACE = "X-Vault-Namespace";
+	static final String VAULT_NAMESPACE = "X-Vault-Namespace";
 
 	/** Vault host. Defaults to 127.0.0.1. */
 	@NotEmpty
@@ -94,15 +94,25 @@ public class VaultEnvironmentRepository implements EnvironmentRepository, Ordere
 	private VaultKvAccessStrategy accessStrategy;
 
 	// TODO: move to watchState:String on findOne?
-	private ObjectProvider<HttpServletRequest> request;
+	private final ObjectProvider<HttpServletRequest> request;
 
-	private EnvironmentWatch watch;
+	private final EnvironmentWatch watch;
+
+	private final ConfigTokenProvider tokenProvider;
 
 	public VaultEnvironmentRepository(ObjectProvider<HttpServletRequest> request,
 			EnvironmentWatch watch, RestTemplate rest,
 			VaultEnvironmentProperties properties) {
+		this(request, watch, rest, properties,
+				new HttpRequestConfigTokenProvider(request));
+	}
+
+	public VaultEnvironmentRepository(ObjectProvider<HttpServletRequest> request,
+			EnvironmentWatch watch, RestTemplate rest,
+			VaultEnvironmentProperties properties, ConfigTokenProvider tokenProvider) {
 		this.request = request;
 		this.watch = watch;
+		this.tokenProvider = tokenProvider;
 		this.backend = properties.getBackend();
 		this.defaultKey = properties.getDefaultKey();
 		this.host = properties.getHost();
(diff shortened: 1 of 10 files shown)


scottfrederick:
> We should probably add new constructors.

@ryanjbaxter Do you think both `VaultEnvironmentRepository` and `VaultEnvironmentRepositoryFactory` need to have backward-compatible constructors? I'll add those and update the PR.

> Also should we consider making these changes in the 2.0.x stream?

I can create an additional PRs for `master` and see what's different in `2.0.x`.

scottfrederick:
@ryanjbaxter PR updated with backward-compatible ctors.

ryanjbaxter:
@scottfrederick if you submit it to 2.0.x stream we can usually just merge it forward without much of an issue

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepository.java
```diff
@@ -116,6 +124,13 @@ public VaultEnvironmentRepository(ObjectProvider<HttpServletRequest> request,
 
 		this.accessStrategy = VaultKvAccessStrategyFactory.forVersion(rest, baseUrl,
 				properties.getKvVersion());
+
+		if (tokenProvider == null) {
```
ryanjbaxter: Instead of passing `null` in the old constructor why not just pass a new `HttpRequestConfigTokenProvider`?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryFactory.java
```diff
@@ -25,22 +25,36 @@
...
+
+	private final ConfigTokenProvider tokenProvider;
 
 	public VaultEnvironmentRepositoryFactory(ObjectProvider<HttpServletRequest> request,
 			EnvironmentWatch watch,
 			Optional<VaultRestTemplateFactory> vaultRestTemplateFactory) {
 		this.request = request;
 		this.watch = watch;
 		this.vaultRestTemplateFactory = vaultRestTemplateFactory;
+		this.tokenProvider = null;
```
ryanjbaxter: Can we just use the http token provider by default here?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepository.java
scottfrederick: You're right, that's better. I'll update it.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryFactory.java
scottfrederick: I was trying to avoid duplicating that defaulting logic in both `VaultEnvironmentRepository` and `VaultEnvironmentRepositoryFactory`, and it has to be in the former. Not a huge amount of duplication so I'm happy to do it either way.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepository.java
scottfrederick: Done

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentRepositoryFactory.java
scottfrederick: Done

Issue #865, opened 2017-12-04
Encrypted properties are not decrypted in plain text files


merusso:
The Spring Cloud Config Server is not decrypting values when serving plain text config files. Is this known/expected behavior?

Let's say I have a property file with an encrypted value, like so:
> app.properties
```properties
# "mysecret"
spring.datasource.password = {cypher}682bc583f4641835fa2db009355293665d2647dade
```
If I try to access this property using "standard" (JSON) or "alternative format" (.properties,.yaml), this value is correctly decrypted by the config server.
> GET /app/profile/branch
```json
{
  "name": "app",
  "profiles": [
    "profile"
  ],
  "label": "branch",
  "version": "1839be51feaa73138feef1ba8653bfe7a4c21fb3",
  "state": null,
  "propertySources": [
    {
      "name": "ssh://git-server/app.git/app.properties",
      "source": {
        "spring.datasource.password": "mysecret"
      }
    }
  ]
}
```

> GET /branch/app-profile.properties
```properties
spring.datasource.password = mysecret
```
However, if I attempt to access this file using the "plain text" format, the value is not being decrypted.

> GET /app/profile/branch/app.properties
```properties
# "mysecret"
spring.datasource.password = {cypher}682bc583f4641835fa2db009355293665d2647dade
```
Similarly, if I have an encrypted value stored in another text file format (.conf for instance), this value will not be decrypted when requesting this file via the config server.

> GET /app/profile/branch/app.conf
```
{
    password    {cypher}682bc583f4641835fa2db009355293665d2647dade
}
```

ryanjbaxter:
When you hit  `/app/profile/branch/app.properties` we are just serving up the raw file and we don't decrypt encrypted values in that case.

bplies-IAS:
While just attempting an upgrade of spring-cloud-config-server from  1.2.2.RELEASE to 1.3.2.RELEASE it also stopped decrypting properties when calling the basic 
`$ curl http://user:password@localhost:PORT/APP-NAME/PROFILE/LABEL`

Using the same JKS store and everything.

bplies-IAS:
I've done some debugging and have been able to reproduce the problem and narrow down what I believe is the cause.

Spring Boot:  1.5.9
Spring Cloud Config Server: 1.3.3.RELEASE 
Cloud Dependencies:  Dalston.SR4
(Matches **Release Trains** described [here](http://projects.spring.io/spring-cloud/))

[...]

PR #1417, opened 2019-07-03
Implement plain text decryption. (spring-cloud#865)

sstiglitz:
### What does this PR do?

This implements the feature request made in GH-865. It adds support to decrypt properties being served as plain text through the ResourcesController. Only JSON, YAML, and Properties file extensions are
supported.

### Design considerations.

- On/off of this feature is through setting both `spring.cloud.config.server.encrypt.enabled` (existing setting) and `spring.cloud.config.server.encrypt.plainTextEncrypt` (entirely new setting) to `true`. By default, plainTextEncrypt is `false`. This is to ensure backwards compatibility and not break existing users which may be expecting encrypted values. The dual enable is a little clunky to me, but couldn't think of a better way given existing features. `spring.cloud.config.server.encrypt.enabled` set to `false` is supposed to turn of decryption entirely so a dependency on this setting seemed right to not break the contract. Please let me know your thoughts.

- Notice I say **file extensions** above. Not file types. For example, a file with a ".json" extension is assumed to be JSON. If it's just text, the system will end up throwing an error due to not being able to parse JSON. I figured a proper file type check was unnecessary given the precedent already being set within EnvironmentController and this being a system-to-system application targeted at developers/operations who can be trusted to maintain consistency between file extensions and file types.

- If a file extension is unsupported, but this new feature is enabled, the unencrypted file will be served and WARN log event will trigger. I see-saw'ed over throwing an exception, but that didn't seem right as a person may want decrypted values when supported, but the untouched file when unsupported. I wasn't sure about the log event either as it might fill up log files on a busy server. Doing nothing at all seems like an equally okay option, but please let me know your thoughts.    

- There is a new dependency to handle YAML parsing. It's using a Jackson YAML module. SnakeYml has similar parsing capabilities, but using Jackson offered good code reuse in an environment already using Jackson, so this seemed like a better choice.


pivotal-issuemaster:
@sstiglitz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1417)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1417) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sstiglitz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1417)!

ryanjbaxter:
@sstiglitz could you update the PR to resolve the merge conflicts?

ryanjbaxter:
ping @sstiglitz

ryanjbaxter:
Closed via https://github.com/spring-cloud/spring-cloud-config/commit/8025d7d17e20f2c1116d056e4ef1b076531a568b

Issue #654, opened 2017-02-26
map AbstractScmEnvironmentRepository semantics to S3 versions


mwnorman:
? spring-cloud-config-serverless-s3 ?
Instead of requiring a git server in the production environment, would it be possible to map
the semantics of AbstractScmEnvironmentRepository to S3 versioning?

I see in the Spring Cloud Incubator an example of a database-backed config server (MongoDB)
and somewhere else I saw a discussion of Cassandra; however, I would like to examine
the possibility of going "serverless", using AWS S3 (perhaps need Lambda?) object versioning.

Any help/hints would be greatly appreciated,
---
Mike Norman

spencergibb:
Not sure `AbstractScmEnvironmentRepository` would be ideal. You'd have to experiment with the S3 apis. I think in general using S3 or other storage services could be useful.

marcellodesales:
@mwnorman We have developed a proprietary `serverless` solution called `Config Publisher` that publishes all known/required `app/profiles` to S3. That allows our customers use Hystrix-enabled clients against S3-compatible storage services plus the Config Service itself. We may Open-source it this year... 

Just `HTTP PUT` the endpoint `${config-server-host}/[${label}/]${appName}/{profiles}` to S3 as is and u can use an app with the Config Client against that endpoint.

mwnorman:
I started my implementation: spring-cloud-config-s3: it depends on spring-cloud-config-server and aws-java-sdk-s3 and has an auto-config class in META-INF/spring.factories:
`org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
org.springframework.cloud.config.s3.CloudConfigS3AutoConfiguration`
which looks like
```
@Configuration
@ConditionalOnBean(CloudConfigS3Configuration.Marker.class)
@EnableConfigurationProperties(CloudConfigS3Properties.class)
@Import({S3EnvironmentRepositoryConfiguration.class})
public class CloudConfigS3AutoConfiguration {}
```
The S3EnvironmentRepositoryConfiguration class looks like
```
@Configuration
public class S3EnvironmentRepositoryConfiguration {
    @Autowired
    protected ConfigurableEnvironment environment;
    @Autowired
    protected CloudConfigS3Properties s3Properties;
    @Bean
    @Primary
    public EnvironmentRepository environmentRepository() {
        return new S3EnvironmentRepository(environment, s3Properties);
    }
    @Bean
[...]

PR #1425, opened 2019-07-15
Add AWS S3 Environment Repository

claymccoy:
This adds s3 as an Environment Repository.
This is based on code linked here:
https://github.com/spring-cloud/spring-cloud-config/issues/654
though it has been significantly reworked and tested.
It has been tested against vanilla buckets, versioned buckets, encrypted buckets with both AES-256 and AWS-KMS.

There are several things that I am not sure I have done 100% correctly. I'm happy to take feedback and fix. Here is a list of specific things that are known unknowns for me.
* Is there anything else that I need to do to make this implementation available in addition to what I added in EnvironmentRespositoryConfiguration?
* I added the AWS s3 java dependency for obvious reasons. There was already an AWS core dependency so I just used the same version. It is a very old version thought (2016). It seems to get the job done, but there is some stuff for STS credentials that I might use in the newer AWS SDK. I'm unsure what the process for testing the other code that requires AWS would be and updating.
* The AWS S3 client is constructed using the Default AWS provider chain. This should handle most cases implicitly (including assume role), though a newer AWS SDK might be preferable.
* This currently works against typical <app>-<profile>.properties and <app>-<profile>.yml files that are in a single AWS account and in a specified bucket. Label is optionally used as another suffix on the file name (<app>-<profile>-<label>.yml). This seems like a straight forward mapping of config server concepts to S3, but I am happy to hear other ideas. I noticed that the Git impl allows multiple repos and a kind of routing layer. This is a possibility later if needed, but this seems like a good place to start.
* I think the yaml and properties are parsed into a map correctly, but have a look at the unit tests to ensure.
* I'm trying to handle exceptional cases like other implementations when there are problems getting the config.
* In the factory I pass in ConfigServerProperties and make use of several of its attributes.
* I added the new test to AdhocTestSuite.
* I've tried serveral things, but I don't have the dependencies resolving correctly in IntelliJ Idea. I can compile and run tests with mvn on the command line, but it would be great to be able to do those things in IJ. I think that it gets confused with the parent pom in spring-cloud-build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,11 @@
 				<version>1.11.52</version>
 			</dependency>
 			<dependency>
+				<groupId>com.amazonaws</groupId>
+				<artifactId>aws-java-sdk-s3</artifactId>
+				<version>1.11.52</version>
+			</dependency>
+			<dependency>
 				<groupId>com.google.auth</groupId>
 				<artifactId>google-auth-library-oauth2-http</artifactId>
 				<version>0.15.0</version>
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
@@ -21,6 +21,7 @@ import java.util.Optional;
 
 import javax.servlet.http.HttpServletRequest;
 
+import com.amazonaws.auth.AWSCredentials;
 import org.apache.http.client.HttpClient;
 import org.eclipse.jgit.api.TransportConfigCallback;
 import org.tmatesoft.svn.core.SVNException;
@@ -35,6 +36,9 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.cloud.config.server.composite.CompositeEnvironmentBeanFactoryPostProcessor;
 import org.springframework.cloud.config.server.composite.ConditionalOnMissingSearchPathLocator;
 import org.springframework.cloud.config.server.composite.ConditionalOnSearchPathLocator;
+import org.springframework.cloud.config.server.environment.AwsS3EnvironmentProperties;
+import org.springframework.cloud.config.server.environment.AwsS3EnvironmentRepository;
+import org.springframework.cloud.config.server.environment.AwsS3EnvironmentRepositoryFactory;
 import org.springframework.cloud.config.server.environment.CompositeEnvironmentRepository;
 import org.springframework.cloud.config.server.environment.ConfigTokenProvider;
 import org.springframework.cloud.config.server.environment.ConfigurableHttpConnectionFactory;
@@ -179,6 +183,18 @@ public class EnvironmentRepositoryConfiguration {
 	}
 
 	@Configuration
+	@ConditionalOnClass(AWSCredentials.class)
+	static class AwsS3FactoryConfig {
+
+		@Bean
+		public AwsS3EnvironmentRepositoryFactory awsS3EnvironmentRepositoryFactory(
+				ConfigServerProperties server) {
+			return new AwsS3EnvironmentRepositoryFactory(server);
+		}
+
+	}
+
+	@Configuration
 	@ConditionalOnClass(SVNException.class)
 	static class SvnFactoryConfig {
 
@@ -298,6 +314,19 @@ class NativeRepositoryConfiguration {
 @Configuration
 @Profile("git")
 class GitRepositoryConfiguration extends DefaultRepositoryConfiguration {
+
+}
+
+@Configuration
+@Profile("s3")
+class AwsS3RepositoryConfiguration {
+
+	@Bean
+	public AwsS3EnvironmentRepository awsS3EnvironmentRepository(
+			AwsS3EnvironmentRepositoryFactory factory,
+			AwsS3EnvironmentProperties environmentProperties) {
+		return factory.build(environmentProperties);
+	}
 
 }
 
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
@@ -1,0 +1,217 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.config.server.environment;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.model.GetObjectRequest;
+import com.amazonaws.services.s3.model.S3Object;
+import com.amazonaws.services.s3.model.S3ObjectIdBuilder;
+import org.yaml.snakeyaml.Yaml;
+
+import org.springframework.cloud.config.environment.Environment;
+import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.cloud.config.server.config.ConfigServerProperties;
+import org.springframework.core.Ordered;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Clay McCoy
+ */
+public class AwsS3EnvironmentRepository implements EnvironmentRepository, Ordered {
+
+	private final AmazonS3 s3Client;
+
+	private final String bucketName;
+
+	private final ConfigServerProperties serverProperties;
+
+	protected int order = Ordered.LOWEST_PRECEDENCE;
+
+	public AwsS3EnvironmentRepository(AmazonS3 s3Client, String bucketName,
+			ConfigServerProperties server) {
+		this.s3Client = s3Client;
+		this.bucketName = bucketName;
+		this.serverProperties = server;
+	}
+
+	@Override
+	public int getOrder() {
+		return this.order;
+	}
+
+	public void setOrder(int order) {
+		this.order = order;
+	}
+
+	@Override
+	public Environment findOne(String specifiedApplication, String specifiedProfile,
+			String specifiedLabel) {
+		final String application = StringUtils.isEmpty(specifiedApplication)
+				? serverProperties.getDefaultApplicationName() : specifiedApplication;
+		final String profile = StringUtils.isEmpty(specifiedProfile)
+				? serverProperties.getDefaultProfile() : specifiedProfile;
+		final String label = StringUtils.isEmpty(specifiedLabel)
+				? serverProperties.getDefaultLabel() : specifiedLabel;
+		StringBuilder objectKeyPrefix = new StringBuilder(application).append("-")
+				.append(profile);
+		if (!StringUtils.isEmpty(label)) {
+			objectKeyPrefix.append("-").append(label);
+		}
+		final Environment environment = new Environment(application, profile);
+		environment.setLabel(label);
+		final S3ObjectIdBuilder s3ObjectIdBuilder = new S3ObjectIdBuilder()
+				.withBucket(bucketName);
+		S3ConfigFile s3ConfigFile = getS3ConfigFile(s3ObjectIdBuilder,
+				objectKeyPrefix.toString());
+		if (s3ConfigFile == null) {
+			throw new NoSuchRepositoryException(
+					"No such repository: (" + s3ObjectIdBuilder
+							.withKey(objectKeyPrefix.toString() + "(.properties | .yml)")
+							.build().toString() + ")");
+		}
+		environment.setVersion(s3ConfigFile.getVersion());
+		final Map config = s3ConfigFile.read();
+		config.putAll(serverProperties.getOverrides());
+		environment.add(new PropertySource(application, config));
+		return environment;
+	}
+
+	private S3ConfigFile getS3ConfigFile(S3ObjectIdBuilder s3ObjectIdBuilder,
+			String keyPrefix) {
+		try {
+			final S3Object properties = s3Client.getObject(new GetObjectRequest(
+					s3ObjectIdBuilder.withKey(keyPrefix + ".properties").build()));
+			return new PropertyS3ConfigFile(properties.getObjectMetadata().getVersionId(),
+					properties.getObjectContent());
+		}
+		catch (Exception eProperties) {
+			try {
+				final S3Object yaml = s3Client.getObject(new GetObjectRequest(
+						s3ObjectIdBuilder.withKey(keyPrefix + ".yml").build()));
+				return new YamlS3ConfigFile(yaml.getObjectMetadata().getVersionId(),
+						yaml.getObjectContent());
+			}
+			catch (Exception eYaml) {
+				return null;
+			}
+		}
+	}
+
+}
+
+abstract class S3ConfigFile {
+
+	private final String version;
+
+	protected S3ConfigFile(String version) {
+		this.version = version;
+	}
+
+	String getVersion() {
+		return version;
+	}
+
+	abstract Map<?, ?> read();
+
+}
+
+class PropertyS3ConfigFile extends S3ConfigFile {
+
+	final InputStream inputStream;
+
+	PropertyS3ConfigFile(String version, InputStream inputStream) {
+		super(version);
+		this.inputStream = inputStream;
+	}
+
+	@Override
+	public Map<?, ?> read() {
+		Properties props = new Properties();
+		try (InputStream in = inputStream) {
+			props.load(in);
+		}
+		catch (IOException e) {
+			throw new IllegalStateException("Cannot load environment", e);
+		}
+		return props;
+	}
+
+}
+
+class YamlS3ConfigFile extends S3ConfigFile {
+
+	final InputStream inputStream;
+
+	YamlS3ConfigFile(String version, InputStream inputStream) {
+		super(version);
+		this.inputStream = inputStream;
+	}
+
+	@Override
+	public Map<?, ?> read() {
+		final Yaml yaml = new Yaml();
+		final Map<String, String> properties = new HashMap<>();
+		try (InputStream in = inputStream) {
+			yaml.loadAll(in).forEach((Object obj) -> this.loadValue(properties, "", obj));
+			return properties;
+		}
+		catch (IOException e) {
+			throw new IllegalStateException("Cannot load environment", e);
+		}
+	}
+
+	private void loadMap(Map<String, String> properties, String baseKey,
+			Map<String, ?> map) {
+		map.keySet().forEach((String configKey) -> {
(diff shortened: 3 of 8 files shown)


pom.xml
```diff
@@ -70,6 +70,11 @@
 				<artifactId>aws-java-sdk-core</artifactId>
 				<version>1.11.52</version>
 			</dependency>
+			<dependency>
+				<groupId>com.amazonaws</groupId>
+				<artifactId>aws-java-sdk-s3</artifactId>
+				<version>1.11.52</version>
```
ryanjbaxter: Probably worth extracting this to a property now that it is used in multiple places

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
```diff
@@ -301,6 +317,19 @@ public NativeEnvironmentRepository nativeEnvironmentRepository(
 
 }
 
+@Configuration
+@Profile("s3")
```
ryanjbaxter: I would be changing the name to just `aws`

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -0,0 +1,217 @@
...
+	YamlS3ConfigFile(String version, InputStream inputStream) {
+		super(version);
+		this.inputStream = inputStream;
+	}
+
+	@Override
+	public Map<?, ?> read() {
+		final Yaml yaml = new Yaml();
+		final Map<String, String> properties = new HashMap<>();
+		try (InputStream in = inputStream) {
```
ryanjbaxter: I think we need to close this `InputStream` at some point

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
```diff
@@ -301,6 +317,19 @@ public NativeEnvironmentRepository nativeEnvironmentRepository(
...
+class AwsS3RepositoryConfiguration {
+
+	@Bean
+	public AwsS3EnvironmentRepository awsS3EnvironmentRepository(
```
ryanjbaxter: Probably should make this conditional to be safe

ryanjbaxter:
* It is probably worth updating the AWS dependency version.  If you do that and all the tests pass it should be fine.

* You say this will work with `.properties` and `.yml` files but we also support serving json and plain text files, is there a limitation on supporting those as well?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
spencergibb: This is [try with resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) and will be closed automatically.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
spencergibb: I don't know. They have a property store that could have an impl in the future.

pom.xml
spencergibb: :+1:

spencergibb:
> This currently works against typical -.properties and -.yml files that are in a single AWS account and in a specified bucket. Label is optionally used as another suffix on the file name (--.yml).

That should come from profile actually

spencergibb:
> I've tried serveral things, but I don't have the dependencies resolving correctly in IntelliJ Idea. I can compile and run tests with mvn on the command line, but it would be great to be able to do those things in IJ.

You need to activate the 'spring' maven profile.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
spencergibb: Maybe 'awss3'

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -0,0 +1,217 @@
...
+		final String application = StringUtils.isEmpty(specifiedApplication)
+				? serverProperties.getDefaultApplicationName() : specifiedApplication;
+		final String profile = StringUtils.isEmpty(specifiedProfile)
+				? serverProperties.getDefaultProfile() : specifiedProfile;
+		final String label = StringUtils.isEmpty(specifiedLabel)
+				? serverProperties.getDefaultLabel() : specifiedLabel;
+		StringBuilder objectKeyPrefix = new StringBuilder(application).append("-")
+				.append(profile);
+		if (!StringUtils.isEmpty(label)) {
+			objectKeyPrefix.append("-").append(label);
```
spencergibb: I see, it does profile, then label. Interesting. In the `NativeEnvironmentRepository` label is used as a directory. Maybe we could do something similar rather than `-{label}`.

spencergibb: There will always be a label, so it should find files without it as well.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -0,0 +1,217 @@
...
+import org.springframework.cloud.config.environment.Environment;
+import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.cloud.config.server.config.ConfigServerProperties;
+import org.springframework.core.Ordered;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Clay McCoy
+ */
+public class AwsS3EnvironmentRepository implements EnvironmentRepository, Ordered {
```
spencergibb: To support the files API, this will need to implement `SearchPathLocator`. This is not required though.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
ryanjbaxter: maybe `awss3`?  `s3` just seems a little vague

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
spencergibb: try w/ resources will close it automatically.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -0,0 +1,217 @@
...
+	final InputStream inputStream;
+
+	YamlS3ConfigFile(String version, InputStream inputStream) {
+		super(version);
+		this.inputStream = inputStream;
+	}
+
+	@Override
+	public Map<?, ?> read() {
+		final Yaml yaml = new Yaml();
```
spencergibb: Lets use `YamlPropertiesFactoryBean` rather than snakeyaml directly. `VaultEnvironmentRepository` users this.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
scottfrederick: While not required, enabling the files/resources API would be desirable for this env repo. Does that require that the files be pre-fetched locally as happens with e.g. git clone, or can files be downloaded on demand?

spencergibb: On demand is fine. `SearchPathLocator` implementation could happen in a follow on PR.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
claymccoy: Yes, I wasn't sure how to best map label to s3 concepts. I saw that in Git it represented a branch. I briefly thought about it representing a version in an s3 bucket or the name of the bucket, but neither seemed right so I just put it at the end.
Label is optional so there won't always be a label right? I could use label as a prefix/directory. Those can be arbitrarily deep, I'm wondering if there is anything limiting label to be only one level deep (without path separators).

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
claymccoy: Okay, not sure what Strings are expected to be put in Locations.locations. Any documentation on SearchPathLocator expectations?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
claymccoy: Thanks, I'll rework that.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
claymccoy: Conditional on classpath, profile something else? Getting the wiring correct on the new S3 config was one of the parts I was least confident that I did correctly.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
spencergibb: I'd prefer the directory style structure since we do that in other places. One level is fine. Though one level in `SearchPathLocator` won't work.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
claymccoy: `awss3` it is then

pom.xml
claymccoy: Okay, I'll probably update it too and run through the unit tests. I do have some (unrelated to this) that fail locally, is that expected or am I missing some set up?

claymccoy:
@ryanjbaxter I can definitely add more documentation if the rest of this looks reasonable. Anything in particular you'd like me to address?

claymccoy:
@ryanjbaxter I can support JSON and plain text as well. Currently for Yaml I flatten the hierarchical structure and put an index on the same paths. Is that correct?
https://github.com/claymccoy/spring-cloud-config/blob/addS3EnvRepo/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryTests.java#L75
I saw this being done to the spinnaker yml in git when running it locally. Should I even be doing that in an EnvironmentRepository? If it is appropriate I guess that I should do the same thing for JSON?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
ryanjbaxter: `@ConditionalOnMissingBean`

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
claymccoy: "Though one level in SearchPathLocator won't work."
Can you expand on that a bit?

ryanjbaxter:
@claymccoy documentation for other environment repositories is under here so you can follow what was done for the others https://github.com/spring-cloud/spring-cloud-config/blob/master/docs/src/main/asciidoc/spring-cloud-config.adoc#environment-repository

> Currently for Yaml I flatten the hierarchical structure and put an index on the same paths. Is that correct?

Does it look like YAML when it is returned?  I believe your `EnvironmentRepository` should not have to do anything special, it should be taken care of upstream.  Although I have not gone through that code myself to confirm that either.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
spencergibb: user can request an arbitrary path. In git any level of sub-directories.

pom.xml
ryanjbaxter: Right now the build is green so I would not expect any failures locally.  Also CircleCI built your branch fine as well

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryFactory.java
```diff
@@ -0,0 +1,44 @@
...
+	final private ConfigServerProperties server;
+
+	public AwsS3EnvironmentRepositoryFactory(ConfigServerProperties server) {
+		this.server = server;
+	}
+
+	@Override
+	public AwsS3EnvironmentRepository build(
+			AwsS3EnvironmentProperties environmentProperties) {
+		final AmazonS3ClientBuilder clientBuilder = AmazonS3ClientBuilder.standard();
```
perek: I believe we should add the ability to override the baseUrl of the s3 client. This will allow support for beta regions of S3, and other s3 compatible storage APIs (GCS, Azure CS)

https://stackoverflow.com/questions/45055921/how-to-set-base-url-in-java-aws-sdk

claymccoy:
@ryanjbaxter Ah, yes you mean real documetation. Not just javadocs. No problem.

Okay, I'll strip out the code I was doing to flatten the Yaml. The less code the better. I have to turn the content into a Map, but it will just be a hierarchical representation. It is hard to intuit the contract when it is just `Map<?, ?> source`.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
claymccoy: Right, so SearchPathLocator be somewhat moot when only one level is allowed? I guess it would return all files in that directory?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
claymccoy: Right, so SearchPathLocator be somewhat moot when only one level is allowed? I guess it would return all files in that directory?

claymccoy:
@ryanjbaxter "You say this will work with .properties and .yml files but we also support serving json and plain text files, is there a limitation on supporting those as well?"

I see in VaultEnvironmentRepository that I can use YamlPropertiesFactoryBean on both JSON and YAML. I'm a little confused by plain text. Do I just look for .txt files and assume they look like a property file?

ryanjbaxter:
@claymccoy you are litterally just serving up the file as text, it could be .txt, .xml, .yml, .conf, etc
https://cloud.spring.io/spring-cloud-static/Greenwich.SR1/single/spring-cloud.html#_serving_plain_text

pom.xml
claymccoy: fixed

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
claymccoy: fixed

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
claymccoy: fixed, except I used @ConditionalOnClass(AmazonS3Client.class) and @ConditionalOnBean(AmazonS3Client.class) like the other RepositoryConfigurations.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
claymccoy: fixed, label now denotes a directory path inside a bucket.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
claymccoy: fixed

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryFactory.java
claymccoy: fixed

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java
ryanjbaxter: I dont see why it is conditional on `AmazonS3Client` it doesn't use it
Also the reason I was suggesting adding `@ConditionalOnMissingBean(AwsS3EnvironmentRepository)` is so someone could optionally provide their own if they needed to tweak anything

claymccoy:
I'm trying to run this locally after modifying `spring-cloud-config-server/src/main/resources/configserver.yml` to have this content:
```
info:
  component: Config Server
spring:
  profiles:
    active: awss3
  cloud:
    config:
      server:
        awss3:
          region: us-east-2
          bucket: cmccoy-config

server:
  port: 8888
management:
  context_path: /admin
```

It looks good at first:
`[           main] o.s.c.c.server.ConfigServerApplication   : The following profiles are active: awss3`

But ultimately fails to start. It doesn't seem to really have the awss3 profile. Either I am running this incorrectly or I don't have something wired up correctly in my new beans.
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Failed to configure a DataSource: 'url' attribute is not specified and no embedded datasource could be configured.

Reason: Failed to determine a suitable driver class

Action:

Consider the following:
	If you want an embedded database (H2, HSQL or Derby), please put it on the classpath.
	If you have database settings to be loaded from a particular profile you may need to activate it (no profiles are currently active).
```

spencergibb:
There are many optional dependencies in `spring-cloud-config-server` where `ConfigServerApplication` lives if you're running in the IDE. Would probably make sense to create a stand-alone version from https://start.spring.io to eliminate those.

claymccoy:
@ryanjbaxter I have been out on vacation and just getting back to this. I think I have addressed all the feedback. Have successfully tested this in a vanilla boot app that is configured to use config server, and made some minor additions to get all the wiring right. Let me know if I need to do anything else for this to be merged. Thanks!

Issue #1295, opened 2019-03-18
spring cloud config server overrides not working when bootsrap enabled


mehdikhelif:
Hello,

We are trying to update our cloud config server to Greenwhich.SR1 and boot 2.1.3.RELEASE but the spring.cloud.config.server.overrides is broken. It is quite easy to reproduce. This happens when spring.cloud.config.server.bootstrap is set to true and overrides are put on git. 
After some investigations, it seems that the problem is in `org.springframework.cloud.config.server.config.ConfigServerMvcConfiguration` 

```java
private EnvironmentRepository encrypted(EnvironmentRepository envRepository,
			ConfigServerProperties server) {
   EnvironmentEncryptorEnvironmentRepository encrypted = new 
   EnvironmentEncryptorEnvironmentRepository(
				envRepository, this.environmentEncryptor);
   encrypted.setOverrides(server.getOverrides());
   return encrypted;
}
```
That `EnvironmentEncryptorEnvironmentRepository` is initialized with the content of ConfigurationProperties `ConfigServerProperties` before it is rebound when the context is refreshed so it does not contain the values coming from Git. I think the solution would be to inject the `ConfigServerProperties` inside `EnvironmentEncryptorEnvironmentRepository` instead of passing the values.

This was tested with Finchley.SR3 and boot 2.0.8.RELEASE with same results.

ryanjbaxter:
What are you upgrading from?

mehdikhelif:
Hello @ryanjbaxter, thx for your quick reply. I am upgrading from Edgware.SR4/boot 1.5.14.RELEASE. But it does not really matter. I get exactly the same when creating a new config-server from scratch.

ryanjbaxter:
Are you using a composite environment as well?
http://cloud.spring.io/spring-cloud-static/Greenwich.RELEASE/single/spring-cloud.html#_embedding_the_config_server

mehdikhelif:
By default we dont but I already tried without success. However, I do not think it is related. The config is loaded correctly, just too late or the bean is created too early, depends how you see it

ryanjbaxter:
Can you provide a [complete, minimal, verifiable sample](https://stackoverflow.com/help/mcve) that reproduces the problem? It should be available as a GitHub (or similar) project or attached to this issue as a zip file.

mehdikhelif:
Hello, you can find example here : https://github.com/mehdikhelif/config-server . I have created a small test verifying what I explained

[...]

PR #1433, opened 2019-07-30
Support context refresh of config server.

TYsewyn:
Whenever the `spring.cloud.config.server.overrides` property was changed
(either through the env actuator endpoint or remote bootstrap
configuration) this was not reflected when retrieving the application's
configuration.
Also fixes gh-1295.


spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/RefreshableConfigServerIntegrationTests.java
```diff
@@ -0,0 +1,133 @@
...
+
+	private static String localRepo = null;
+
+	@LocalServerPort
+	private int port;
+
+	@BeforeClass
+	public static void init() throws IOException {
+		// mock Git configuration to make tests independent of local Git configuration
+		SystemReader.setInstance(new MockSystemReader());
```
spencergibb: Does this need to be undone in `after()`?

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/RefreshableConfigServerIntegrationTests.java
```diff
@@ -0,0 +1,133 @@
...
+		assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
+
+		response = new TestRestTemplate().postForEntity(actuatorEndpoint + "/refresh",
+				null, Void.class);
+		assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
+
+		environment = new TestRestTemplate().getForObject(
+				"http://localhost:" + this.port + "/foo/development/", Environment.class);
+		assertThat(environment.getPropertySources()).isNotEmpty();
+		assertThat(environment.getPropertySources().get(0).getSource().get("foo"))
```
spencergibb: What is the value of foo prior to the refresh?

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/RefreshableConfigServerIntegrationTests.java
TYsewyn: I honestly can't answer this. This test setup was based on `ConfigClientOnIntegrationTests`.

spencergibb: lol

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/RefreshableConfigServerIntegrationTests.java
TYsewyn: There weren't any properties at first. See [L85](https://github.com/spring-cloud/spring-cloud-config/pull/1433/files/cdb127e941056fae62b0577a32b4bea188621fb6#diff-b144d4082f116fc22971058c17bd001cR85)
It's only after setting the `overrides` and refreshing the context that the foo application will have this `foo: bar` property.

spencergibb: And because it's using `spring.cloud.config.server.overrides.foo` and not just `foo` we know this works then?

TYsewyn: Yes. We're emulating an application `foo` which is running with the `development` profile and is asking for its properties using the REST endpoint. We're also calling the `/env` & `/refresh` actuator endpoints to change the `spring.cloud.config.server.overrides.foo` property. 
Since we see that we only get the "overridden" `foo` property after the context refresh we are sure that the properties have been set and the `EnvironmentController` bean has successfully been recreated with the new overrides.

spencergibb: Awesome. Could you just copy that explanation into the test?

TYsewyn: Resolved with [cb85fc3](https://github.com/spring-cloud/spring-cloud-config/pull/1433/commits/cb85fc3154a80312e65d3709c9c41e94c12833dd).

PR #1458, opened 2019-08-31
Feature/gh 1283 exception

nhomble:
Should help resolve GH-1283
When the configurable application context throws an error when constructing the properties (aka snake yml throws a syntax error underneath), then we can wrap the exception in a custom exception to catch upstream.


nhomble:
@spencergibb @ryanjbaxter sorry for the delay since my first pr - let me know if this needs more work. Thanks!

PR #1466, opened 2019-09-15
Update documentation: change variable names in resource paths.

jeremydr2:
Change `{name}` to `{application}` to match other uses of
`spring.config.name` being used as such.
Explicitly call out that files in the `Serving Plain Text` section can
be paths and not just a filename.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -1143,7 +1143,7 @@ TIP: If you testing with curl, then use `--data-urlencode` (instead of `-d`) or 
 
 Take the encrypted value and add the `{cipher}` prefix before you put it in the YAML or properties file and before you commit and push it to a remote (potentially insecure) store.
 
-The `/encrypt` and `/decrypt` endpoints also both accept paths in the form of `/*/{name}/{profiles}`, which can be used to control cryptography on a per-application (name) and per-profile basis when clients call into the main environment resource.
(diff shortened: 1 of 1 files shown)


jeremydr2:
I believe that using `{application}` consistently throughout the document (it is already the preferred term over `{name}`) and removing `{name}` entirely will enable newcomers to more quickly understand that there are not two separate terms.

I also think that explicitly calling out that `{path}` can be a full filepath with slashes is necessary, as otherwise some (like me) might initially assume that `{path}` has similar limitations to `{application}` where it cannot contain slashes, or that you are not allowed to specify a path at all.

I realize these things are obvious to more experienced users of spring cloud config server, but they were not initially obvious to me, and I believe this change could prevent more confusion by future newcomers.

Haybu:
hi @jeremydr2, not realizing yours but another PR is also merged for the same, would close this issue if you're good.

jeremydr2:
@Haybu , Why did you make a separate commit that was almost identical to mine and then merge it? This bothers me from a contribution perspective. I understand my change was miniscule but I felt happy about being able to contribute to this project, and it feels like you're taking that effort away from me.

Additionally, I do not think this is sufficient. It is unclear from the docs as written and as you merged that `path` as mentioned in "Serving Plain Text" can be a full filepath with `/`, such as `resources/log.xml`, since in other places you cannot use `/` in the html path variables used by config server.

spencergibb:
He clearly states he didn't realize this PR was here.

Haybu:
@jeremydr2 please submit a PR, your contribution is always welcome. Apology for any inconvenience caused by not realizing your PR.

jeremydr2:
@spencergibb , he references this PR (#1466) in his commit message.

PR #1455, opened 2019-08-28
Sharing default application config in CredHub

Albertoimpl:
Always including application as default so that it can be shared across all applications to be consistent with other repository implementations.

Sharing secrets to the application from the default profile.

Updated the docs to reflect the changes

PR #1477, opened 2019-09-24
Sharing default application config in CredHub

Albertoimpl:
Always including application as default so that it can be shared across all applications to be consistent with other repository implementations.

Sharing secrets to the application from default profile.

Updated the docs to reflect the changes.
So that it can get into Greenwich creating a PR against 2.1.x.


Albertoimpl:
@spencergibb feel free to close this other one: https://github.com/spring-cloud/spring-cloud-config/pull/1455

PR #1497, opened 2019-10-27
Support proxy configuration(not global) to connect Config Server

jeanpaulct:
The scenario is the config client need to connect to Config Server through a proxy, but the others components(service providers) not necessarily use that same proxy.

Added support to use an exclusive configuration proxy (http/socks) to connect config client to config server.


spencergibb:
Given that no one else has asked for this, I'd like to see if anyone else is interested before merging

spencergibb:
Given that no one else has asked for this, I'd like to see if anyone else is interested before merging

iskradelta:
You can always use the jdks proxy instead of letting the application know or care about it, just pass -DsocksProxy=etc to your java command. Also this would be moving a configuration into a configuration-client, awkard. So no vote from me.

jeanpaulct:
> You can always use the jdks proxy instead of letting the application know or care about it, just pass -DsocksProxy=etc to your java command. Also this would be moving a configuration into a configuration-client, awkard. So no vote from me.

I have an application that has three components: spring-config-client(1), authentication-provider(2) and business-provider(3), all of them use a http connection to perform operations, the second and third must connect to internal(company's policy) network but spring-config-client must to connect internet(github) to fetch configuration application, so the networking team has created a proxy to allow the connection on internet.
If I use jdk proxy all of the components will use the proxy to connect through http, spring-config-client will work but the second and third component will break.

ryanjbaxter:
Closing due to lack of votes

PR #1507, opened 2019-11-12
Return HTTP Client of configuration if there is an exact match on the git url and not a placeholder

### LINKED ISSUE #1480: Error: "More than one git repo URL template matched URL"
We recently tried to upgrade from Edgware.SR5 to Greenwich SR3. ConfigServer reports the following error for certain requests:

> More than one git repo URL template matched URL

The error is coming from the `HttpClientConfigurableHttpConnectionFactory.lookupHttpClientBuilder` method. With Greenwich SR3, a [bugfix](https://github.com/spring-cloud/spring-cloud-config/pull/1429) was implemented in that class based on a support ticket my team opened. However, it appears it was only a partial fix.

We have a bootstrap similar to this:

```
spring:
  cloud:
    config:
      server:
        git:
          uri: https://blahblah/v1/repos/vvv_configs-{application}
          proxy:
            https:
              host: ${proxyuri}
              port: '8080'

          repos:

            WebApp:
              pattern: webapp
              searchPaths: SharedDomains/*, {application}
              basedir: ${basedirParent}\\baseWebApp
              uri: https://blahblah/v1/repos/vvv_configs-webapp
              proxy:
                https:
                  host: ${proxyuri}
                  port: '8080'
```

The problem is that if you call configserver for application "webapp", the `lookupHttpClientBuilder` method finds **both** uris shown in the sample bootstrap above. The `lookupHttpClientBuilder` expected to find one and only one. The service will continue on to process the request, but logs error:

> More than one git repo URL template matched URL: https://blahblah/v1/repos/vvv_configs-webapp/info/refs?service=git-upload-pack, proxy and skipSslValidation config won't be applied. Matched templates: https://blahblah/v1/repos/vvv_configs-{application}, https://blahblah/v1/repos/vvv_configs-webapp

We need to use the proxy, so this won't work for us.

This did not happen with Finchley. (The `HttpClientConfigurableHttpConnectionFactory` class did not exist in that release.)

1) I wanted to first relay the issue and request that it be fixed.
2) Short-term, I am trying to find a workaround. I am hoping you can help. We have a strong need to upgrade to the latest release very soon.

**Workaround:**
For a workaround, I tried adding a default repo entry to the end of the bootstrap and removing the default git entry, as shown below. Unsurprisingly, the service won't even start if you don't have `spring.cloud.config.server.git.uri` defined. So this did not work:

```
spring:
  cloud:
    config:
      server:
        git:
          repos:
            WebApp:
              pattern: webapp
              searchPaths: SharedDomains/*, {application}
              uri: https://blahblah/v1/repos/vvv_configs-webapp
              proxy:
                https:
                  host: ${proxyuri}
                  port: '8080'
                  
            Default:
              pattern: '*'
              uri: https://blahblah/v1/repos/vvv_configs-{application}
              proxy:
                https:
                  host: ${proxyuri}
                  port: '8080'
```

We could add the `spring.cloud.config.server.git.uri` back in to point to just one repo instead of using {application} placeholder and have the `Default` for everything else. But for that one repo specified on `spring.cloud.config.server.git.uri`, we'll get the error, so that won't work. We could put a dummy repo in for `spring.cloud.config.server.git.uri`, but that is too much of a hack for a long-term solution.

Can you think of a way for us to configure repos so that only one repo url is ever found? Is there a way to specify the `Default` repo as in my example, but in addition to `pattern`, add an exclusion to exclude the repo present as the default? Something like the example below. In that example, note:

- The `spring.cloud.config.server.git.uri` is set to vvv_configs-**authenticationAccounts** instead of vvv_configs-**{application}**

- The `Default` repo has a `patternExclude` (which isn't a real parameter).

```
spring:
  cloud:
    config:
      server:
        git:
          uri: https://blahblah/v1/repos/vvv_configs-authenticationAccounts
          proxy:
            https:
              host: ${proxyuri}
              port: '8080'

          repos:

            WebApp:
              pattern: webapp
              searchPaths: SharedDomains/*, {application}
              uri: https://blahblah/v1/repos/vvv_configs-webapp
              proxy:
                https:
                  host: ${proxyuri}
                  port: '8080'
                  
            Default:
              pattern: '*'
              patternExclude: authenticationAccounts
              uri: https://blahblah/v1/repos/vvv_configs-{application}
              proxy:
                https:
                  host: ${proxyuri}
                  port: '8080'
```

Can `pattern` be regex, or is asterisk for wildcard the only matching that is supported?

Note that eliminating having a token in the uri (vvv_configs-{application}) and explicitly defining a repo entry for every service and app is not a feasible workaround for us. We have many dozens (100s?) of apps and services. It's not feasible to add a repo entry into the bootstrap for each of them and also have to continue adding for every new service we create.

### LINKED ISSUE #1482: The wrong http settings may be used if there are two git repo configs for the same url
While investigating issue [Error: "More than one git repo URL template matched URL"](https://github.com/spring-cloud/spring-cloud-config/issues/1480), I spotted another possible defect in the HttpClientConfigurableHttpConnectionFactory implementation.

If someone has multiple git repo configs for the same url, but with different http settings, only one of the http settings will be used, regardless of how the request coming in. 

Note: This does not impact my team, but since I spotted it, I thought I would relay this.

If someone has a configuration such as:

```
spring:
  cloud:
    config:
      server:
        git:
          uri: https://blahblah/v1/repos/vvv_configs-{application}
          proxy:
            https:
              host: ${proxyuri}
              port: '8080'

          repos:

           # No proxy in dev environment
            WebApp-Dev:
              pattern: webapp/dev
              searchPaths: SharedDomains/*, {application}
              basedir: ${basedirParent}\\baseWebApp
              uri: https://blahblah/v1/repos/vvv_configs-webapp
				  
           # Proxy needed for all environments except dev
            WebApp:
              pattern: webapp
              searchPaths: SharedDomains/*, {application}
              basedir: ${basedirParent}\\baseWebApp
              uri: https://blahblah/v1/repos/vvv_configs-webapp
              proxy:
                https:
                  host: ${proxyuri}
                  port: '8080'
```

The `HttpClientConfigurableHttpConnectionFactory.addConfiguration` method puts http clients into its internal map with the url as key. Since url is the same, that means that there will only be one entry for the vvv_configs-webapp URL in the map, even though there are two different configs for that URL. Whether the http client will be the one using the proxy or the one without is indeterminate (second one added to the map wins, obviously).

I did a quick test, and confirmed this behavior. Whether I call with "dev" profile or not, it will always use the same http client object. I am using Greenwich SR3.

ryanjbaxter:
Fixes #1480
Fixes #1482

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
@@ -105,6 +105,14 @@ public class HttpClientConfigurableHttpConnectionFactory
 			return HttpClients.custom();
 		}
 		if (builderMap.size() > 1) {
+			List<String> keys = builderMap.keySet().stream().filter(key -> {
+				String[] tokens = key.split(PLACEHOLDER_PATTERN);
+				return tokens.length == 1;
+			}).collect(Collectors.toList());
+
+			if (keys.size() == 1) {
+				return builderMap.get(keys.get(0));
+			}
 			this.log.error(String.format(
 					"More than one git repo URL template matched URL:"
 							+ " %s, proxy and skipSslValidation config won't be applied. Matched templates: %s",
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
@@ -105,6 +105,11 @@ public class HttpClientConfigurableHttpConnectionFactory
 			return HttpClients.custom();
 		}
 		if (builderMap.size() > 1) {
+			/*
+			 * Try to determine if there is an exact match URL or not. So if there is a placeholder in the URL, filter
+			 * it out. We should be left with only URLs which have no placeholders.
+			 * That is the one we want to use in the case there are multiple matches.
+			 */
 			List<String> keys = builderMap.keySet().stream().filter(key -> {
 				String[] tokens = key.split(PLACEHOLDER_PATTERN);
 				return tokens.length == 1;
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,6 @@ jobs:
     branches:
       ignore:
         - gh-pages # list of branches to ignore
-    resource_class: large
     steps:
       - checkout
       - restore_cache:
--- a/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/RefreshableConfigServerIntegrationTests.java
+++ b/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/RefreshableConfigServerIntegrationTests.java
@@ -79,9 +79,9 @@ public class RefreshableConfigServerIntegrationTests {
 	}
 
 	/*
-	 * We're emulating an application "foo" which is running with the "development" profile
-	 * and is asking for its properties using the REST endpoint. We're also calling the
-	 * /env & /refresh actuator endpoints to change the
+	 * We're emulating an application "foo" which is running with the "development"
+	 * profile and is asking for its properties using the REST endpoint. We're also
+	 * calling the /env & /refresh actuator endpoints to change the
 	 * `spring.cloud.config.server.overrides.foo` property. Since we see that we only get
 	 * the overridden "foo" property after the context refresh we are sure that the
 	 * properties have been set and the EnvironmentController bean has successfully been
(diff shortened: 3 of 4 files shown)


ryanjbaxter:
@marnee01 can you test out this fix?  I tested it with your sample app and it works.  I think this implementation is simpler as we dont have to iterate through all the http clients

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
```diff
@@ -105,6 +105,14 @@ private HttpClientBuilder lookupHttpClientBuilder(final URL url) {
 			return HttpClients.custom();
 		}
 		if (builderMap.size() > 1) {
+			List<String> keys = builderMap.keySet().stream().filter(key -> {
```
spencergibb: Can you comment here what the goal is?

ryanjbaxter: @spencergibb Basically I am determining if there is an exact match URL or not.  So if there is a placeholder in the URL I filter it out.  We should be left with only URLs which have no placeholders.  That is the one we want to use in the case there are multiple matches.

Issue #1427, opened 2019-07-15
RSA keystore requires both public/private key to encrypt


minhster99:
Bug report

This bug is related to https://github.com/spring-cloud/spring-cloud-config/issues/539

According to the docs
"The encryption is done with the public key, and a private key is needed for decryption. Thus in principle you can configure only the public key in the server if you only want to do encryption (and are prepared to decrypt the values yourself locally with the private key)."

This implies the private key isn't required in the server keystore when used purely for encryption. I have created a keystore with only the public key in it and the server has picked it up. However when attempting to encrypt using the endpoint, I get the following exception

```
Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is java.lang.IllegalStateException: Cannot load keys from store: class path resource [keystore-public.jks]] with root cause
java.lang.NullPointerException: null
	at org.springframework.security.rsa.crypto.KeyStoreKeyFactory.getKeyPair(KeyStoreKeyFactory.java:70)
	at org.springframework.cloud.config.server.encryption.KeyStoreTextEncryptorLocator.locate(KeyStoreTextEncryptorLocator.java:87)
	at org.springframework.cloud.config.server.encryption.EncryptionController.checkEncryptorInstalled(EncryptionController.java:166)
	at org.springframework.cloud.config.server.encryption.EncryptionController.encrypt(EncryptionController.java:126)
```

If my server keystore has both the private and public key in it, the problem goes away. We prefer not to do this from a security point of view as we want our config clients to perform the decryption and look after their own private keys.

TYsewyn:
Hi @minhster99! Did you disable decryption in your `bootstrap.properties|yml` file of your config server by setting `spring.cloud.config.server.encrypt.enabled=false`?

minhster99:
Hi @TYsewyn , yes I have that disabled but I wouldn't expect that to matter since I'm hitting the encrypt URL. I believe that only applies when you retrieve properties?

TYsewyn:
> We prefer not to do this from a security point of view as we want our config clients to perform the decryption and look after their own private keys.

As per the docs:
[...]

PR #1518, opened 2019-12-09
Added support for trusted certificates

TYsewyn:
Closes gh-1427


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
```diff
@@ -169,6 +170,22 @@ private void checkEncryptorInstalled(String name, String profiles) {
 		}
 	}
 
+	private void checkDecryptionPossible(String name, String profiles) {
+		if (this.encryptor == null) {
```
ryanjbaxter: I would prefer we avoid duplicating the code from `chckEncryptorInstalled`.  Does the order of these checks matter?  Could we call `checkEncryptorInstalled` then check if we can actually decrypt it?

TYsewyn: Nope, the order doesn't matter. That's a remainder of trying to implement this functionality without touching `spring-security-rsa`. Will clean this up.

Issue #1474, opened 2019-09-20
Refactor: use Spring Vault for accessing Vault with Config Server


scottfrederick:
There have been several request to support additional Vault authentication options with Config Server (e.g. #1240, #1464, https://github.com/spinnaker/spinnaker/issues/4879). This would best be accomplished by taking advantage of the [authentication support already provided by Spring Vault](https://docs.spring.io/spring-vault/docs/current/reference/html/#vault.core.authentication). 

A good first step toward this goal would be to refactor the existing Vault support in Config Server to use Spring Vault instead of a `RestTemplate` without adding new features.

PR #1522, opened 2019-12-16
Migrate VaultEnvironmentRepository from RestTemplate to Spring Vault.

scottfrederick:
This PR aims to migrate the Spring Cloud Config Server support for Vault from using a `RestTemplate` directly to using Spring Vault and `VaultTemplate`. These changes should provide exactly the same features with backward-compatible properties-based auto-configuration and Java configuration, while allowing for more features in the future via Spring Vault's extensive support for Vault integration. 

Fixes https://github.com/spring-cloud/spring-cloud-config/issues/1474

### DIFF
--- a/spring-cloud-config-dependencies/pom.xml
+++ b/spring-cloud-config-dependencies/pom.xml
@@ -16,6 +16,7 @@
 	<description>Spring Cloud Config Dependencies</description>
 	<properties>
 		<jgit.version>5.1.3.201810200350-r</jgit.version>
+		<spring-vault.version>2.2.0.RELEASE</spring-vault.version>
 		<spring-credhub.version>2.0.0.RELEASE</spring-credhub.version>
 	</properties>
 	<dependencyManagement>
@@ -39,6 +40,11 @@
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-config-monitor</artifactId>
 				<version>${project.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.vault</groupId>
+				<artifactId>spring-vault-core</artifactId>
+				<version>${spring-vault.version}</version>
 			</dependency>
 			<dependency>
 				<groupId>org.springframework.credhub</groupId>
--- a/spring-cloud-config-server/.flattened-pom.xml
+++ b/spring-cloud-config-server/.flattened-pom.xml
@@ -159,6 +159,13 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>org.springframework.vault</groupId>
+      <artifactId>spring-vault-core</artifactId>
+      <version>2.2.0.RELEASE</version>
+      <scope>compile</scope>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>org.springframework.credhub</groupId>
       <artifactId>spring-credhub-core</artifactId>
       <version>2.0.0.RELEASE</version>
--- a/spring-cloud-config-server/pom.xml
+++ b/spring-cloud-config-server/pom.xml
@@ -54,6 +54,11 @@
 			<artifactId>spring-security-rsa</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.vault</groupId>
+			<artifactId>spring-vault-core</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.credhub</groupId>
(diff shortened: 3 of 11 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultKvAccessStrategy.java
```diff
@@ -36,77 +32,11 @@
...
 	 * Return secrets from Vault. The response is represented as JSON object marshaled to
 	 * {@link String}.
-	 * @param headers must not be {@literal null}.
-	 * @param backend secret backend mount path, must not be {@literal null}.
 	 * @param key key within the key-value secret backend, must not be {@literal null}.
 	 * @return the marshaled JSON object or {@literal null} if the key was not found.
 	 * @throws RestClientException in case of a transport/access failure.
 	 * @see com.fasterxml.jackson.annotation.JsonRawValue
 	 */
-	String getData(HttpHeaders headers, String backend, String key)
```
spencergibb: Maybe default method that calls `getData(key)`?

scottfrederick:
>

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultKvAccessStrategy.java
scottfrederick: Resolved in 9c2993cfb16697b7702d374ebe14d6a435cc9d5c

scottfrederick:
One thing that we probably need to change here: I made the dependency on Spring Vault `<optional>true</optional>` to match how we handle Spring CredHub and some other libs that users need to manually add to the Config Server app classpath only if they want to use the corresponding `EnvironmentRepository`. That means user apps that currently use the Vault environment repo will break unless they add the Spring Vault dependency. That's probably not the best thing for backward compatibility, so it's probably better to make the Spring Vault dependency non-optional.

spencergibb:
Not sure about making it non-optional. That would mean everyone gets it even if they don't require it. This already has some breaking API changes, so requiring them to add a new dependency doesn't seem that unrealistic.

scottfrederick:
Sounds good. If you think it's acceptable to include the breaking API change to `VaultKvAccessStrategyFactory`, what do you think about making it `VaultKvAccessStrategy` so there is less surface area to the public API in the future?

spencergibb:
I'm still debating the breaking API changes actually. I'd like to hear from other team members about that @ryanjbaxter @marcingrzejszczak @OlgaMaciaszek @Haybu @TYsewyn.

Either we allow breaking API changes or we create a whole new package with a new object graph and deprecate the old and activate one or the other based on the presence of spring-vault on the classpath.

scottfrederick:
> or we create a whole new package with a new object graph ... and activate one or the other based on the presence of spring-vault on the classpath.

That probably wouldn't be very hard to do, and would give a path forward for the Vault enhancements that is driving this migration. I can start taking a look at that.

spencergibb:
That would alleviate any concerns I have. It should just be a copy of all vault classes to a new package and restore the old ones and add some conditionals in autoconfiguration. Then the old stuff can be removed in ilford. Thanks for taking a look.

TYsewyn:
I agree with @spencergibb since this would be the most easy and straightforward way to migrate from Hoxton to Ilford.
It would also make our lives easier in terms of support and maintenance since fixes in Ilford can easily be backported to Hoxton.

scottfrederick:
@spencergibb I re-worked this to make a `SpringVaultEnvironmentRepository` completely separate from the existing `VaultEnvironmentRepository`. It's actually cleaner this way, without imposing backward-compatibility restrictions on the Spring Vault setup code. The changes now are purely additive except for the conditional auto-config in `EnvironmentRepositoryConfiguration` and extracting much of `VaultEnvironmentRepository` to a base class to eliminate duplication with the new implementation. 

`SpringVaultEnvironmentRepository` will not support `skipSslValidation` with this separation. When we add support for the additional Spring Vault authentication options we can add [SSL configuration](https://docs.spring.io/spring-vault/docs/current/reference/html/#vault.client-ssl) also.

snowe2010:
This pr seems to have removed the token check mentioned [here](https://cloud.spring.io/spring-cloud-config/multi/multi__spring_cloud_config_client.html#_vault) for clients. All api requests to config server return vault secrets in the open now, without passing a `token`. How can we get the old functionality back?

PR #1526, opened 2019-12-18
Add support for additional Vault authentication methods.

### LINKED ISSUE #1475: Support additional Vault authentication methods using declarative Config Server properties
Config Server currently supports authentication to Vault using token auth, where the token can be provided in an `X-Config-Token` header in the HTTP request from a Config Client, or provided in a `spring.cloud.config.server.vault.token` property (which is useful when embedding Config Server). 

Config Server support for Vault should be extended to include all supported Vault auth methods with declarative `spring.cloud.config.server.vault` properties. These properties could look very similar to the [properties provided by Spring Cloud Vault](https://cloud.spring.io/spring-cloud-vault/reference/html/#vault.config.authentication). 

Support for additional auth methods via HTTP request headers from a Config Client should be evaluated also, with separate issues as appropriate. 

Depends on #1474

scottfrederick:
This PR adds support for all the Vault authentication methods supported by Spring Vault, using the exact property model implemented by [Spring Cloud Vault](https://cloud.spring.io/spring-cloud-vault/reference/html/#vault.config.authentication).

Fixes #1475

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultClientConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultClientConfiguration.java
@@ -1,0 +1,582 @@
+/*
+ * Copyright 2018-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.config.server.environment.vault;
+
+import java.io.ByteArrayInputStream;
+import java.net.URI;
+import java.util.Base64;
+import java.util.concurrent.atomic.AtomicReference;
+
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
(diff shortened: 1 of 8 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultEnvironmentRepositoryFactory.java
```diff
@@ -49,23 +37,20 @@
 
 	private final EnvironmentWatch watch;
 
-	private final ConfigTokenProvider tokenProvider;
+	private final SpringVaultClientConfiguration clientConfiguration;
 
 	public SpringVaultEnvironmentRepositoryFactory(
```
ryanjbaxter: I think we should add a new constructor rather than change the existing one so we avoid breaking anyone

scottfrederick: This class and its constructor are brand new, didn't exist before yesterday: https://github.com/spring-cloud/spring-cloud-config/commit/5e33794332815290b75af47874487db69f47ccc0.

ryanjbaxter: Ah ok :) . I didnt get a chance to look at it

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultClientConfiguration.java
```diff
@@ -0,0 +1,582 @@
...
+	 */
+	public ClientAuthentication clientAuthentication() {
+
+		AuthenticationMethod authentication = this.vaultProperties.getAuthentication();
+
+		if (authentication == null) {
+			return new ConfigTokenProviderAuthentication(configTokenProvider);
+		}
+
+		switch (authentication) {
```
spencergibb: I'm a little concerned about the size of this class. I'm ok refactoring it later since it's configuration and not a public api. I'd wonder if something like a `VaultAuthenticationFactory` interface with a `supports(AuthenticationMethod)` and `getClientAuthentication(VaultProperties, RestOperations)` would be helpful. That way each method could be a separate small class, you could also do things like `@ConditionalOnClass`

scottfrederick: +1 on that. I was trying to stick pretty close to the two classes from Spring Vault and Spring Cloud Vault that do similar configuration to have the best chance of getting something that works in short order, but decomposing this into a set of smaller classes would be better.

PR #1534, opened 2020-01-06
Use a more robust filter when dealing with multiple keys

### LINKED ISSUE #1533: Error: "More than one git repo URL template matched URL"
The issue reported in https://github.com/spring-cloud/spring-cloud-config/issues/1480 is not completely fixed. If the git URL contains placeholder at the end, then the issue still occurs.

1480 had a sample project to demonstrate the issue, but was using a Bitbucket-style URI that had placeholder in the middle of the URI. We are actually using CodeCommit, though I don't have the ability to create repos there directly, so will not be providing a sample project. (Given the sample project I already sent, you can use your own CodeCommit repos that you have set up for testing and plug those in and should therefore be able to easily reproduce the issue.)

**Additional information**
In sample project provided for 1480, the URI with placeholder is as follows, with the placeholder in the middle of the URI.

https://github.com/marnee01/mderider-{application}.git

This new code in the fixed version, as shown below, will find that splitting on the above style of URI produces a token length of 2, so it does not meet the filter. The correct one (without placeholder) does meet the filter criteria and **is** returned, so the result is one and only one matched URI, as desired. So, it is fixed for that style of URI.

```
List<String> keys = builderMap.keySet().stream().filter(key -> {
    String[] tokens = key.split(PLACEHOLDER_PATTERN);
    return tokens.length == 1;
    }).collect(Collectors.toList());
```
However, in CodeCommit the URI looks like this, with {application} at the end:

https://git-codecommit.us-east-1.amazonaws.com/v1/repos/blah_configs-{application}

In this case, the above code snippet, when the URI is split, it produces only one token, so the one with placeholder **does** meet the criteria. Thus, the code snippet collects two items, including the one with placeholder. Thus the error subsequently occurs.

Recommend that you review your various testing scenarios and test all of them against the final fix, not just the one in my sample project. In case this helps the developers to understand and or test the difference in splitting on these two URI patterns:
```
public static void main(String[] args) {
        final String key = "https://git-codecommit.us-east-1.amazonaws.com/v1/repos/blahs_configs-{application}";
        final String[] tokens = key.split(PLACEHOLDER_PATTERN);
        System.out.println("tokens length when placholder at end: " + tokens.length);
        
        final String key2 = "https://github.com/marnee01/mderider-{application}.git";
        final String[] tokens2 = key2.split(PLACEHOLDER_PATTERN);
        System.out.println("tokens length when placholder in the middle: " + tokens2.length);
}
```

ryanjbaxter:
Fixes #1533

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
@@ -111,10 +111,8 @@ public class HttpClientConfigurableHttpConnectionFactory
 			 * which have no placeholders. That is the one we want to use in the case
 			 * there are multiple matches.
 			 */
-			List<String> keys = builderMap.keySet().stream().filter(key -> {
-				String[] tokens = key.split(PLACEHOLDER_PATTERN);
-				return tokens.length == 1;
-			}).collect(Collectors.toList());
+			List<String> keys = builderMap.keySet().stream().filter(key ->
+				!Pattern.compile(PLACEHOLDER_PATTERN).matcher(key).find()).collect(Collectors.toList());
 
 			if (keys.size() == 1) {
 				return builderMap.get(keys.get(0));
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
@@ -111,8 +111,9 @@ public class HttpClientConfigurableHttpConnectionFactory
 			 * which have no placeholders. That is the one we want to use in the case
 			 * there are multiple matches.
 			 */
-			List<String> keys = builderMap.keySet().stream().filter(key ->
-				!Pattern.compile(PLACEHOLDER_PATTERN).matcher(key).find()).collect(Collectors.toList());
+			List<String> keys = builderMap.keySet().stream().filter(
+					key -> !Pattern.compile(PLACEHOLDER_PATTERN).matcher(key).find())
+					.collect(Collectors.toList());
 
 			if (keys.size() == 1) {
 				return builderMap.get(keys.get(0));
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
@marnee01 feel free to give this a test

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactoryTest.java
```diff
@@ -177,6 +177,24 @@ public void multipleMatchesWithPlaceholder() throws Exception {
 		assertThat(actualHttpClientBuilder).isSameAs(expectedHttpClientBuilder);
 	}
 
+	@Test
+	public void multipleMatchesWithPlaceholderAtEnd() throws Exception {
+		MultipleJGitEnvironmentProperties properties1 = new MultipleJGitEnvironmentProperties();
+		properties1.setUri("https://github.com/marnee01/mderider-{application}");
```
spencergibb: I'd remove the username here

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
```diff
@@ -111,10 +111,9 @@ private HttpClientBuilder lookupHttpClientBuilder(final URL url) {
 			 * which have no placeholders. That is the one we want to use in the case
 			 * there are multiple matches.
 			 */
-			List<String> keys = builderMap.keySet().stream().filter(key -> {
-				String[] tokens = key.split(PLACEHOLDER_PATTERN);
-				return tokens.length == 1;
-			}).collect(Collectors.toList());
+			List<String> keys = builderMap.keySet().stream().filter(
+					key -> !Pattern.compile(PLACEHOLDER_PATTERN).matcher(key).find())
```
spencergibb: Could `Pattern.compile()` be moved to a constant?

ryanjbaxter: Good call!

marnee01:
Hi, Spencer. Thanks for addressing this so quickly. Sorry for my delay. I was OOO for a few days. Which build snapshot(s) is this change in? (I tried Greenwich.BUILD-SNAPSHOT. The fix does not seem to be present in that snapshot.)

spencergibb:
It will be in Greenwich and Hoxton shortly

PR #1539, opened 2020-01-18
Ensures that config token provider has no overriding bean names

marcingrzejszczak:
without this change when adding the vault token `spring.cloud.config.server.vault.token` we have 2 config token providers on the classpath, the `VaultConfiguration#configTokenProvider` and the `EnvironmentRepositoryConfiguration$DefaultConfigTokenProvider` both with the same bean name. Starting from Boot 2.2 (AFAIR) that's unacceptable by default.

with this change we're changing the bean name so that we don't override the other bean name.


marcingrzejszczak:
I've forgotten to set `spring.profiles.active=vault` :grimacing:

marcingrzejszczak:
I've forgotten to set `spring.profiles.active=vault` :grimacing:

spencergibb:
Cherry-picked to 2.1.x and merged forward via d1579518c7261875376c2932f01ede9681fce979

Issue #661, opened 2017-03-03
Allowing composite config server to ignore failures and continue processing


spencergibb:

warroyo:
is there a plan to add this in any upcoming releases?

ryanjbaxter:
Hope so, but havent started work on it yet.

warroyo:
@ryanjbaxter I just got back into doing some things with the config server and ran into this issue again.  has there been any progress on this? looking at the docs It seems like it is still an issue.

spencergibb:
It is, but if there's an exception, later repos won't be queried, iirc.

ryanjbaxter:
OK I understand now, thanks.

shopey:
Hi, is this planned to be resolved in any upcoming release? it been hanging around for a while :)

jaynsuthar:
Hello,
Is there any plan to add this in near future? It seems to be a security concern to provide the token every time the client wants to query something from config server even though the client only needs to query from the git repository.

habuma:
I'm working on a PR for this and have considered a few options. Bear in mind that my interest in this is being able to continue to serve Git properties even though the Config Server may have a Vault backend and the client hasn't sent a Vault token on the request. 

Two things I've tried and both work (to varying degrees):

- In `CompositeEnvironmentRepository.findOne()`, catch any exceptions thrown by calling `findOne()` on the subordinate repositories. The `catch` block I've written doesn't do anything but log the failure. In the end, the failing backend simply doesn't appear in the response. This option's benefit is that it works for any failure case, not just Vault backends. The downside (or, arguably a benefit) is that there's nothing other than the log message to let the client even know that there is another backend that could've provided properties, but it was unreadable for some reason. I was thinking of also adding an empty `PropertySource` to the environment, but at that point in the code there's not much info available to construct a good name for the property source (e.g., something like "vault:myapp"). Of course, one option is to expose a `getName()` method on `EnvironmentRepository`, although that's a biggish change for such a small benefit.
- In `AbstractVaultEnvironmentRepository.findOne()` (and likewise, in other environment repos that might fail), catch any exceptions and add an empty property source with the same name that is used in the success case.

I think that the first option is the one that's probably best, although the second option gives a bit more info to the client. It would also be nice if there were some other property in `PropertySource` to communicate failure cases. But that's only if it's deemed important to even include an empty property source in the response. 

[...]

Issue #610, opened 2017-01-13
Spring cloud config server - health check is broken when using vault 


vasilievip:
When I provide vault token, I'm getting keys from vault and git:
```
curl -X "GET" "http://localhost:8080/myservice.properties" -H "X-Config-Token: 27004d54-f1ac-31eb-572e-7fb1445a6be7"
coupons.setting1: placeholder1.value
placeholder1: placeholder1.value
vaultkey1: secret
```
When I go to health check, I'm getting service down status due to missing token when traversing status of vault repo:
```
"configServer": {
"status": "DOWN",
"repository": {
"application": "app",
"profiles": "default"
},
"error": "java.lang.IllegalArgumentException: Missing required header: X-Config-Token"
},
```

spencergibb:
thanks for the report

vasilievip:
Vault repo has required token from config client. Not sure about design decisions which led to this. Why git integration don't ask config clients for login/password to git repo? 
Would it be acceptable to add some sort of "deafultVaultToken" and, if config client did not provide one, use default instead?

```
VaultEnvironmentRepository.java
...
	String read(String key) {

		String token = request.getHeader(TOKEN_HEADER);
		if (!StringUtils.hasLength(token)) {
			throw new IllegalArgumentException("Missing required header: " + TOKEN_HEADER);
		}
...
```

ryanjbaxter:
If the git repo requires a username and password you can supply that in the configuration
```
spring:
  cloud:
    config:
      server:
        git:
          uri: https://github.com/spring-cloud-samples/config-repo
          username: trolley
          password: strongpassword
```

spencergibb:
it's specifically the vault token

vasilievip:
> If the git repo requires a username and password you can supply that in the configuration

correct, this user/password will be configured on config server side, but vault integration requires token to be configured on config client side (e.g. each and every from my 100500 services must supply such token :). Maybe it makes sense to setup vault token on config server side for vault as its done for git?
This will also fix health endpoint.

spencergibb:
@vasilievip the point of the vault integration is that each client has a unique token.

jwcarman:
So, the best practice is to set up a token for each application that needs to use the configuration server?  Just trying to understand the intent here.  How do folks propose we secure that application-specific token on the client side?  Environment variables?  Also, when I start up my configuration server, it appears that it is expecting the spring.cloud.vault.token value to be set.  It throws this exception:

[...]

PR #1545, opened 2020-01-24
Return an empty property source for Vault if not token given.

habuma:
This PR includes changes to return an empty property source for a config server with a Vault backend if the client doesn't send an `X-Config-Token` header, rather than return an error.

This addresses several challenges when working with Vault backends, not the least of which are:

 - If a config server is backed by both Vault and Git and has several clients, *all* clients currently must be configured with a Vault token, even if they only require properties from Git. This seems like a security concern to provide Vault tokens to entities that do not need them. The only alternative is to create two config servers, one with a Vault backend and one without.
 - If a config server is backed by both Vault, the Actuator health endpoint will always be "DOWN" unless the request to the health endpoint includes the `X-Config-Token` header (which would be unusual for such a client of that endpoint to even be aware of the token). (Separately, it might be wise for the health indicator to be more Vault-savvy and invoke the Vault server's own health endpoint which doesn't require a token. But this PR doesn't address that.)

Note that the response will include Vault property sources even in the absence of a Vault token. But they will be empty and the property source name will have a parenthetical comment describing why they are empty. I'm open to suggestions on how better to communicate that to the client.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -8,7 +8,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-config</artifactId>
-		<version>3.0.0.BUILD-SNAPSHOT</version>
+		<version>2.2.2.BUILD-SNAPSHOT</version>
 		<relativePath>..</relativePath>
 	</parent>
 
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -8,7 +8,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-config</artifactId>
-		<version>2.2.2.BUILD-SNAPSHOT</version>
+		<version>3.0.0.BUILD-SNAPSHOT</version>
 		<relativePath>..</relativePath>
 	</parent>
 
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.springframework.cloud</groupId>
 	<artifactId>spring-cloud-config</artifactId>
-	<version>3.0.0.BUILD-SNAPSHOT</version>
+	<version>2.2.2.BUILD-SNAPSHOT</version>
 	<packaging>pom</packaging>
 	<name>Spring Cloud Config</name>
 	<description>Spring Cloud Config</description>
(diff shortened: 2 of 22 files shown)


spencergibb:
before I review, we've branched for Ilford. If you want this for Hoxton, rebase to 2.2.x and retarget.

scottfrederick:
@habuma Vault authentication can be provided with a few different mechanisms now. 

The legacy `VaultEnvironmentRepository` supports auth via the `X-Config-Token` header from a client request or via a `spring.cloud.config.server.vault.token` property in a config server. 

The newer `SpringVaultEnvironmentRepository` supports the `X-Config-Token` header as well as multiple authentication options with a `spring.cloud.config.server.vault.authentication` and all the [support provided by Spring Cloud Vault](https://cloud.spring.io/spring-cloud-vault/reference/html/#vault.config.authentication.vault-agent).

If I'm reading this change correctly, the added try/catch in the base class shared between these two implementations would get triggered with any authentication failure, regardless of the authentication method being used. If that's the case, then the message `(empty because X-Config-Token is missing)` could be misleading.

spencergibb:
Ping @habuma RE: retarget to 2.2.x

habuma:
> Ping @habuma RE: retarget to 2.2.x

@spencergibb Hmmm...I rebased to 2.2.x, but probably shouldn't have done that while listening in on a meeting. I actually didn't expect it for Hoxton, as it kinda changes the behavior enough that I wouldn't expect it for an SR or minor release. I was expecting for Ilford. So, perhaps I should undo my rebase? Before I do that, please advise on what you'd do. Ilford is fine. 

@scottfrederick Yes, I think you're right. So perhaps the message should be different. I'm thinking something more generic authentication-wise, but still hinting that the token might be the reason. What do you think?

scottfrederick:
> So perhaps the message should be different. I'm thinking something more generic authentication-wise, but still hinting that the token might be the reason. What do you think?

Yeah, maybe as simple as "because authentication was not provided in X-Config-Token header or in sever configuration" or something similar. Not sure how long that text could/should be given that it's in the name of a property source.

spencergibb:
Two things, I think @ryanjbaxter found a could alternative for the health indicator. https://github.com/spring-cloud/spring-cloud-config/issues/610#issuecomment-593462365

Also, I'd rather do something like #661 which is general-purpose as opposed to specific to vault.

spencergibb:
Closing if favor of #661

PR #1546, opened 2020-01-27
Use LinkedHashSet for deterministic iterations

contextshuffling:
Test `org.springframework.cloud.config.monitor.GiteePropertyPathNotificationExtractorTests#giteeSample` depends on `extract` in `BasePropertyPathNotificationExtractor`. In the function, it converts a HashSet to an array. But HashSet does not guarantee any specific order of entries. Therefore, the assertion will fail if the iteration order differs.

This PR proposes to change the HashSet to LinkedHashSet for a deterministic iteration order (same as insertion order).


spencergibb:
Cherry-picked to 2.1.x and merged forward via 44c4b0e401b6fb8414db2e43ff9e48114cb0d8b8

PR #1564, opened 2020-03-05
Add X-Vault-Namespace also to auth requests. Would get 400 Bad Reques…

rosuvlad:
Add X-Vault-Namespace also to auth requests. Would get 400 Bad Request - Invalid Role ID otherwise

PR #1566, opened 2020-03-06
Fix for namespace not being propagated on Auth schemes (except Token)

rosuvlad:
Add X-Vault-Namespace also to auth requests. 
Would get 400 Bad Request - Invalid Role ID (for AppRole auth) otherwise.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultClientConfiguration.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultClientConfiguration.java
@@ -96,9 +96,7 @@ public class SpringVaultClientConfiguration extends AbstractVaultConfiguration {
 				endpointProvider, requestFactory);
 
 		if (vaultProperties.getNamespace() != null) {
-			restTemplateBuilder.customizers(
-					restTemplate -> restTemplate.getInterceptors().add(VaultClients
-							.createNamespaceInterceptor(vaultProperties.getNamespace())));
+			restTemplateBuilder.customizers(this::applyNamespaceInterceptor);
 		}
 
 		return restTemplateBuilder;
@@ -125,6 +123,13 @@ public class SpringVaultClientConfiguration extends AbstractVaultConfiguration {
 		return new SslConfiguration(keyStoreConfiguration, trustStoreConfiguration);
 	}
 
+	@Override
+	public RestOperations restOperations() {
+		RestTemplate restOperations = (RestTemplate) super.restOperations();
+		applyNamespaceInterceptor(restOperations);
+		return restOperations;
+	}
+
 	private SslConfiguration.KeyStoreConfiguration getKeyStoreConfiguration(
 			Resource resourceProperty, String passwordProperty) {
 
@@ -138,6 +143,15 @@ public class SpringVaultClientConfiguration extends AbstractVaultConfiguration {
 		}
 
 		return SslConfiguration.KeyStoreConfiguration.of(resourceProperty);
+	}
+
+	private RestOperations applyNamespaceInterceptor(RestTemplate restTemplate) {
+		if (vaultProperties.getNamespace() != null) {
+			restTemplate.getInterceptors().add(VaultClients
+					.createNamespaceInterceptor(vaultProperties.getNamespace()));
+		}
+
+		return restTemplate;
 	}
 
 	/**


rosuvlad:
Using:

springCloudVersion: Hoxton.SR1
org.springframework.boot: 2.2.4.RELEASE
org.springframework.cloud:spring-cloud-config-server: 2.2.1.RELEASE
We are having Enterprise Vault available in our company and we need to use the namespace segregation.

The problem is that the namespace is enforced for Auth endpoints (except the root one, token) in our Vault configuration.

In our case AppRole would be something specific for each namespace and the current state of spring-config-server is not passing X-Vault-Namespace in the calls for the AppRole flow.
By running the code without the fix, we would encounter a 400 - BadRequest - Invalid Role ID from the Vault backend.

rosuvlad:
Original wrongfully targeted PR: https://github.com/spring-cloud/spring-cloud-config/pull/1564

spencergibb:
@scottfrederick or @mp911de?

kamalakarp:
@scottfrederick @mp911de - Can you please let us know if we need to do anything to push this PR ?

kamalakarp:
I was referring the same issue in https://github.com/spring-cloud/spring-cloud-config/issues/1589

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultClientConfiguration.java
```diff
@@ -125,6 +123,13 @@ public SslConfiguration sslConfiguration() {
 		return new SslConfiguration(keyStoreConfiguration, trustStoreConfiguration);
 	}
 
+	@Override
+	public RestOperations restOperations() {
```
mp911de: Note: The need to override this method should be considered a workaround. Can be removed once spring-projects/spring-vault#546 is resolved.

mp911de:
Looks good from my perspective. The PR even revealed an inconsistency in Spring Vault.

scottfrederick:
This change looks good to me, but there should be a tested added such that the test would have failed before this change but succeeds after the change.

kamalakarp:
It requires enterprise vault which is licensed version and namespaces option not available in open source vault.

mp911de:
Enterprise vault can run up to 30 minutes without a valid license to test namespaces. However, a namespace test arrangement requires a bit of CI and test fixture setup, see https://github.com/spring-projects/spring-vault/blob/master/spring-vault-core/src/test/java/org/springframework/vault/core/VaultNamespaceSecretIntegrationTests.java#L84-L125

kamalakarp:
@rosuvlad - is this something you want to do on your fork  ?

rosuvlad:
I can give it a try to create some tests to cover this but it will take some time. I am not proficient with this pipeline.

scottfrederick:
@rosuvlad I can work on adding a test when the PR is merged if you'd prefer. Just let me know what you'd like to do.

rosuvlad:
Thank you Scott! I would really like that!

scottfrederick:
@rosuvlad This is merged with [one small change](https://github.com/spring-cloud/spring-cloud-config/commit/959f492ddf77cbfe84dc4d3f6633ebf4118019d7#diff-b1d3f3a0748166d6a2b94ac7db493ed6R133-R143). Instead of overriding `AbstractVaultConfiguration#restOperations` and explicitly adding the namespace interceptor there, this change overrides the method with an exact copy of the change that @mp911de [applied to Spring Vault](https://github.com/spring-projects/spring-vault/commit/e1d8c51999efd1a4bbb87394364a7ea3c58a888f#diff-7068d8b46418a1a6cd18c84dbe473bac). With the new tests in place, this will make it easy for us to delete this method override later when Spring Cloud Config pulls in a Spring Vault upgrade.

rosuvlad:
Cool! Thank you for the support offered! Have a great day ahead!

kamalakarp:
Thank you @scottfrederick @rosuvlad for the support.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultClientConfiguration.java
higonetr: Hi,
I need to implement it with namespaces. There is a better way to do it? 

`@Override
public RestOperations restOperations() {
    RestTemplate restTemplate = (RestTemplate) super.restOperations();
    // To define a namespace
    List<ClientHttpRequestInterceptor> clientHttpRequestInterceptorList = new LinkedList<>();
    ClientHttpRequestInterceptor namespaceInterceptor = VaultClients.createNamespaceInterceptor(NAMESPACE);
    clientHttpRequestInterceptorList.add(namespaceInterceptor);
    restTemplate.setInterceptors(clientHttpRequestInterceptorList);
    //

    return restTemplate;
}`

PR #1573, opened 2020-03-11
Expansion of Vault Azure MSI Configuration properties and bug fix.

justbert:
This merge request is to align Spring Cloud Config Server's Vault Repository using Azure MSI authentication with recent changes in Spring Cloud Vault (https://github.com/spring-projects/spring-vault/pull/542). It allows for expanded configuration of the Azure MSI Authentication Options for non-standard scenarios. 

I believe I've also fixed a bug that prevented the Azure Auth path from being configured via properties resulting in the authentication path always using the default in the AzureMsiAuthenticationOptions.


justbert:
Hmmm... Checkstyle is complaining for Config Server, though not for things I have touched. Anything I should do?

spencergibb:
Cherry-picked to 2.2.x with some polish and a test via fa39c01

PR #1588, opened 2020-03-25
Deterministic failure/warning with application- prefixed names

### LINKED ISSUE #1586: Better issue triage when ${spring.application.name} prefixed with "application-"
The [current documentation](https://cloud.spring.io/spring-cloud-static/spring-cloud-config/2.2.2.RELEASE/reference/html/#_locating_remote_configuration_resources) clarifies the unexpected behavior when `${spring.application.name}` of an application that uses Spring Cloud Config is prefixed with `application-`:

> When setting the property ${spring.application.name} do not prefix your app name with the reserved word application- to prevent issues resolving the correct property source.

While this note has proved useful to me in the past, I have seen team members struggle to debug the issue when application name starts with `application-` multiple times now. I can imagine others might be in the same boat. For users, who use Spring Config Server everyday, going back to the documentation to find this note might not be very intuitive.

(assuming `${spring.application.name}` starts with `application-`)
In the Spring Cloud Config Client before sending the request to the server:
- If failfast is enabled, throw an exception indicating application name cannot start with `application-`
- If failfast is _not_ enabled, log a warning and return an null from `ConfigServicePropertySourceLocator.locate(..)`

Walliee:
resolves #1586

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -35,6 +35,7 @@ import org.springframework.boot.origin.OriginTrackedValue;
 import org.springframework.cloud.bootstrap.config.PropertySourceLocator;
 import org.springframework.cloud.bootstrap.support.OriginTrackedCompositePropertySource;
 import org.springframework.cloud.config.client.ConfigClientProperties.Credentials;
+import org.springframework.cloud.config.client.validation.InvalidApplicationNameException;
 import org.springframework.cloud.config.environment.Environment;
 import org.springframework.cloud.config.environment.PropertySource;
 import org.springframework.core.annotation.Order;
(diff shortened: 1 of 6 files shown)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -89,6 +91,20 @@ public ConfigServicePropertySourceLocator(ConfigClientProperties defaultProperti
 	public org.springframework.core.env.PropertySource<?> locate(
 			org.springframework.core.env.Environment environment) {
 		ConfigClientProperties properties = this.defaultProperties.override(environment);
+
+		if (StringUtils.startsWithIgnoreCase(properties.getName(), "application-")) {
+			InvalidApplicationNameException exception = new InvalidApplicationNameException(
+					properties.getName());
+			if (properties.isFailFast()) {
+				throw exception;
```
spencergibb, line 99: I'm not sure this is what I had in mind. logging the warning is fine. I was thinking if there was some common exception that happens we could analyze it, but not fail.

Walliee: Sorry for the delayed response on this. 

That's the thing no exception/error response is returned by config server. The server returns results that are not consistent with the expectation. So here we are adding a check in the client; If fail-fast is enabled, throw an exception which will prevent application startup, otherwise log a warning and return null.

Issue #1574, opened 2020-03-12
Support integration with AWS Secrets Manager


Walliee:
[AWS Secrets Manager](https://docs.aws.amazon.com/secretsmanager/latest/apireference/Welcome.html) is an AWS service that enables you to store, manage, and retrieve, secrets.

Similar to #850, which talks about support for AWS Parameter Store; supporting AWS Secrets Manager behind the Config Server abstraction would be a useful feature.

Walliee:
Happy to take a stab at it if the team agrees.

spencergibb:
Not sure we'd want one without the other

tejas-pandilwar:
@spencergibb I am going to start working on this [feature](https://github.com/tejas-pandilwar/spring-cloud-config).

Issue #308, opened 2015-12-18
Failed to walk directory: \\C\Users\user\git\config-repo


cforce:
I get following error for a local git repo using the monitor feature.
with cfg: 
```yml
spring: 
  profiles: default,local,native

  cloud: 
    config: 
      server: 
        native: 
          searchLocations: "[file://${user.home}/dev-cfg-folder]"
        git: 
          cloneOnStart: true
          basedir: target/config
          uri: file://${user.home}/git/config-repo

```
2015-12-18 15:39:03,985 ERROR org.springframework.cloud.config.monitor.FileMonitorConfiguration (FileMonitorConfiguration.java:291) - Failed to walk directory: \\C\Users\user\git\config-repo

dsyer:
Try `file:///${user.home}/dev-cfg-folder` (no brackets and an extra "/").

PR #1598, opened 2020-04-23
Add support for the AWS Systems Manager Parameter Store

### LINKED ISSUE #850: Possible integration with AWS Parameter Store
Hello,

is there anything programmed about an integration between Spring Cloud Config and AWS Parameter Store?
Since it looks like AWS tool for storing configurations, I thought it would be nice having some integration between that and Spring Cloud Config.

iulian-antohe:
Hello folks!

I took a stab at adding support for the AWS Systems Manager Parameter Store as a configuration backend. Due to limitation on parameter value length which can be up to 4 KB for the Standard tier and 8 KB for the Advanced tier, the Parameter Store is not suited for storing configuration files. Still, it's ideal for storing Key-Value data which can be further encrypted with KMS.

Design: for a given application and profile(s) a unique ordered list of parameter hierarchies is computed (e.g. `/config/application-default/`, `/config/service-production/`) first. Next, for each hierarchy, an AWS Parameter Store API call is issued for the retrieval of all parameters within that hierarchy and a new property source containing the returned parameters is added to the Environment. This way the returned parameters are always up-to-date and considering the Parameter Store API limit of up to 1,000 requests per second, there shouldn't be a problem even for large scale deployments.

Things I've done:

- Updated the AWS SDK version to make use of STS & ECS container credentials and `getParametersByPath` SSM module method.
- Added unit tests and the new test suite class in `AdhocTestSuite`.
- Added implementation and more tests to achieve high code coverage.
- Authentication was left to the discretion of the [Default Credential Provider Chain](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default).
- Both main and test code was formatted according to the Spring Cloud style.
- Documented the Parameter Store configuration backend.
- Tested manually and all the automated tests have passed subsequent local runs.

Things that have been intentionally left out:
- Labels: I wasn't too sure about the possibility of linking the AWS parameter label (which is used mostly for identifying AWS resources) and the configuration label. Can the label be the AWS account ID or number from which the parameters are retrieved?
- Versions: since parameters' versions are not in sync (e.g. `/config/application/foo` can be at version 10 whilst `/config/application/bar` can be at version 3), setting the environment version to be the greatest or the lowest found amongst all parameters it's not really representative.

Please review and let me know if anything else is needed.

Fixes #850


tejas-pandilwar:
@iulian-antohe just looking to understand why did you decide to use aws-java-sdk instead of [AWS Parameter Store Support](https://github.com/spring-cloud/spring-cloud-aws/blob/master/spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocator.java) provided by spring-cloud-aws?

iulian-antohe:
@tejas-pandilwar As a matter of fact, the class you pointed out (which is part of [spring-cloud-aws](https://github.com/spring-cloud/spring-cloud-aws) pull request [#308](https://github.com/spring-cloud/spring-cloud-aws/pull/308) authored by @jkuipers) was originally intended for this project. However, as @spencergibb pointed out in issue #850 it _does not belong in this project_ and _it would make more sense as a module in spring-cloud-aws_.
There are also some subtle differences:

- the [spring-cloud-aws](https://github.com/spring-cloud/spring-cloud-aws) Parameter Store support:
  - implements `PropertySourceLocator` and it is geared towards retrieving a `PropertySource` that is loaded during the 'bootstrap' phase. If the property doesn't exist or it cannot be accessed and the fail-fast flag is set then the application bootstrap is stopped.
  - is designed to be used by a single Spring Cloud application to retrieve Key-Value properties specific to that application only, along with the shared properties directly from AWS Parameter Store.
  - doesn't have support for advanced filtering, parameter versions and labels.
- this implementation:
  - implements `EnvironmentRepository` and it is geared towards retrieving an `Environment` with multiple property sources (e.g. Parameter Store properties and overrides). If one of the parameters doesn't exist or it cannot be accessed, it simply won't be added to the environment and the config server won't crash.
  - is designed to be used by many applications simultaneously (can be non-Spring applications) and upon every Config server request will return the most up-to-date properties. In this case, the Config server will act as a proxy between all applications and AWS Parameter Store. It would bring value in a composite setup e.g. regular properties from a Git repository and sensitive properties (passwords, keys, tokens etc.) from AWS Parameter Store.
  - can have support for advanced filtering, parameter versions and labels depending on the review of this PR.
  - follows the existing model of retrieving properties from AWS S3.

Considering all of the above, it seemed appropriate to go the aws-java-sdk way.

However, there's no point in denying that this implementation and the [spring-cloud-aws](https://github.com/spring-cloud/spring-cloud-aws) Parameter Store support are very similar indeed for the part of retrieving parameters from the Parameter Store. I think instead of reusing the [spring-cloud-aws-parameter-store-config](https://github.com/spring-cloud/spring-cloud-aws/tree/master/spring-cloud-aws-parameter-store-config) in this project, it would be best (from a DRY perspective) to move the core functionality into a separate `spring-aws-parameter-store` project (like [spring-vault](https://github.com/spring-projects/spring-vault) and [spring-credhub](https://github.com/spring-projects/spring-credhub)) and then reuse it in both [spring-cloud-aws](https://github.com/spring-cloud/spring-cloud-aws) and this project.

tejas-pandilwar:
@iulian-antohe thanks for the explanation. I am looking into [Issue #1574](https://github.com/spring-cloud/spring-cloud-aws/blob/master/spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocator.java) spring-cloud-config integration with AWS Secret Manager and wanted to be consistent in my approach.

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -893,6 +905,99 @@ Configuration files are stored in your bucket as `{application}-{profile}.proper
...
+|no
+|`aws:ssm:parameter:`
+|The prefix that is added to the property source's name to show their provenance.
+
+|*prefix*
+|no
+|`/config`
+|Prefix indicating L1 level in the parameter hierarchy for every property loaded from the AWS Parameter Store.
+
+|*profileSeparator*
```
spencergibb, line 968: `profile-separator`. We should refer to properties in kebab-case in docs.

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -893,6 +905,99 @@ Configuration files are stored in your bucket as `{application}-{profile}.proper
...
+|no
+|`-`
+|String that separates an appended profile from the context name.
+
+|*recursive*
+|no
+|`true`
+|Flag to indicate the retrieval of all AWS parameters within a hierarchy.
+
+|*decryptValues*
```
spencergibb, line 978: `decrypt-values`

docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -893,6 +905,99 @@ Configuration files are stored in your bucket as `{application}-{profile}.proper
...
+|no
+|`true`
+|Flag to indicate the retrieval of all AWS parameters with their value decrypted.
+
+|*maxResults*
```
spencergibb, line 983: `max-results`.

docs/src/main/asciidoc/spring-cloud-config.adoc
iulian-antohe: Happy to oblige. But I think the suggested changes are not consistent with the existing docs format as camel case is used in these places and a few others:
https://github.com/spring-cloud/spring-cloud-config/blob/615f7521b4235bb8f1e11d8f333d95df1c3ef7b7/docs/src/main/asciidoc/spring-cloud-config.adoc#L416-L434
https://github.com/spring-cloud/spring-cloud-config/blob/615f7521b4235bb8f1e11d8f333d95df1c3ef7b7/docs/src/main/asciidoc/spring-cloud-config.adoc#L605-L614
https://github.com/spring-cloud/spring-cloud-config/blob/615f7521b4235bb8f1e11d8f333d95df1c3ef7b7/docs/src/main/asciidoc/spring-cloud-config.adoc#L722

spencergibb:
ping @maciejwalkowiak here as well.

docs/src/main/asciidoc/spring-cloud-config.adoc
iulian-antohe: @spencergibb @maciejwalkowiak ☝️

lentus:
@spencergibb  @iulian-antohe  my team and I are eagerly awaiting this feature, any updates?

iulian-antohe:
@lentus I've synced with the latest upstream changes. The PR is still awaiting a review/guidance from @spencergibb and @marcingrzejszczak

Issue #816, opened 2017-10-06
Question - Possibility to enhance cloneOnStart to not delete directory


tipsytopsy:
Hello,

We're using cloud config server with a git backend and our current setup of bitbucket cluster doesn't make it to be considered HA (because it is not clustered across multiple datacenters). So I would like to support a failover mode in cloud config.

I would ideally like to use the "cloneOnStart" feature, but since it deletes and recreates the base directory - it would mean that a config server has a hard dependency on bitbucket to even start the server. How can I enhance it to not delete the base directory and rather do a hard reset on start up, so the previous state of all my git repos will keep the application functioning?

The current implementation of the environment repository makes it very tough to enhance it due to most of the logic being in private methods.

Please advice.

Thanks!

tipsytopsy:
Wanted to add - I know that clone-on-start=false and setting a basedir will allow the server to serve from cache, but there is a possibility that a node in a datacenter would not have received any traffic for some time and *may* not have cloned the repo yet. So best scenario would be to have clone-on-start but not delete basedir if it exists, just do a reset --hard. Does it makes sense?

gmcouto:
Yeah, it would be very useful to have a "fetch-on-start" that works only when baseDir is set. So when the config-server is restarted for some reason, it would try to update the local repository.

I was able to replicate this with this code though:
```
	HttpHeaders createHeaders(String username, String password){
		return new HttpHeaders() {{
			String auth = username + ":" + password;
			byte[] encodedAuth = Base64.encodeBase64(
					auth.getBytes(Charset.forName("US-ASCII")) );
			String authHeader = "Basic " + new String( encodedAuth );
			set( "Authorization", authHeader );
		}};
	}

	@Value("${server.port}")
	String port;

	@Value("${spring.security.user.password}")
	String password;

	@EventListener
	public void refreshCache(ApplicationReadyEvent event) {
		log.info("STARTED: Fetching latest data");
		RestTemplate template = new RestTemplate();
		template.exchange("http://localhost:"+port+"/application/default", HttpMethod.GET, new HttpEntity<Map>(createHeaders("user", password)), Map.class);
		log.info("FINISHED: Fetching latest data");
	}
```
It just listen to application ready event, then probe itself through REST as if it were another application. Not elegant, but will do for now.

artemptushkin:
I've just oped a [pr](https://github.com/spring-cloud/spring-cloud-config/pull/1608) that changes this logic to:

[...]

PR #1608, opened 2020-04-30
Clone on start optimization

artemptushkin:
gh-816
When `closeOnStart = true` and `.git` directory already exists then do not clone remote repo and open a connection to an existing one. Otherwise, clone on start as usual


ryanjbaxter:
Closing due to inactivity

PR #1678, opened 2020-07-31
Server-side resolve Vault secrets inside application's property sources

donbeave:
This feature will add the ability to replace property value with the value from Vault KV secret.

It works similarly to how `{cipher}` works (https://cloud.spring.io/spring-cloud-config/reference/html/#_encryption_and_decryption).

For example, if activated `vault` profile and `application.yml` contains some keys with `{vault}` prefix:

```yaml
spring:
  datasource:
    username: dbuser
    password: '{vault}:dbs/myapp#pass'
```

Spring Cloud Config server on the server-side will request the Vault server will load secret with the path `dbs/myapp` and retrieve the value from `pass` key and will put it in `spring.datasource.password` property.

Same as `{cipher}`, if a value cannot be parsed, it is removed from the property source and an additional property is added with the same key but prefixed with invalid and a value that means “not applicable” (usually <n/a>).


donbeave:
@spencergibb ok, sorry for that. Maybe I better make a separate PR for these `static`/`final` changes?

spencergibb:
focused PRs are much better. `static` and `final` have no material impact of adding new vault encryptor support.

donbeave:
@spencergibb done, pls check again. Build and tests are success on my local machine:

image

It looks like probably CircleCI is not configured properly for this repository, because the error is not related to my code changes.

PR #1712, opened 2020-09-23
Don't throw exception when a repository can't be read [#661]

habuma:
This change simply ignores (but logs) any exceptions that take place when a composite repository is reading from one of its subordinate repositories. 

This addresses the issue where a Config Server has been configured with a Vault backend, but the client may not have a Vault token to request properties with. Currently, lack of a valid Vault token results in an error. With this change, properties from other backends will still be served and the Vault backend(s) will be absent from the response. 

As I noted in a comment on #661, it might be desirable and possible to include the unreadable backends in the response, albeit with no properties. But I'm going to go ahead and submit this PR and if there is reason to include empty property sources in the response, then I'll accept that as useful feedback.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepository.java
@@ -23,7 +23,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.springframework.cloud.config.environment.Environment;
-import org.springframework.cloud.config.environment.PropertySource;
 import org.springframework.core.OrderComparator;
 
 /**
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepository.java
@@ -37,15 +37,18 @@ public class CompositeEnvironmentRepository implements EnvironmentRepository {
 
 	protected List<EnvironmentRepository> environmentRepositories;
 
+	private boolean failOnError;
+
 	/**
 	 * Creates a new {@link CompositeEnvironmentRepository}.
 	 * @param environmentRepositories The list of {@link EnvironmentRepository}s to create
 	 * the composite from.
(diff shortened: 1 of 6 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepository.java
```diff
@@ -19,6 +19,9 @@
 import java.util.Collections;
 import java.util.List;
 
+import org.slf4j.Logger;
```
spencergibb, line 22: commons logging

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepository.java
```diff
@@ -60,7 +65,12 @@ public Environment findOne(String application, String profile, String label, boo
 		}
 		else {
 			for (EnvironmentRepository repo : environmentRepositories) {
-				env.addAll(repo.findOne(application, profile, label, includeOrigin).getPropertySources());
+				try {
+					env.addAll(repo.findOne(application, profile, label, includeOrigin).getPropertySources());
+				}
+				catch (Exception e) {
+					log.info("Error adding environment for " + repo);
```
spencergibb, line 72: After some discussion, the team would like this to be opt-in via configuration. The new behavior could surprise users and it's only a single property on config server, not clients.

So 
```
if log errors = true
   log error
else
  rethrow
```

spencergibb (changes requested):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CompositeEnvironmentRepository.java
habuma: What would you suggest as a good name for such a configuration property? `spring.cloud.config.server.fail-on-composite-error` with a default of `true`? It's a little wordy, but all other names I came up with were equally wordy or not clear. And...that's why I'm asking in case someone has a better idea for a name.

spencergibb: that name is fine.

spencergibb (approved):

Issue #1643, opened 2020-06-25
Remove ResourceController mappigs with param useDefaultLabel


spencergibb:
ResourceController has multiple `@RequestMapping`s that would conflict, currently to avoid passing the label we use the `useDefaultLabel` request parameter to select. Starting with spring 5.3 this is no longer happening.

Given
```
	@RequestMapping("/{name}/{profile}/{label}/**")
// and 
	@RequestMapping(value = "/{name}/{profile}/**", params = "useDefaultLabel")
```
Calling `/foo/default/file.txt?useDefaultLabel` used to route to the later, now it routes to the former and produces an error because the file is treated as a label.

spencergibb:
After writing some [samples](https://github.com/bclozel/mapping), it appears that using the `useDefaultLabel` request parameter to select a different `@RequestMapping` hasn't worked for some time. It appears that no one is using this as no one is reporting problems. This method will get removed instead.

dinhani:
I was starting to use the the ResourceController for serving plain-text files, but it did not work.
Is the support for plain-text files being removed or just migrated to something else?

spencergibb:
@dinhani no, this is to remove specific use of a query parameter named `useDefaultLabel` that doesn't work anyway.

ThomasVitale:
@spencergibb can I help with this task?

spencergibb:
@ThomasVitale sure

ThomasVitale:
I've just created a PR: https://github.com/spring-cloud/spring-cloud-config/pull/1728
I've removed both mappings in `ResourceController` using the `useDefaultLabel`. Is it ok?

PR #1728, opened 2020-10-20
Remove ResourceController mappings with param useDefaultLabel

ThomasVitale:
`ResourceController` has multiple `@RequestMappings` that would conflict. Currently, to avoid passing the label, we use the `useDefaultLabel` request parameter to select. Using that parameter to select a different @RequestMapping doesn't work. The two mappings using the parameter have been removed now.

Fixes gh-1643

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/resource/ResourceController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/resource/ResourceController.java
@@ -106,13 +106,6 @@ public class ResourceController {
 		return retrieve(request, name, profile, label, path, resolvePlaceholders);
 	}
 
-	@RequestMapping(value = "/{name}/{profile}/**", params = "useDefaultLabel")
-	public String retrieve(@PathVariable String name, @PathVariable String profile, ServletWebRequest request,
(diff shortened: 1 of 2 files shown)


adcis:
@ThomasVitale 
Documentation still claims this property can be used
https://cloud.spring.io/spring-cloud-config/multi/multi__serving_plain_text.html

However it was a very useful function. How can I now fetch plain text from "default label" when config client is not aware of label name?

adcis:
Hi,

Actually I used this function a lot in older spring version, but now wanted also to use this https://github.com/spring-cloud/spring-cloud-config/pull/1417. Looks like after upgrading to version that includes this feature, `useDefaultLabel` stopped working, do you have any workaround that would avoid client to "be aware" of default label? 

I'd like to serve a plain text with encryption from default branch that the client is now aware of.

alftheo:
I created myself a controller with requestmapping
`@RequestMapping(value = "/{name}/{profile}/{path:.*}", params = {"useDefaultLabel"})`

and that works fine. Can we add that into the project?

spencergibb:
@alftheo please open a new issue

alftheo:
I've created a new issue https://github.com/spring-cloud/spring-cloud-config/issues/1823 with an attached pull request. Let me know if something is missing bad or wrong

Issue #1729, opened 2020-10-25
Cloud Config Server should be able bootstrap resources


rolaca11:
**Is your feature request related to a problem? Please describe.**
When the `spring.cloud.config.server.bootstrap` property is set to `true`, the config-server should configure itself like it was any other config-client, which it does well, as long as you don't need a resource, like a `logback-spring.xml` during bootstraping.
This is not a problem for clients, because I can just use a config like this: `logging.config=${spring.cloud.config.uri:http://localhost:8888/}/${spring.application.name}/${spring.profiles.active:default}/master/logback-spring.xml` and the client will send another request to the config server asking for this resource.

The problem is that during config-server bootstraping, this endpoint is not yet available.

**Describe the solution you'd like**
I don't know if there are other usecases (not logging reinitialization related) where this issue arises. If there aren't (which I suspect is the case) than the solution is simply to try and request the resource bypassing the endpoint, for example by invoking `this.repository.findOne(name, profiles, label, includeOrigin);` directly.

> _code taken from org.springframework.cloud.config.server.environment.EnvironmentController:136_

If there are other usecases, we might need a more sophisticated solution.

**Describe alternatives you've considered**
I have considered some hacky solutions by defining root search locations, but these are shady and situational.

spencergibb:
After discussion with the team, we've decide the premise is good. We would need something general that works in many cases. It might require some design on our side first.

PR #1732, opened 2020-11-01
Allow config server to bootstrap resources implements #1729

rolaca11:
implements #1729


spencergibb:
Hmm... I'm unsure about an implementation as such. Marking for team discussion.

rolaca11:
I'm not 100% confident either. Maybe this is a too special solution for this one case only.

Which part are you not sure about?

spencergibb:
This exactly:
> Maybe this is a too special solution for this one case only.

rolaca11:
How could we extend it?

Do you think introducing a `resource:/[resource-from-server]` be a better solution? It could be introduced in the client side either for requesting resources from the config server

spencergibb:
After discussion with the team, this is too specific, though the premise is good. We would need something more specific. It might require some design on our side first.

PR #1743, opened 2020-11-10
Do not discard the root exception when things go wrong

aclement:
The root exception can detail an important root cause. For
example with native-images it can indicate a flag you need
to set for the native-image to compile correctly:
Caused by: java.net.MalformedURLException:
  Accessing an URL protocol that was not enabled. The URL
  protocol http is supported but not enabled by default.
  It must be enabled by adding the -H:EnableURLProtocols=http
  option to the native-image command.


spencergibb:
Cherry-picked to 2.2.x and merged forward via bf2c61f72220b7c5df0df658da7bd353af8eb915

PR #1828, opened 2021-03-10
Fix actuator for sample

wind57:


wind57:
just trying to fix the sample application so that `/actuator/env` works, plus update the documentation. Let me know if this makes sense or more comments are needed. thank you.

spring-cloud-config-sample/pom.xml
```diff
@@ -56,6 +56,10 @@
 			<groupId>org.springframework.retry</groupId>
 			<artifactId>spring-retry</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
```
spencergibb, line 60: This will break tests. This module is less of a sample and more of integration tests.

spencergibb (changes requested):
Remove changes in spring-cloud-config-sample and leave the documentation.

spring-cloud-config-sample/pom.xml
wind57: thank you for the feedback. I wanted to change the `sample` because of this in the documentation : _see the test cases for the config-client or the sample application_. That `or the sample application` is what I thought is about a "sample". Should that be removed from the doc then?  or there is a sample in some other place?

docs/src/main/asciidoc/quickstart.adoc
```diff
@@ -163,25 +163,24 @@ spring.application.name: myapp
...
-    "name": "servletContextInitParams",
-    "properties": {}
-  },
-  {
-    "name": "configserver:https://github.com/spring-cloud-samples/config-repo/foo.properties",
-    "properties": {
-      "foo": {
-        "value": "bar",
-        "origin": "Config Server https://github.com/spring-cloud-samples/config-repo/foo.properties:2:12"
+      "name": "bootstrapProperties-https://github.com/spring-cloud-samples/config-repo/foo.properties",
```
spencergibb, line 174: This is not what properties look like when use `spring.config.import` which is now the default.

wind57: I am a little bit confused by your comment. I took the latest `master` from `spring-cloud-config`, started `spring-cloud-config-server` and ran : `curl localhost:8080/actuator/env` from a client that has the server enabled.

The output is exactly as in this PR. Sorry, but what am I missing?

PR #1833, opened 2021-03-15
Feature(Vault KV2): to support a path after mount-path (backend) under which the key(s)/application(s) be found

sasiperi:
Hello,

**Usecase/Problem**
Please note that below is a problem that can not be worked around for KV2 Vault APIs, due to the "data" that exactly need to go right after mount-path. 
1. In big corporations, we have needs, to group secret types together (e.g All-API-Secrets; All-Secrets-for-Pipelines for Team/ORG within the company) and manage policies accordingly.
2. Then there are use cases such that we create a separate mount-path for each type of secrets then application-level secrets (apps and app,profile) for an ORG can go under a separate path under (after) mount-path. So that we do not need to put all keys(applications) across ORGs(or teams) right under-mount path.
3. For example something like KV2 mount-path = cloud-api-secrets; under which we will have a path "ORG-1 (or Team-A)" and all applications (keys) belonging to ORG-1 will be created under org-1. I can also have ORG-2(or Team-B) as another path and all applications(keys) belonging to ORG2 will be created under org-2
- That means an effective path where keys can be found for org-1 would be /cloud-api-secrets/data/org-1
- That means an effective path where keys can be found for org-2 would be /cloud-api-secrets/data/org-2
3. The current version of the config-server does not support this. As it detects KV1 or KV2 based on a kvVersion prop, getPath() method in then KV2Strategy class returns "backend/data/{key}" as the effective path.
4. I made an enhancement to support a "path" between mount-path (backend) and "keys" (that's a folder/path under which a specific Org/Team's keys/apps can be found).

**Changes made**
1. I introduced another prop for KV2 called pathToKey (a path after mount-path, under which keys/apps can be found, with an empty default value  (i.e. by default it searches for keys right under mount-path(backend).
2. If pathToKey is configured, then conditionally for KV2, this path inserted between backend and key as "backend/data/path2key/{key}
- KV1 this is not a problem as *data* does not need to be included at a specific place in the API call.
4.  I made changes accordingly.
5. Fixed tests as needed.
6. Added a new test case called  "testVaultKV2WithPath2Key" for this use case alone.
7. Did "checkstyle" validations and committed to this feature branch, that I forked from "spring-cloud-config" master.

**Other points to be noted (considerations)**
- This will work even if the path is nested for example mount-path/data/path2Keys where path2Keys = nameA/nameA1 (effective path e.g. /secrets/data/nameA/nameA1/{key}) as long as all "keys" intended for that team/org are right under the path,  keys will be searched recursively under this path (e.g. application, application,profile1... for all profiles) 
- But *will not* recursively search in all paths for keys, in other words, **it is not equivalent to** to searchPaths feature in GIT Backend. It will only look for keys right under "pathToKeys" only (if nothing configured for this prop, it will search right under mount-path).
- I considered implementing a recursive search through multiple nested folders like searchPaths, but it felt like making it overly complicated, making it slower with KV2 (kv2 is already slower than kv1).  
- Plus I can not think of a use case as to why one would require "search folders" for secrets and make "vault policies" management maintenance complex as well with too many levels of nesting. OR if policies are created only at Team/ORG, I did not see a valid need for nesting.

Please let me know if this can be accepted?

Regards
sp

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultEnvironmentProperties.java
@@ -52,19 +52,20 @@ public class VaultEnvironmentProperties implements HttpEnvironmentRepositoryProp
 
 	/** Vault backend. Defaults to secret. */
 	private String backend = "secret";
-		
 
 	/**
 	 * The key in vault shared by all applications. Defaults to application. Set to empty
 	 * to disable.
 	 */
 	private String defaultKey = "application";
-	
-	/** 
-	 * KV2 API required "data" after "mount-path". There could be folder/path structure, where the keys/applications are grouped.
-	 * This property is the path after mount-path, under which application(s) are located (appended after "data")
-	 * Default value is blank, which means all grouped applications are located right under the mount-path
-	 *  
+
+	/**
+	 * KV2 API required "data" after "mount-path". There could be folder/path structure,
+	 * where the keys/applications are grouped. This property is the path after
+	 * mount-path, under which application(s) are located (appended after "data") Default
+	 * value is blank, which means all grouped applications are located right under the
+	 * mount-path
+	 *
 	 */
 	private String pathToKey = "";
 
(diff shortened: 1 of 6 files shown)


pivotal-issuemaster:
@sasiperi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1833)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1833) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sasiperi Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-config&pullRequestId=1833)!

jagan23527001:
When this fix will be available to use?. We have same issue when using Kv2 vault.

spencergibb:
We have a milestone planned later this month, but it won't be released until November

Issue #1834, opened 2021-03-15
Activate a different profile from within a default one


Haybu:
**Describe the bug**
Currently (with SB > 2.4.x) a profile that gets activated from within a default profile section is not taking effect. A remote application gets configured still with the default profile's properties. 

Properties source for the activated profile should be reloaded. However with SB 2.4.x, it's not possible to specify `spring.profiles.active` from within a default section, see [this](https://github.com/spring-projects/spring-boot/issues/24890) issue. However, that is possible now with SB 2.5.x (snapshot now).

The attached PR works only with remote client applications that use SB 2.5.x

PR #1835, opened 2021-03-15
reload properties of remotely activated profile fixes gh-1834

Haybu:


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -69,14 +69,20 @@
...
 	private ConfigClientProperties defaultProperties;
 
+	private final String ACTIVE_PROFILES_PROPERTY_NAME = "spring.profiles.active";
+
 	public ConfigServicePropertySourceLocator(ConfigClientProperties defaultProperties) {
 		this.defaultProperties = defaultProperties;
 	}
 
 	@Override
-	@Retryable(interceptor = "configServerRetryInterceptor")
```
spencergibb, line 77: I think it needs to stay here for some rare backwards compatibility issues since the call to `locate(properties)` wouldn't be retryable.

spencergibb (commented):
Maybe an test? or you could show me a failing sample?

ryanjbaxter:
Should we wait until we fork the repo since this is only compatible with boot 2.5.x?  How would we test it?

spencergibb:
yeah, this will have to wait a bit. Not sure we're going to fork for 2.5.x. There should be compatibility tests. Maybe there's a way to only run a test only for a specific version of boot.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -69,14 +69,20 @@
...
 	public org.springframework.core.env.PropertySource<?> locate(org.springframework.core.env.Environment environment) {
 		ConfigClientProperties properties = this.defaultProperties.override(environment);
+		return locate(properties);
+	}
+
+	@Retryable(interceptor = "configServerRetryInterceptor")
+	public org.springframework.core.env.PropertySource<?> locate(ConfigClientProperties properties) {
```
spencergibb, line 85: Is the only usage of this the method above?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -138,6 +155,15 @@ public org.springframework.core.env.PropertySource<?> locate(org.springframework
 
 	}
 
+	// to check if different profile is activated within the default profile
+	private boolean checkIfProfileIsActivatedInDefault(String[] profiles, Map<String, Object> map) {
+		List<String> profilesList = Arrays.asList(profiles);
+		return (profilesList.size() == 1 && profilesList.get(0).equalsIgnoreCase("default")
```
ryanjbaxter, line 161: @spencergibb do you recall why this would only need to be done for the default profile?  What if the default profile contains `spring.profiles.active` and then subsequent configuration properties which get activated based on that profile also contain `spring.profiles.activate`?  In that case we would not also activate those profiles.

ryanjbaxter:
Fixed by https://github.com/spring-cloud/spring-cloud-config/pull/2260

PR #1844, opened 2021-03-24
Use optional in classpath search location

### LINKED ISSUE #1843: config-server with native profile serves application.yml
According to the docs [here](https://cloud.spring.io/spring-cloud-config/reference/html/#_file_system_backend) config-server should not serve its own `application.properties` (or yml presumably) to clients, even if `native` profile is active.
It seems that Ilford config-server has a regression on this point.
I've attached sample config-server and config-client apps. The client fails to start when the server is running as it tries to use the `server.port` set in the server's application.yml.

This is blocking the spring-cloud-services tile upgrade to boot 2.4 and Ilford as we use the native profile when a config-server is created with no explicit backing config repo. Without this, a user can't create an empty config-server and update it afterwards with backing config.

[config-server-native-bug.zip](https://github.com/spring-cloud/spring-cloud-config/files/6192498/config-server-native-bug.zip)

ryanjbaxter:
Fixes #1843

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
@@ -51,8 +51,8 @@ import org.springframework.util.StringUtils;
  */
 public class NativeEnvironmentRepository implements EnvironmentRepository, SearchPathLocator, Ordered {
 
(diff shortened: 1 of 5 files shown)


ryanjbaxter:
I'm not really sure if this is the best approach...

ryanjbaxter:
cc: @dyroberts

spencergibb:
Interesting. Should they all be optional?

spencergibb:
There should probably be a new test, since we didn't catch this before

ryanjbaxter:
@spencergibb other locations now using `optional` and test added as well

spencergibb (approved):

PR #1865, opened 2021-04-19
Update documentation around discovery first lookup

### LINKED ISSUE #1854: Discovery First configuration doesn't work without bootstrap enabled
In the old 2.x.x versions we used the discovery first bootstrap for loading configurations for our services and we wanted to keep this at the new 3.0.x version.

We followed the instructions on your site: https://docs.spring.io/spring-cloud-config/docs/3.0.0/reference/html/#discovery-first-bootstrap
but failed to successfully load configurations via the discovery until the bootstrap was enabled.

So, until we added the dependency in our service:

        <dependency>
            <groupId>org.springframework.cloud</groupId>
            <artifactId>spring-cloud-starter-bootstrap</artifactId>
        </dependency>

we couldn't load configurations via the discovery.

We also tried to use the spring.config.use-legacy-processing=true property but without luck.

So did we miss something in the properties or does the bootstrap really needs to be enabled so the discovery first works, and it is just missing in the documentation?

ryanjbaxter:
There needs to be a `spring.config.import` statement for this to work

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -1516,10 +1516,7 @@ The net result of this behavior is that all client applications that want to con
 
 If you use a `DiscoveryClient` implementation, such as Spring Cloud Netflix and Eureka Service Discovery or Spring Cloud Consul, you can have the Config Server register with the Discovery Service.
 
-Using the `DiscoveryClient` to locate the config server requires that the config client *must use* <<config-first-bootstrap,config first bootstrap>> and *cannot use* <<config-data-import,config data import>>.
-
 If you prefer to use `DiscoveryClient` to locate the Config Server, you can do so by setting `spring.cloud.config.discovery.enabled=true` (the default is `false`).
-The net result of doing so is that client applications all need a `bootstrap.yml` (or an environment variable) with the appropriate discovery configuration.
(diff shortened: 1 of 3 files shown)


docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -1512,14 +1512,14 @@ When a config client starts, it binds to the Config Server (through the `spring.
 The net result of this behavior is that all client applications that want to consume the Config Server need a `bootstrap.yml` (or an environment variable) with the server address set in `spring.cloud.config.uri` (it defaults to "http://localhost:8888").
 
 [[discovery-first-bootstrap]]
-==== Discovery First Bootstrap
+==== Discovery First Lookup
 
-If you use a `DiscoveryClient` implementation, such as Spring Cloud Netflix and Eureka Service Discovery or Spring Cloud Consul, you can have the Config Server register with the Discovery Service.
+WARNING: Unless you are using <<config-first-bootstrap, config first bootstrap>>, you will need to have a `spring.config.import` property in your configuration properties.
+For example, `spring.config.import=optional:configserver:`.  This functionality will be changed in a future release.
```
spencergibb, line 1518: Remove 'This functionality will be changed in a future release.' An import will always be required. No need to comment on future enhancements.

spencergibb (changes requested):

spencergibb (approved):

PR #1887, opened 2021-05-24
"default-label" should now have "main" as value

adrhc:
`default-label` should now have `main` as value


spencergibb:
superseded by #1998

PR #1904, opened 2021-06-09
Support new delimiters in spring.config.location

### LINKED ISSUE #1895: Wrong property source order when properties are in nested folders
**Describe the bug**

After I upgraded spring cloud config server from version 2.3.10 to 2.4.5 the order in which properties files are processed changed. Before the upgrade properties from profile specific files, e.g. `{application}-{profile}.properties`, had precedence over general files like `{application}.properties`, and after the upgrade it is the opposite. 

I am using cloud config server with `spring.cloud.config.server.git.searchPaths` option set to `'**'` so that I can have properties files structured in directories, like this:

- dir1/{application}.properties
- dir1/dir2/{application}-{profile}.properties

I also checked this with v2.4.7-SNAPSHOT and it is the same there

**Sample**

I created a simple cloud config server based on `https://start.spring.io/` code generated . Set the content of `application.properties` to

```
spring.cloud.config.server.git.uri: https://github.com/grabarczyk-t/cloud-config-test.git
spring.cloud.config.server.git.searchPaths: '**'
```

In the git repository I have two property files:

- `project/application.properties`
- `project/test/application-test.properties`

both with just one property `sample.property`, with different value in each file. When I run the server locally and make request to `http://localhost:8080/application/test` I get:

In the `2.3.10` version:

```
{"name":"application","profiles":["test"],"label":null,"version":"32248076b1c298d26c0f4afc9f6464f0e452e18f","state":null,"propertySources":[{"name":"https://github.com/grabarczyk-t/cloud-config-test.git/project/test/application-test.properties","source":{"sample.property":"value from application-test.properties"}},{"name":"https://github.com/grabarczyk-t/cloud-config-test.git/project/application.properties","source":{"sample.property":"value from application.properties"}}]}
```

In the `2.4.5` version:

```
{"name":"application","profiles":["test"],"label":null,"version":"32248076b1c298d26c0f4afc9f6464f0e452e18f","state":null,"propertySources":[{"name":"https://github.com/grabarczyk-t/cloud-config-test.git/file:C:\\Users\\Tom\\AppData\\Local\\Temp\\config-repo-14384905862201453365\\project\\application.properties","source":{"sample.property":"value from application.properties"}},{"name":"https://github.com/grabarczyk-t/cloud-config-test.git/file:C:\\Users\\Tom\\AppData\\Local\\Temp\\config-repo-14384905862201453365\\project\\test\\application-test.properties","source":{"sample.property":"value from application-test.properties"}}]}
```

I also checked it with a simple cloud config client application. I have profile set to `test` (`spring.profiles.include=test`), and when I check the value of this property using `@Value`:

```
@Value("${sample.property}") private String sampleProperty;
```

the `sampleProperty` has the value from `application-test.properties` when the 2.3.10 config server is running, and value from the `application.properties` when the `2.4.5` config server is running

ryanjbaxter:
Fixes #1895
Requires Boot 2.4.7


spencergibb (changes requested):
how about a test?

spencergibb (approved):
Could delete the hooks from test repo, but LGTM

philwebb:
If back compatibility is an issue then you could pick the delimiter based on the presence of the `ConfigDataLocation.split()` method.

ryanjbaxter:
I think we are fine with not being compatible with releases prior to Boot 2.4.7.

PR #1926, opened 2021-07-08
Make GitCredentialsProviderFactory a bean

### LINKED ISSUE #1896: provide the git username/password at run time(dynamically)
Hi,

From what I found so far, It seems that the only way to provide the git repo username/password
can be either on the property file or as an environment variable. 
Both ways are unacceptable by our Infosec.  For example if someone will hack and will connect to the K8s pod
he can access the files or the environment variables. 
We need to retrieve it at runtime from vault.

Encrypt them also requires to put the secret somewhere. From the example I found so for it seems it need to be saved in the bootstrap file, so again we are in the problem of saving the secret on a file.  
Is there a way I can pass the username/password dynamically at runtime?

Or if that not possible to pass the secret?

Thanks,
Avner

ryanjbaxter:
Fixes #1896


spencergibb (approved):
Can't believe we missed this.

avnerstr:
@ryanjbaxter  and @spencergibb 

I'm not sure this fix will help to solve composite repos.
when  I shifted to 

  cloud:
    config:
      server:
        composite:

structure

I started to get errors:

Invocation of init method failed; nested exception is java.lang.IllegalStateException: You need to configure a uri for the git repository.

Issue #1845, opened 2021-03-25
Try multiple URLs when error response code received


marnee01:
We want to configure our clients with two URLs for two config servers - each one pointing to a different git management server. If one git management server goes down, the client should receive an error then try the second URL. However, the conditions for retry are too limited for our needs:

1. The main issue is that if git server can not be reached, config server returns 404. (404 seems inappropriate, because it is not a client error. However, I would guess changing it now would not be feasible considering it changes existing behavior.) When 404 is received, the client does not try the second URL. It makes sense not to do a retry for a 4XX error conceptually, since they should be client errors. However, git server not being reached is not a client issue, and we want the client to try the next URL in that case.

2. Another issue is with 500 errors. The documentation states, "if the Config Server returns a 500 (Internal Server Error) response...the Config Client does not try to fetch properties from other URLs". However, 500 is a server-side error, not client error. Although this is not our main concern, it seems best to have client try next URL. (Perhaps the first URL points to a config server with a defect, but second URL points to a newer or older one without the defect, so trying the other one could be beneficial.)

**Proposed Solution**
Allow for configuring a list of HTTP response code and/or response code series for which client will try the next URL in the list. The current behavior could remain as default behavior.

Note: Another reason for 404 is if client passes bad label. For us, this is usually only in non-prod environments. I'd rather have the client try next URL when it's a bad label (although unlikely to change the response), than to not retry when the 404 is due to git server being down.

marnee01:
A simpler solution would be a single flag to enable trying multiple URLs on any error response, even though it would mean trying when the issue is clearly a client-side error. So:

**Default/current behavior**: Only try the next URL in the list if `ResourceAccessException` is caught.
**New behavior**: If property XYZ is set to true, then also try the next URL if `HttpClientErrorException` is caught.

Note that the code is in the `ConfigServicePropertySourceLocator.getRemoteEnvironment(...)` method.

marnee01:
This is very important to us. If I can get approval to spend the time, I may try to implement this and push the change to the project.  However, it would need to get pushed back to spring-cloud-config-client 2.1, I believe. Most of our services are still at spring boot 2.1 and won't be upgraded any time soon. (Our upgrade steps not trivial.)

I've never done this before. How feasible would it be for me to patch it back to previous versions or to have you do the same with my change to the latest?

reboot-armani:
@marnee01 has this been merged to the latest config client version? or version can I find it in?

[...]

PR #1946, opened 2021-08-11
Add support for multiple-url-strategy configuration for clients.

marnee01:
Add support for new property spring.cloud.config.multiple-uri-strategy.
The value must be one of: always, connection_timeout_only. The default
value is connection_timeout_only. This default setting maintains existing
behavior.

If a client has multiple URLs in the spring.cloud.config.uri property,
and if multiple-uri-strategy is set to "always", then if the client gets
any error from config server whatsoever or gets no response, the client
will try the next URL in the list.

The existing and default behavior is that the other URLs in the list will be
tried only if and when the client receives no response from config server.

This is mainly to allow for a client failing over to secondary URLs when
it receives a 404 due to config server being unable to reach its git server
(and config server does not have the requested configs in its local git cache).
With the default behavior, the client receives a 404 and never tries the
next URL in the list(which might point to a different git server that is
currently up). Another benefit is that it allows for a client failing over
to a secondary config server if the first one returns a 503 (OUT_OF_SERVICE).

Fixes gh-1845.


marnee01:
This would also need a documentation update. If and when this approach is approved, please point me to the repo. I will make that change, also (unless the normal process is for owners to update documentation).

marnee01:
I am working on adding the missing tests. However, I am struggling to get the project to find some of the dependencies. I have maven 3.8.1 and the latest Eclipse with Spring Tools Suite. (Maven and Eclipse are new to me.) I set the Maven profile to "spring" in Eclipse across all projects. (Properties -> Maven -> Active Maven Profiles). I also set it as an active profile in Maven's settings.xml. But it gives attached errors, unable to find snapshot and milestone dependencies and other errors. I can't select Maven profile for the specific project via RC -> Maven -> Select Maven Profiles.... When I try that, it gives the same error as when project is being loaded. (I also tried to load in Intellij with same issue)

Other than the README, is there any further documentation on how to get project set up in Eclipse? I had no luck searching for a resolution.

I can build and run tests at commandline, but that's very slow-going. (No IDE help, just text editing the old fashioned way.)

[ProjectLoadErrors.txt](https://github.com/spring-cloud/spring-cloud-config/files/8092038/ProjectLoadErrors.txt)

ryanjbaxter:
Are you building against java 17?

marnee01:
@ryanjbaxter Can you tell me where `ConfigServerConfigDataLoader` is used? The `getRemoteEnvironment` methods in `ConfigServerConfigDataLoader` and `ConfigServicePropertySourceLocator` are almost identical. I updated `ConfigServerConfigDataLoader` with the same changes I made to `ConfigServicePropertySourceLocator`, but don't see a unit test for `ConfigServerConfigDataLoader` and don't know how to functionally test it because I see no usages of `ConfigServerConfigDataLoader`. There are no javadocs on either class.

marnee01:
I don't think there are any tests that already cover multiple URIs in `ConfigServerConfigDataLoader` for me to update. I don't have enough knowledge of spring integration tests and this project to add tests from scratch without probably days of work which I am not prepared to do right now. I was able to do manual testing to cover this using an actual spring boot client app. When `ConfigServerConfigDataLoader` was originally introduced for the new Spring Boot 2.4 way to import configuration data (https://github.com/spring-cloud/spring-cloud-config/pull/1656/files), no unit test was added that I can see.

Unless someone else wants to add the missing test(s) to cover `ConfigServerConfigDataLoader` or this can be merged without tests covering that code, this will probably have to remain open for now. Another alternative is if someone can point me to a sample integration test that does not use Docker and which I could easily clone and update to cover multiple URIs; I didn't see one that jumped out at me, but maybe one exists. (There are no javadocs on integration tests. It's not even obvious to someone unfamiliar with the project what is being tested/asserted in these integration tests.)

marnee01:
@spencergibb @ryanjbaxter The tests should be good now. Apologies for missing the tests. I remember writing tests (although I didn't add tests for `ConfigServerConfigDataLoader` because when it was originally introduced with https://github.com/spring-cloud/spring-cloud-config/pull/1656, adding the unit test for that was missed). I have now created `ConfigServerConfigDataLoaderTests`, also. I'm thinking I missed a commit when fiddling with my fork last year.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
```diff
@@ -116,6 +116,13 @@
 	 */
 	private String[] uri = { "http://localhost:8888" };
 
+	/**
+	 * The strategy to use when call to server fails and there are multiple URLs
+	 * configured on the uri property (default
+	 * {@link MultipleUriStrategy#CONNECTION_TIMEOUT_ONLY}).
+	 */
+	private MultipleUriStrategy multipleUriStrategy = MultipleUriStrategy.CONNECTION_TIMEOUT_ONLY;
```
ryanjbaxter, line 124: Since this is going into main I wonder if the strategy can be `ALWAYS` by default?  @spencergibb what do you think?

ryanjbaxter (changes requested):
Can you add some documentation on the new property?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java
spencergibb: Sounds good to me

marnee01: Will change it. (I was thinking to keep current behavior, but I like ALWAYS as the default better.)

ryanjbaxter: We can introduce breaking changes in the main branch

marnee01:
@ryanjbaxter @spencergibb I was testing this with a little local client app today with my latest change. When I have:

```
spring.config.import=configserver:
spring.cloud.config.uri=https://configserverdev.pearsonvue.com/ConfigServer, \
  https://configserverfailoverdev.pearsonvue.com/ConfigServerFailover
```
The first URL it tries is the default URL `http://localhost:8888`. Since I have `spring.cloud.config.uri`, it shouldn't be trying that one. 

This was not happening last week. (I have the old logs to prove it.) Maybe I fiddled with something in the configs to cause this to happen, but I think that my `application.properties` is still the same, based on local history in my IDE. Did something change in the `main`? I don't see how my changes could have broken this. I tested my changes which are working, so I'll go ahead and push the requested change, but wanted to ask about this.

Another thing that seems to have changed is that when you don't have `optional` in `spring.config.import=configserver:` or you have have fail-fast=true, it does not produce a log file showing each URL that it tries. They only way I can tell that it is trying multiple URLs as expected is to put a breakpoint in `ConfigServerConfigDataLoader`. This was also not the case last week. (Again, my log files show this.) Did this also change? Not producing a log could make it difficult to add alerts or otherwise troubleshoot failures to connect to config server.

Again, this might be due to some local change I made (i.e. user error), but if so, I can't spot it.

ryanjbaxter:
I wonder if it was this change...
https://github.com/spring-cloud/spring-cloud-config/commit/cee78ddd86f0ea9b9dd943f19d16afaa0c5e59f6

marnee01:
@ryanjbaxter Oh, boy. I opened up a can of worms :-)

1. I got the latest and it is still hitting localhost. Do you want me to submit a new issue for that? Also, for the issue of not producing a log if it fails to connect to configserver, should I submit an issue?
2. When I provide two URLs as a YAML list under `spring.config.import` (instead of `spring.cloud.config.uri`), the **second** URL listed is tried first. Is the order in which they are tried indeterminate?  If it should hit them in order, do you want me to submit an issue?
3. The value `multiple-uri-strategy` for the connection timeout only is `connection_timeout_only`, with underscores, since that is the name of the enum constant. . It just occurred to me that I don't recall any spring-related property values using underscores. Is there something else you'd prefer?
4. Even with the latest code, if multiple URLs are provided using `spring.config.import` and all would cause connection timeout, still only one is called if fail-fast is true and/or config server is not optional. (By contrast, when fail-fast is false and configserver is optional, then after getting timeout on one, it will try the next one.) This seems to be because with this version of configs, it calls ConfigServerConfigDataLoader.load twice - once for each URL. So the logic regarding multiple URLs in ConfigServerConfigDataLoader.getRemoteEnvironment [here](https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java#L311) does not take effect. And then as soon as first URL fails, it fails fast [here](https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java#L188). Is this the desired behavior?

ryanjbaxter:
1. When I look at the commits from main over the past week, I don't see anything obvious that would cause the change in behavior.  Can you reproduce it without your changes?  Can you reproduce it using 2021.0.1?  If so open a separate issue and we can address that separately.
2. Thats more of a Spring Boot function than anything to do with Spring Cloud. From the Spring Boot docs `Several locations can be specified under a single spring.config.import key. Locations will be processed in the order that they are defined, with later imports taking precedence` so I think this behavior is correct.
3. I would use `-` rather than `_`
4. If something is not labeled `optional` than it is considered required so yes I would expect it to fail to start.  Same with `fail-fast=true`, if we fail on the first URL I would expect it to fail to start.  @spencergibb do you agree?

spencergibb:
I agree

marnee01:
Please also see my above comment from earlier today (if not already).

1. I was able to reproduce the two issues without latest from `main` and without my changes. I submitted  https://github.com/spring-cloud/spring-cloud-config/issues/2051 and  https://github.com/spring-cloud/spring-cloud-config/issues/2052. As for trying to reproduce with 2021.0.1, do you want me to just change the `spring-cloud.version` in the POM file to that?.

ryanjbaxter:
> I was able to reproduce the two issues without latest from main and without my changes. I submitted https://github.com/spring-cloud/spring-cloud-config/issues/2051 and https://github.com/spring-cloud/spring-cloud-config/issues/2052. As for trying to reproduce with 2021.0.1, do you want me to just change the spring-cloud.version in the POM file to that?.

Yes

>Due to relaxed binding, the enum constant can be CONNECTION_TIMEOUT_ONLY, while the value in the application file can be connection-timeout-only (with dashes). This is how I would document it. Is that good enough?

Yup

> I think I see the difference between spring.cloud.config.uri and the spring.config.import configsrver URLs. Under spring.config.import, it will try ALL of the URLs listed, even if first one it tries is successful. Is that correct? That isn't obvious when you're used to working with spring.cloud.config.uri, where each one is tried until one is successful. It looks like you could set fail-fast=true on the "last" URL in spring.config.import (with fail-fast=false for the others) and then it would behave like the old uri property. Is that true? (I tried it and it did not seem to work for me, but I didn't try hard - I might have something not set up quite right.)

`spring.config.import` is a Spring Boot function that we are just plugging into.  There could we a wide range of import sources so it makes sense to continue to try all of them rather than fail on one along the way.  Again Spring Boot controls this functionality not Spring Cloud.  I am not sure how you would set `fail-fast` on a per URL basis.

ryanjbaxter:
> @ryanjbaxter I missed your question about documentation. Yes, I can add documentation. Can you provide instructions as to where and how to update? (Is the process for updating documentation documented somewhere?)

The documentation is here https://github.com/spring-cloud/spring-cloud-config/blob/main/docs/src/main/asciidoc/spring-cloud-config.adoc



> This behavior is very different from using spring.cloud.config.uri. Basically, if you have fail-fast or config server is not optional, then there is no way to provide multiple URLs to try if you specify via spring.cloud.import.

Can you provide a more concrete example?

> I believe you should document this - that every URL will be ignored except the last one And I should document that the new spring.cloud.config.multiple-uri-strategy property does not apply if you specify the URLs via spring.cloud.import. Is there any plan to retire spring.cloud.config.uri?

Not at the moment.

marnee01:
@ryanjbaxter 

I've pushed a commit with documentation updates. It includes documenting the differing behavior between specifying multiple URLs under `spring.cloud.config.uri` versus multiple URLs under `spring.config.import`. (Thanks for the documentation link!)

> [MARNEE] This behavior is very different from using `spring.cloud.config.uri`. Basically, if you have fail-fast or config server is not optional, then there is no way to provide multiple URLs to try if you specify via `spring.cloud.import`.
> 
> [RYAN] Can you provide a more concrete example?

Without my code, if you have a setup such as the following, it will try the 2nd URL (configserver2) and if it gets a connection timeout, it fails immediately and does not try the other URL (configserver1).

```
spring:
    application:
        name: MyApp
    cloud:
        config:
            password: MyPass
            username: MyUsername
            fail-fast: true
    config:
        import: "configserver:"
            - "configserver:https://configserver1.company.com/ConfigServer1"
            - "configserver:https://configserver2.company.com/ConfigServer2"
    profiles:
        active: dev

```
(With the above, if you have `fail-fast: false` and configserver2 times out, it will try configserver1, but of course, if that also fails, the service still starts up.)

By contrast, with the following (using `uri` instead of specifying the URLs in the `import`), it will try the 1st URL (configserver1). If that fails with connection timeout, then it will try the next URL (configserver2):
```
spring:
    application:
        name: MyApp
    cloud:
        config:
            uri: https://configserver1.company.com/ConfigServer1, https://configserver2.company.com/ConfigServer2
            password: MyPass
            username: MyUsername
            fail-fast: true
    config:
        import: "configserver:"
    profiles:
        active: dev
```

ryanjbaxter:
Thanks for the example.  IMO there is not much we can do to make them behave the same because the `spring.config.import` behavior is really controlled by Spring Boot.  I think this OK, its a side effect of moving to `spring.config.import`

@spencergibb curious as to your opinion here.

marnee01:
@ryanjbaxter Thanks for the explanation. I did add documentation explaining the difference. (I realize it's Spring Boot, but I hope we can leave that info there because otherwise I think a lot of people like me will be confused.)

Are there any other changes you'd like from me?

spencergibb:
Yeah, having multiple `configserver:` imports vs multiple uri's for one import is going to behave differently, not much we can do there.

marnee01:
Thanks for all the help and feedback with this one! @ryanjbaxter @spencergibb

PR #1948, opened 2021-08-16
fix related to issue #1896. If there is a use of PatternMatchingJGitE…

avnerstr:
fix related to issue #1896. If there is a use of PatternMatchingJGitEnvironmentRepository repositories,  the GitCredentialsProviderFactory bean has to be set on those repositories as well + add test for that

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -205,7 +205,7 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 		this.gitCredentialsProviderFactory = gitCredentialsProviderFactory;
 	}
 
-	public GitCredentialsProviderFactory getGitCredentialsProviderFactory() {
+	GitCredentialsProviderFactory getGitCredentialsProviderFactory() {
 		return gitCredentialsProviderFactory;
 	}
 
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -205,6 +205,10 @@ public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 		this.gitCredentialsProviderFactory = gitCredentialsProviderFactory;
 	}
 
+	public GitCredentialsProviderFactory getGitCredentialsProviderFactory() {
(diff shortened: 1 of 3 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -205,6 +205,10 @@ public void setGitCredentialsProviderFactory(GitCredentialsProviderFactory gitCr
 		this.gitCredentialsProviderFactory = gitCredentialsProviderFactory;
 	}
 
+	public GitCredentialsProviderFactory getGitCredentialsProviderFactory() {
```
ryanjbaxter, line 208: I dont think this needs to be public if it is just used for tests

ryanjbaxter (changes requested):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
avnerstr: I saw this practice in other cases here - like the  MultipleJGitEnvironmentRepository.getRepos() - it was used only in tests before my change.

What do you suggest instead?

ryanjbaxter: I think if its just used in tests right now there is no need to make it `public`

avnerstr: so to change it to what?  package level?

ryanjbaxter: Yup, that is what I was thinking.

avnerstr: done

avnerstr: done. For some reason tests failed. How do I run them again?

avnerstr:
I'm trying to run the build again without luck. @ryanjbaxter please advice

ryanjbaxter:
Not sure whats going on with the build.  I ran the build without your changes on CircleCi and on our Jenkins servers without an issue.  Can you build it locally?

avnerstr:
@ryanjbaxter  it seems that your request to change the scope to package level cause the build to break. Before that it passed.

avnerstr:
@ryanjbaxter I don't know what happened - now it breaks all the time. I will create another fork tomorrow and will try to create new PR

Issue #1947, opened 2021-08-14
Bug in creation of PatternMatchingJGitEnvironmentRepository


avnerstr:
**Describe the bug**
In your new code, you give the ability to create bean from GitCredentialsProviderFactory
but when working with PatternMatchingJGitEnvironmentRepository. The factory in the PatternMatchingJGitEnvironmentRepository remains the original GitCredentialsProviderFactory. 

I suggest to add to the build method in MultipleJGitEnvironmentRepositoryFactory the line at the end:

repository.getRepos().forEach((name, repo) ->repo.setGitCredentialsProviderFactory(gitCredentialsProviderFactory));

If this fine by you - I will add it to the code

avnerstr:
Yes

avnerstr:
@ryanjbaxter are you going to merge my fix? and any ETA? I prefer not use my own jar in production code

PR #1951, opened 2021-08-18
fix bug according to #1947 - when working with PatternMatchingJGitEnv…

avnerstr:
…ironmentRepository. The GitCredentialsProviderFactory instance in the PatternMatchingJGitEnvironmentRepository remains the original GitCredentialsProviderFactory and not the one set by the build method in MultipleJGitEnvironmentRepositoryFactory.


avnerstr:
@ryanjbaxter - done

Issue #1952, opened 2021-08-20
Return Boolean and Integer Properties when using JDBC backend 


naveenkumarpandian:
**Is your feature request related to a problem? Please describe.**
When Using JDBC as backend - the values are always returned as String. We have node JS applications consuming the properties for config server and it is tedious to iterate through values and parse it. We understand the design is based on data type Value DB column which is most likely Varchar. 

However we do use Postgres and would like to store value as a JSONB datatype - so it would support string,integer and boolean.
{"value":9999} or { "value": true}

**Describe the solution you'd like**
PropertiesResultSetExtractor is a static class which creates a Map<String,String>. It would be good to inject this as a spring bean and define a custom implementation to return a Map<String,Object> 

with PG as backend,
SELECT key,json_value -> 'value' from test where application=? and profile=? and label=?

**Describe alternatives you've considered**
We tried to handle this with below JACKSON settings to false, however it does not work 
spring.jackson.mapper.allow-coercion-of-scalars=false

**Additional context**

PR #1960, opened 2021-09-11
Allow customizating Jdbc PropertyResultSetExtractor - Fixes gh-1952

beatfreaker:
Allow customizating Jdbc PropertyResultSetExtractor - Fixes gh-1952

@ryanjbaxter do you think this is the correct approach?


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JdbcEnvironmentRepository.java
```diff
@@ -57,19 +56,21 @@
...
+	private final PropertiesResultSetExtractor extractor;
 
 	private int order;
 
 	private String sql;
 
 	private boolean failOnError;
 
-	public JdbcEnvironmentRepository(JdbcTemplate jdbc, JdbcEnvironmentProperties properties) {
+	public JdbcEnvironmentRepository(JdbcTemplate jdbc, JdbcEnvironmentProperties properties,
```
ryanjbaxter, line 67: Make this as deprecated and create a new constructor with the addition parameter

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/PropertiesResultSetExtractor.java
```diff
@@ -0,0 +1,28 @@
...
+package org.springframework.cloud.config.server.environment;
+
+import java.util.Map;
+
+import org.springframework.jdbc.core.ResultSetExtractor;
+
+/**
+ * @author Chintan Radia
+ */
+public interface PropertiesResultSetExtractor extends ResultSetExtractor<Map<String, ?>> {
```
ryanjbaxter, line 26: Not sure I understand the need for this interface

ryanjbaxter (changes requested):
Could you submit this agains the 3.0.x branch?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/PropertiesResultSetExtractor.java
beatfreaker: was trying to expose an interface which a developer can implement to provide a custom PropertyResultSetExtractor implementation.
Without this, we have to modify the existing class as below.
```java
public static class PropertiesResultSetExtractor implements ResultSetExtractor<Map<String, Object>> {

		@Override
		public Map<String, Object> extractData(ResultSet rs) throws SQLException, DataAccessException {
			Map<String, Object> map = new LinkedHashMap<>();
			while (rs.next()) {
				map.put(rs.getString(1), rs.getString(2));
			}
			return map;
		}
	}
```

and if anyone wants to provide customer implementation, they have to create a child class and provide its bean.
and using AutoConfiguration class we can control If we want to use `PropertiesResultSetExtractor` or `CustomPropertiesResultSetExtractor`

```java
public static class CustomResultSetExtractor extends JdbcEnvironmentRepository.PropertiesResultSetExtractor {
@Override
public Map<String, Object> extractData(ResultSet rs) throws SQLException, DataAccessException {
	Map<String, Object> map = new LinkedHashMap<>();
	while (rs.next()) {
		map.put(rs.getString(1), rs.getInt(2));
	}
	return map;
}
}


@Bean
@ConditionalOnMissingBean(value = JdbcEnvironmentRepository.PropertiesResultSetExtractor.class)
public JdbcEnvironmentRepository.PropertiesResultSetExtractor propertiesResultSetExtractor() {
	return new JdbcEnvironmentRepository.PropertiesResultSetExtractor();
}
```

Is this the better approach?

ryanjbaxter: You could just make `PropertiesResultSetExctractor` a top level class, that might be easiest, then users can just extend that.  Then use the `@ConditionalOnMissingBean` approach.

PR #1971, opened 2021-10-04
Modified usage of Redis containers for tests

robertmcnees:
Made this change so that spring-cloud-config uses the same library as spring-cloud-gateway for using Redis containers in JUnit tests.


OlgaMaciaszek (approved):
LGTM. CI tests fail due to Docker setup. @robertmcnees  I assume you've tested locally? Please monitor if it passes on Jenkins after the merge. The CI setup probably needs updating.

robertmcnees:
Yes, it ran locally with Docker desktop.  The build failed because CircleCI couldn't find a valid Docker environment.  Is there something I can do so it builds successfully with CircleCI?  I don't recall gateway having difficulty finding a Docker environment in CircleCI.

robertmcnees:
Just made a 3rd major revision.  After talking with @jonatan-ivanov I think this is the one that meets the requirements.  

Docker tests will run by default and will be switched off only when a new profile is activated.  I modified the circleci build script to activate the new profile because the build runs in a container.  `@Testcontainers` cannot be used inside of another container.

Thanks to all for taking a look and helping me find new directions when I got stuck.  Let me know if this approach works.  I think we can do something similar in Gateway (and other Cloud projects?).

A quick overview of where I've been with this so far.

1. Started with only `@Testcontainers(disabledWithoutDocker=true)`.  While this worked, it could decrease our test coverage by mistake if a Docker environment is not available when it should be.
2. Kept the strategy from above, but added another test that would fail if running in a Jenkins environment and Docker is not able to start.  Essentially a quick fix to make sure test coverage didn't go down on the Jenkins build, but other builds still had the same problem that Docker tests could be accidentally missed.
3. Added a new profile in maven that excludes tests that rely on docker.  This profile must be specified, so by default the Docker tests will still run.  Since CircleCi cannot run docker in docker, modified the CircleCI build script to use the profile that skips docker required tests.

jonatan-ivanov (approved):
It looks good, I have one question though:
Is there a reason why we don't want CircleCI to run tests with testcontainers? Is it too slow or flakey?

jonatan-ivanov:
Oh, this answers my previous [question](https://github.com/spring-cloud/spring-cloud-config/pull/1971#pullrequestreview-794679649):
>I modified the circleci build script to activate the new profile because the build runs in a container. @Testcontainers cannot be used inside of another container.

ON the other hand, there is a possibility to use Testcontainers on CircleCI. You can have a [`machine` executor](https://circleci.com/docs/2.0/executor-types/) that runs in a VM and not in a Docker container. [Micrometer does this](https://github.com/micrometer-metrics/micrometer/blob/dca07253eacc81afe93112155c1c4e4084a9988c/.circleci/config.yml#L77-L81), it has a job for the non-testcontainers tests that run in a docker container and it has a separate job for the testcontainers tests that don't.

ryanjbaxter (approved):

.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             - ~/.m2
       - run:
           name: "Running build"
-          command: ./mvnw -s .settings.xml clean org.jacoco:jacoco-maven-plugin:prepare-agent install -U -P sonar -nsu --batch-mode -Dmaven.test.redirectTestOutputToFile=true -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+          command: ./mvnw -s .settings.xml clean org.jacoco:jacoco-maven-plugin:prepare-agent install -U -P sonar,withoutDockerTests -nsu --batch-mode -Dmaven.test.redirectTestOutputToFile=true -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
```
spencergibb, line 26: I think I agree with @jonatan-ivanov that these could run with a `machine` type as they do in gateway.

https://github.com/spring-cloud/spring-cloud-gateway/blob/main/.circleci/config.yml#L4

spencergibb (changes requested):

.circleci/config.yml
robertmcnees: I agree changing the CircleCI build is the ideal scenario.  I tried that with [an early commit](https://github.com/spring-cloud/spring-cloud-config/pull/1971/commits/8dccda6bb941e1832a41070d265fd5da2acbc293) but it caused a lot of errors that were unrelated to this change of introducing Testcontainers.  If I remember correctly I think it dealt with configuring the Java Cryptography Architecture (JCA) on the CircleCI environment.  
I was planning to get this code ready to go first and then revisit changing the CircleCI build with the team later.  But we can certainly tackle that now if that's the preference.

jonatan-ivanov: I see two issues, both are pretty weird:
```
Cannot mkdir: No such file or directory tar: home/appuser
```
I'm not sure why is this happening, Micrometer is doing the same (writing `~/.gradle` on a `machine`).


This one scares me a little:
```
Caused by: java.security.InvalidKeyException: Illegal key size
```
One possible cause of this is not having the "Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy" files installed. The problem with this is that Java 8 has these files since [Java 8u161](https://www.oracle.com/java/technologies/javase/8u161-relnotes.html#JDK-8170157) (4 years), previously you needed to install them. This makes me wonder if `machine: true` will give us some ancient image that has a very old version of Java 8.

Can we try what happens when we [set the machine image](https://circleci.com/docs/2.0/executor-types/#using-machine) to the [latest ubuntu that circleci has](https://circleci.com/docs/2.0/configuration-reference/#available-machine-images)?

robertmcnees: Great idea @jonatan-ivanov !  Specifying the machine image got rid of the errors around the `Illegal key size`.

However to get the build to work I still had to remove the `restore_cache` step.  In a previous build this error came back:

`Cannot mkdir: Permission denied tar: home/appuser/.m2`

Any ideas how to restore the cache?  I've been comparing it to other Spring Cloud projects and can't find any practical difference.

spencergibb: I had the same issue when transitioning gateway to machine. Basically, remove the cache, then add it back. We can do that after it is merged.

Issue #1963, opened 2021-09-15
Config client does not fail fast enough with retry


hpoettker:
When retry is enabled in the config client by putting `spring-retry` on the classpath, and connecting to the config server finally fails, the `ConfigClientFailFastException` is not directly thrown from `ConfigServerConfigDataLoader`. Instead, an `ApplicationListener` on `BootstrapContextClosedEvent` is registered that adds the exception to the application context from which it is picked up later by the auto-configured `ConfigClientFailFastListener`.

This has the disadvantage that the actual throwing of the exception and the resulting application failure occur quite late in the application start-up. For example, if a `@Configuration` class depends on config values provided by the config server, the application likely fails with a `BeanCreationException` before the `ApplicationStartedEvent` is triggered that the `ConfigClientFailFastListener` listens on.

While the application does ultimately fail as intended, it does not fail fast. In particular, the log is not as helpful as expected as the logged warnings and errors do not refer to the effectively swallowed `ConfigClientFailFastException`.

I have implemented a minimal example here: https://github.com/hpoettker/scc-sample

It contains the following class which provides a `MessageProvider` which is just a dummy class:

```
@Configuration
public class ClientConfiguration {

  @Value("${client.config.foo}")
  private String message;

  @Bean
  public MessageProvider messageProvider() {
    return new MessageProvider(message);
  }

}
```

When the applications starts without a running config server, it will fail with the (abbreviated) log:

```
 WARN 86577 --- [main] s.c.a.AnnotationConfigApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'clientConfiguration': Injection of autowired dependencies failed; nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'client.config.foo' in value "${client.config.foo}"
ERROR 86577 --- [main] o.s.boot.SpringApplication               : Application run failed
```

If `spring-retry` is removed from the classpath, the following much more helpful log is provided:

```
[main] ERROR org.springframework.boot.SpringApplication - Application run failed
org.springframework.cloud.config.client.ConfigClientFailFastException: Could not locate PropertySource and the resource is not optional, failing
  at org.springframework.cloud.config.client.ConfigServerConfigDataLoader.doLoad(ConfigServerConfigDataLoader.java:193)
  ...
[...]

PR #1977, opened 2021-10-10
Let config client fail before bean creation

hpoettker:
If the config client fails and fail fast is enabled, the `ConfigClientFailFastException` will already be thrown when the bootstrap context is closed instead of during auto-configuration.

This allows the deferred logs to be printed properly while it avoids failure on the creation of beans that depend on values provided by the config server.

The change simplifies the handling of the `ConfigClientFailFastException` that has been introduced in https://github.com/spring-cloud/spring-cloud-config/commit/e9fa34032b67301831dfe73882bb61b1b2330d43.

Fixes gh-1963.

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientAutoConfiguration.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 the original author or authors.
+ * Copyright 2013-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


hpoettker:
The failed codecov check seems like a false positive to me. It probably got tripped up because `ApplicationFailFastTests` is contained in a different module (`spring-cloud-config-sample`) than the change itself (`spring-cloud-config-client`).

hpoettker:
This PR and the issue it's intended to solve have now been open for quite some time. I still think that the problem described in the issue makes operating the config service in a real world scenario of corporate firewalls harder to troubleshoot than necessary.

Any feedback on the PR would really be appreciated. If it needs work, I'm perfectly willing to resolve any issues.

hpoettker:
I've resolved a merge conflict and rebased the PR on `main`. The code coverage alerts still seem like alse positives to me.

The PR has now been open for a year. And over this time, there have been thumps ups on the issue and the PR and also a supportive comment on the issue.

Is there a known unintended change in behavior that the proposal would cause? If yes, please let me know, and I'll try to change the PR accordingly.

Is the PR only perceived as risky without a clear idea what might break? In that case, the upcoming release of version 4 seems to me like a very good point in time to take this risk.

babubabu:
Is there any progress on this PR? May I help in any ways?
I just debugged today in our CI infrastructure, because the config-client failed and errors were hidden by other exceptions due to missing configuration properties

ryanjbaxter:
This seems to make sense to me.  

@spencergibb what do you think?

spencergibb:
Seems ok to me

babubabu:
Thanks a lot!

Issue #1989, opened 2021-10-22
Lookup spring.config.import entries in subdirectories then Environment type is Native


vananiev:
**Is your feature request related to a problem? Please describe.**
I have a `bootstrap.properties` with native env type enabled
```properties
spring.cloud.config.server.composite[0].type=native
spring.cloud.config.server.composite[0].search-locations=test/{application}/{label}
```
and `application.yml` with imports of yaml files grouped in directories
```yml
spring.config.import:
  - a.yml
  - car/b.yml
  - car/c.yml
  ...
  - airplane/d.yml
  - airplane/e.yml
  ...
```
car, airplane, and other directories are created in production servers if need. So, `application.yml` works with app local configs, but does not work then configs are moved to Spring Cloud Server (client app just starts without `car/*`, `airplane/*` imports without error also in case of `spring.cloud.config.fail-fast=true`).

**Describe the solution you'd like**
Spring Cloud Server finds and reads all config files, but filter out files not directly placed in `spring.cloud.config.server.composite[0].search-locations` by `NativeEnvironmentRepository.matchesLocation():323` method
```java
if (normal.startsWith(finalPattern) && !normal.substring(finalPattern.length()).contains("/")) {
    matches = true;
    break;
}
if (locations != null) {
    return !Arrays.stream(locations).map(this::cleanFileLocation)
        .noneMatch(location -> location.startsWith(finalPattern)
            && !location.substring(finalPattern.length()).contains("/"));
}
```
IMHO, `normal.substring(finalPattern.length()).contains("/")` check is redundant. Why is it needed? IMHO, check only limits functionality without adding value, limits functionality well working with local FS configs.

**Describe alternatives you've considered**
No alternatives. Feature working with local FS imports doesn't work with cloud config.

PR #1990, opened 2021-10-22
Fix spring.config.import for environment type native

vananiev:
Fixes import
```yaml
spring.config.import:
  - a.yml
  - car/b.yml
```
of car/b.yml for Native environment.

Relates gh-1989

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
@@ -314,6 +314,9 @@ public class NativeEnvironmentRepository implements EnvironmentRepository, Searc
 			if (!pattern.contains(":")) {
 				pattern = "file:" + pattern;
 			}
+			if (pattern.startsWith("optional:")) {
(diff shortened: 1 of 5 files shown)


vananiev:
Ok, I'll add tests

ryanjbaxter:
OK I talked this over with the team and we think the change is fine.  Also if you could make the change related to the `return` statement I mentioned above that would be great.

Also could you submit this against the 3.0.x branch?  We will merge it forward to main.

vananiev:
> Also could you submit this against the 3.0.x branch? We will merge it forward to main.

Ok.

vananiev:
> Yes we should only return if its true and not if false

> Also if you could make the change related to the return statement I mentioned above that would be great.

I do not fully understand what I should do with `return`. Can you clarify it?

ryanjbaxter:
if `!Arrays.stream(locations).map(this::cleanFileLocation).noneMatch(location -> location.startsWith(finalPattern)
								&& !location.substring(finalPattern.length()).contains("/"));` does not return `true` we should not return and continue to iterate through the list.

vananiev:
@ryanjbaxter, all done

ryanjbaxter (approved):

PR #1998, opened 2021-11-11
Moves default label to main with fallback to master

### LINKED ISSUE #1722: Consider changing the default branch from master
Recently GitHub changed the default branch name from `master` to `main`. You can find more about it [here](https://github.com/github/renaming).

This breaks the out of the box experience with Spring Cloud Config and it should be considered to change the default name. To minimize the impact on current users, this change should be clearly documented. Maybe there should be a friendly error message when it cannot find the default branch. A default branch name property on the server side could also be useful.

ryanjbaxter:
Fixes #1722


spencergibb (approved):
Could this go in 3.0.x as well?

Issue #1997, opened 2021-11-10
Option.PROFILE_SPECIFIC is missing for profile specific PropertySources from Vault  


marbon87:
PropertySources from cloud config get the wrong order in spring boot applications, if the property sources from cloud config contain profile specific properties from vault and the path in vault is defined as advised in [reference documentation](https://cloud.spring.io/spring-cloud-config/reference/html/#_multiple_properties_sources) with the postfix `,profile`, e.g. `/secret/my-app,profile`

The cloud config server serves the correct order when requesting properties trought http:

```
{
  "name": "my-app",
  "profiles": [
    "profile_a"
  ],
  "label": null,
  "version": null,
  "state": null,
  "propertySources": [
    {
      "name": "vault:my-app,profile_a",
      "source": { }
    },
    {
      "name": "vault:my-app",
      "source": { }
    },
    {
      "name": "ssh://git@mygit.git.git/application-profile_a.yml",
      "source": { }
    },
    {
      "name": "ssh://git@mygit.git.git/application.yml",
      "source": { }
    }
  ]
}
```

But when i look at the environment or health endpoint of the app, that requests the properties, the resulting property sources of the app are:

```
{
  status: "UP",
  components: {
    clientConfigServer: {
      status: "UP",
      details: {
        propertySources: [
          "configserver:ssh://git@mygit.git/application-profile_a.yml",
          "configserver:vault:my-app,profile_b",
          "configserver:vault:my-app",
          "configserver:ssh://git@mygit.git/application.yml",
          "configClient"
        ]
      }
    }
  }
}
```

The problem is that [ConfigServerConfigDataLoader.java](https://github.com/spring-cloud/spring-cloud-config/blob/e645c802157caa88a8ed50ecb9067335f0c7522f/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java#L164-L166) checks if the name of the property source contains `"-profile."` before adding the Option.PROFILE_SPECIFIC.

My suggestion to fix this, is to add another if else-block, that checks if the name starts with "vault" and ends with the profile name.

ryanjbaxter:
Do you specify the order in your config server configuration?  Maybe providing a sample would help.

marbon87:
Here is the example:
https://github.com/marbon87/spring-cloud-config-issue-1997

See the readme.md to get it running.

[...]

PR #2004, opened 2021-11-23
Add Option.PROFILE_SPECIFIC to profile specific PropertySources from Vault

marbon87:
Fixes gh-1997

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
 		<spring-cloud-commons.version>4.0.0-SNAPSHOT</spring-cloud-commons.version>
 		<aws-java-sdk.version>1.11.911</aws-java-sdk.version>
 		<google-api-services-iam.version>v1-rev20201112-1.30.10</google-api-services-iam.version>
-		<testcontainers.version>1.15.1</testcontainers.version>
+		<testcontainers.version>1.16.2</testcontainers.version>
 		<wiremock.version>2.31.0</wiremock.version>
 		<maven-checkstyle-plugin.failsOnError>true</maven-checkstyle-plugin.failsOnError>
 		<maven-checkstyle-plugin.failsOnViolation>true
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java
@@ -161,8 +161,11 @@ public class ConfigServerConfigDataLoader implements ConfigDataLoader<ConfigServ
 							// https://github.com/spring-cloud/spring-cloud-config/issues/1874
 							for (String profile : resource.getAcceptedProfiles()) {
 								// TODO: switch to match
-								if (propertySourceName.contains("-" + profile + ".")) {
-									// TODO: switch to Options.with() when implemented
+								// , is used as a profile-separator for property sources
+								// from vault
+								// - is the default profile-separator for property sources
+								if (propertySourceName.matches(".*[-,]" + profile + ".*")) {
+									// // TODO: switch to Options.with() when implemented
 									options.add(Option.PROFILE_SPECIFIC);
 								}
 							}
--- a/spring-cloud-config-sample/src/test/resources/logback-test.xml
+++ b/spring-cloud-config-sample/src/test/resources/logback-test.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<configuration>
+
+	<contextListener class="ch.qos.logback.classic.jul.LevelChangePropagator">
+		<resetJUL>true</resetJUL>
+	</contextListener>
+
+	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+		<encoder>
+			<pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+		</encoder>
+	</appender>
+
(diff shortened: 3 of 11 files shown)


ryanjbaxter:
There are a number of test failures can you take a look?

marbon87:
Hi @ryanjbaxter, i fixed the formatting failures in the new test class, but i have noch idea why the others tests are failing on circleci.
If i run `maven verify` on my machine, all tests are passing.

marbon87:
Hi @ryanjbaxter, the test is also failing on the main branch. The class org.springframework.security.web.access.WebInvocationPrivilegeEvaluator is missing. I think it is related to last commits from @spencergibb 
image

ryanjbaxter:
Have you pulled in the latest changes?  Main is not passing on circleci.
https://app.circleci.com/pipelines/github/spring-cloud/spring-cloud-config/510/workflows/faecde66-b87f-4a2c-bb33-180cc20057ee/jobs/4120

marbon87:
Looks good now, i rebased on main.

spring-cloud-config-sample/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
ryanjbaxter, line 1: Why is this needed?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java
```diff
@@ -161,8 +162,9 @@ else if (ALL_OPTIONS.size() > 2) {
 							// https://github.com/spring-cloud/spring-cloud-config/issues/1874
 							for (String profile : resource.getAcceptedProfiles()) {
 								// TODO: switch to match
-								if (propertySourceName.contains("-" + profile + ".")) {
-									// TODO: switch to Options.with() when implemented
+								if (propertySourceName.contains("-" + profile + ".") || propertySourceName
+										.matches("^" + PREFIX + "vault:.*" + "," + profile + "$")) {
```
ryanjbaxter, line 166: Can you add a comment about why we need this?  It will help with understanding he code

spring-cloud-config-sample/src/test/resources/logback-test.xml
marbon87: Otherwise Testcontainers is logging in DEBUG mode, which leads to a lot of debug logs from docker

pom.xml
```diff
@@ -122,6 +122,13 @@
 				<artifactId>wiremock-jre8</artifactId>
 				<version>${wiremock.version}</version>
 			</dependency>
+			<dependency>
```
ryanjbaxter, line 125: @robertmcnees can you take a look at this and make sure it aligns with the work you did with TestContainers?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java
spencergibb: I'm very hesitant to add anything repository specific. We should find a more generic way

marbon87: The safest way would be to add an additional argument _profile_ to the property source at the server side and match that on the client against the active profiles. But this is a bigger refactoring. I don't know if you like to make such big changes in a minor or patch release. 

Another option would be to simply check if the propert source name contains the profile name. But i do not think this is a good idea, as there can be false positives.

marbon87: > 

extracted the condition to a private method and add documentation.

pom.xml
robertmcnees: Is there a reason to import testcontainers twice?  The pom pulls in 1.15.1 already.  Could we change `${testcontainers.version}` instead?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java
ryanjbaxter: Why can't we just look for the pattern `,profile_`?

marbon87: We can check for `,profile` (without `_`).
The code can look like this:

```
if(propertySourceName.contains(".*[-,]" + profile + ".*")){
...
}
```

pom.xml
marbon87: Oh sorry, missed that. I will update the property.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java
ryanjbaxter: Let's go with that as it's similar to the logic already there

ryanjbaxter (approved):
This looks good to me.  @spencergibb what do you think?

ryanjbaxter:
Can you make the PR agains the 3.0.x branch?

ryanjbaxter:
Merged in https://github.com/spring-cloud/spring-cloud-config/commit/b5ac67167135fd6301843446ccacb68194a2f8f9

Issue #1611, opened 2020-05-01
S3 backed AWS Config server does not read default property file


allanslim:
**Describe the bug**
 I created a Cloud Config Server that is backed by S3, but when I curl for the properties, it does not read the content of the default (application.yml) file. With the Git version, it reads the content of application.yml and {application-name}-{profiles}.yml one after the other. I looked at the code (AwsS3EnvironmentRepository) and I don't think it has functionality to read any files from S3 other than what than the {application-name}-{profiles}.yml file that it is being requested. Does anyone have a solution for this? The goal is to avoid common properties to be replicated per clients. Any help will be great appreciated. Thanks!

mattmonksdlg:
I am also facing the same issue.  We are trying to switch from using a Git-backed config service to an S3-backed service, and this is causing us issues.  It would be great if the `AwsS3EnvironmentRepository` could behave in the same way as the Git one.

Also - the `AwsS3EnvironmentRepository` does not throw any sort of exception when the bucket name is wrong, or cannot be accessed.  This is because it swallows all exceptions in the `getS3ConfigFile` method without logging anything and eventually just returns an empty environment.  This is not super helpful when trying to figure out environmental / permissions issues in AWS.  Could this also be improved?  Let me know if I should open a separate request.  Thanks!

divinedragon:
This is especially important for people who are migrating from Git based configurations to S3 based configurations and they are not compatible.

rodfrancisco:
I am also facing the same/similar issue where it behaves differently from `NativeEnvironmentRepository`. For `NativeEnvironmentRepository`, if a `{label}` is included as part of the client request the config server will return a union of configuration that is from the default properties (application.yml), service properties (service.yml) and the labelled version of those files ({label}/application.yml, {label}/service.yml). For `AwsS3EnvironmentRepository` it just returns the labelled version of the service properties ( {label}/service.yml ).

amathias:
I am also having this problem - is anyone working on this ?

avnerstr:
I also see this problem. Is someone working to solve this bug ?! it's totally wrong behaviour and it's different than the git behaviour

claraccio:
Yes we're also seeing this issue, would be great to have this fixed, as the behavior is different from GIT-based.
In the meantime we are running our app with both profiles `default,something` and having an `appName-default.yml` file in S3.

jaganm2018:
[...]

PR #2053, opened 2022-02-27
Fixes gh-1611 - AWS S3 environment repository can read the default property file as well as a profile-specific file

dmaiken:
When fetching properties from S3, always look for the default property file `[application-name].[yml|json|properties]` as well as `[application-name]-[profile].[yml|json|properties]`. Previously it was only doing this if we were supplying the property that was configured as the default on the Config Server (usually "default").

This aligns the AWS S3 environment repository behavior with that of the native implementation, git.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
@@ -18,7 +18,10 @@ package org.springframework.cloud.config.server.environment;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Properties;
+import java.util.stream.Collectors;
 
 import com.amazonaws.services.s3.AmazonS3;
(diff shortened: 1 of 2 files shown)


dmaiken:
Should this be merging into a development branch and not `main`?

spencergibb:
Main is our development branch

dmaiken:
Okay, double checking. I had seen PRs to branches like [spring-cloud:3.1.x](https://github.com/spring-cloud/spring-cloud-config/tree/3.1.x), and I wanted to confirm I should still me merging into `main`

spencergibb:
I see. Bug fixes still go into 3.1.x

ryanjbaxter (changes requested):
What I think we should do is add the `default` profile here
https://github.com/spring-cloud/spring-cloud-config/pull/2053/files#diff-e1e94601aac668a55b5e0fc3dce9925418ab34fb01ee47da33b38724a8192977L73

if `specifiedProfiles` or `defaultProfiles` does not already contain it.  That should pick up the configuration file without the profile name

ryanjbaxter:
Pulled changes into https://github.com/spring-cloud/spring-cloud-config/pull/2234

PR #2056, opened 2022-03-02
[config-server] Allow specification of down health status

antechrestos:
Allow developer to set the property `spring.cloud.config.server.health.down-health-status` to a custom value. This may be usefull when you make the clone critical (init container for instance), yet want your pod not to be down if your remote is unavailable or cannot be reached; the service will be restricted yet should be seen as alive by the platform.

For instance we may use `RESTRICTED` status that is used by spring cloud admin to display it in yellow...

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerHealthIndicator.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerHealthIndicator.java
@@ -30,6 +30,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.health.AbstractHealthIndicator;
 import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.Status;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.config.environment.Environment;
 import org.springframework.cloud.config.environment.PropertySource;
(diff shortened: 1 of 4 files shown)


antechrestos:
@spencergibb @ryanjbaxter this pull request is for discussion.

BTW @spencergibb , I may have spotted an issue in the documentation. [Health Indicator](https://github.com/spring-cloud/spring-cloud-config/blob/main/docs/src/main/asciidoc/spring-cloud-config.adoc#health-indicator) section indicates that to disable the health check, you need to set `management.health.config.enabled=false`. However, in the code, [another property is referenced](https://github.com/spring-cloud/spring-cloud-config/blob/863285d4939b85d38a0f231fcb8a7f7050c0f4da/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EnvironmentRepositoryConfiguration.java#L144)  - the one we find on [your stackoverflow answer](https://stackoverflow.com/questions/58122608/how-to-disable-configserver-health-page-checking-for-default-label-git) or in the [official documentation](https://cloud.spring.io/spring-cloud-config/reference/html/#_health_indicator). I can make another PR if needed.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/ConfigServerHealthIndicator.java
```diff
@@ -48,6 +48,8 @@
 
 	private Map<String, Repository> repositories = new LinkedHashMap<>();
 
+	private String downHealthStatus = "DOWN";
```
ryanjbaxter, line 51: I would set this to `Status.DOWN.getCode()`

ryanjbaxter (changes requested):
Can you add a test that actually uses the property?  I don't think it will actually work the way the code is now since there is no setter.

ryanjbaxter:
`management.health.config.enabled=false` should work base on this
 https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.health.auto-configured-health-indicators

ryanjbaxter (approved):

antechrestos:
@ryanjbaxter thanks

PR #2059, opened 2022-03-10
Allow for hyphen in path but exclude periods. 

### LINKED ISSUE #2058: Profiles with `-` in the name cause 404 from config server
**Describe the bug**

Spring Cloud Config Server version 3.1.1
Spring Cloud Config Client version 3.1.1

When a label is not provided the config client will make a request to the `/{name}/{profile}` endpoint. As of this change - https://github.com/spring-cloud/spring-cloud-config/pull/2031 - the EnvironmentController no longer supports profiles with a `-` present in their name and a 404 is returned.

**Sample**

Make a request to a running config server using a profile with a `-` in it. For example: `http://localhost:8888/example-service/cloud,pr-acceptance` will return a 404. If the offending profile is removed or the `-` is removed from the profile name it works as expected. The client will log the following warning: `Could not locate PropertySource: None of labels [] found`

ryanjbaxter:
Fixes #2058

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
@@ -104,12 +104,14 @@ public class EnvironmentController {
 		this.acceptEmpty = acceptEmpty;
 	}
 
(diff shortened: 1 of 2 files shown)


spencergibb:
Why exclude dashes or dots?

spencergibb (approved):

Issue #2064, opened 2022-03-17
ConfigServerConfigDataLoader constructor should use DeferredLogFactory  


ruansheng8:
**Describe the bug**

Spring cloud version : 4.0.0-SNAPSHOT

Spring cloud config client : ConfigServerConfigDataLoader debug level log is not working
ConfigServerConfigDataLoader constructor should use DeferredLogFactory  (see ConfigDataLoaders)

And i fixed it in #2063

**Sample**

My config client application.yml config is 
```
spring:
  config:
    import:
      - "configserver:http://localhost:8888"
logging:
  level:
    org.springframework.cloud.config.client.ConfigServerConfigDataLoader: debug     
```
if constructor use 
```
	public ConfigServerConfigDataLoader(Log logger) {
		this.logger = logger;
	}
```
Then debug level is not working.
It should be 
```
	public ConfigServerConfigDataLoader(DeferredLogFactory logFactory) {
		this.logger = logFactory.getLog(getClass());
	}
```
Reference : org.springframework.boot.context.config.ConfigDataLoaders

ryanjbaxter:
@ruansheng8 I saw you accidentally deleted the branch, can you open another PR, but this time submit it against the 3.1.x branch?

PR #2066, opened 2022-03-18
ConfigServerConfigDataLoader should use DeferredLogFactory

ruansheng8:
Fix debug level logs are not working in ConfigServerConfigDataLoader .
ConfigServerConfigDataLoader should use DeferredLogFactory.

Reference : org.springframework.boot.context.config.ConfigDataLoaders constructor

See issue: #2064


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLoader.java
```diff
@@ -68,8 +69,8 @@
 
 	protected final Log logger;
 
-	public ConfigServerConfigDataLoader(Log logger) {
-		this.logger = logger;
+	public ConfigServerConfigDataLoader(DeferredLogFactory logFactory) {
```
ryanjbaxter, line 72: Could you leave the current constructor and create a second constructor?

ryanjbaxter (changes requested):

PR #2081, opened 2022-04-25
First pass on observation

marcingrzejszczak:


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/ObservationEnvironmentRepositoryAspect.java
```diff
@@ -0,0 +1,65 @@
...
+		this.registry = tracer;
+	}
+
+	@Around("execution (* org.springframework.cloud.config.server.environment.EnvironmentRepository.*(..))")
+	public Object observationFindEnvironment(final ProceedingJoinPoint pjp) throws Throwable {
+		// @formatter:off
+		ObservationEnvironmentRepositoryContext context = new ObservationEnvironmentRepositoryContext(pjp);
+		Observation observation = ConfigDocumentedObservation.CONFIG_OBSERVATION.observation(this.registry, context)
+			.keyValuesProvider(this.keyValuesProvider).start();
+		try (Observation.Scope scope = observation.openScope()) {
```
jonatan-ivanov, line 48: You might do `return observation.observeChecked(() -> pjp.proceed());`. As far as I remember, it should do the same.

jonatan-ivanov (approved):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/VaultKvAccessStrategySupport.java
```diff
@@ -70,7 +71,7 @@ public String getData(HttpHeaders headers, String backend, String key) {
 
 			ResponseEntity<VaultResponse> response = this.rest.exchange(urlTemplate, HttpMethod.GET,
 					new HttpEntity<>(headers), VaultResponse.class, key);
-			HttpStatus status = response.getStatusCode();
+			HttpStatusCode status = response.getStatusCode();
```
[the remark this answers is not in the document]
marcingrzejszczak: The code didn't compile - that's why I fixed it. Sure, I'll look into other lines

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/ObservationEnvironmentRepositoryAspect.java
marcingrzejszczak: Ah yeah, that's because `pjp.proceed` throws a `Throwable` not an `Exception`

spring-cloud-config-server/pom.xml
```diff
@@ -42,6 +42,15 @@
...
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-observation</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.aspectj</groupId>
+			<artifactId>aspectjrt</artifactId>
+			<optional>true</optional>
```
spencergibb, line 52: This causes all the tests in the samples module to fail with org.aspectj.lang.ProceedingJoinPoint

marcingrzejszczak: We could but we would need to apply changes in A LOT of places (but it could be done). I'll take another look at this

marcingrzejszczak:
@spencergibb this commit (https://github.com/spring-cloud/spring-cloud-config/pull/2081/commits/03ad19a06acb72921c76fdc5e979f179019b0793) migrates away from AOP

marcingrzejszczak:
We should do more tests with composites to ensure that we're not double wrapping anything

Issue #2083, opened 2022-05-02
Cloud Config Server returns 404 when profile name contains more than one dash "-"


bvtk:
Spring Cloud Config supported dashes "-" in profile names when using spring-cloud-dependencies: 2020.0.3. 

We wanted to migrate to Version: 2021.0.1, but there was a problem, that you'll always recive a 404 (NOT FOUND), wenn die profile name contains a dash "-". Reason for this Error is probably this change:

Commit: https://github.com/spring-cloud/spring-cloud-config/commit/facb18da95e81f7058a63b408c180fe333a5c29d 
Class: https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java

We saw, that in the newest version, there will be a fix for that problem in Version 2021.0.2 so we waited for it. The good news it, that now the EnvironmentController does support dashes in Profile Names. But now, when the profile name contains json, you recieve a 404 (NOT FOUND). For example: enable-json-conversion, enable-json

ryanjbaxter:
So you are saying that if the profile name contains `-json` it returns a 404?

bvtk:
Yes first i thought, more than one dash is the problem, but it is, when the profile name contains json. For example enable-json-conversion

PR #2089, opened 2022-05-11
Update EnvironmentController @GetMapping - Fixes gh-2083

VincentBostoen:
As stated in #2083  when passing profile that contains a keyword in ```[yml, yaml, json, properties]```, ```EnvironmentController``` returns a 404 HttpStatus.  

I quite struggle to understand the evolution of the controller getMapping filtering but I suppose that there was a will of refusing profiles having an extension in ```[.yml, .yaml, .json, .properties]```.

Here is a proposal that goes in this direction.


ryanjbaxter:
Can you change the base back to 3.1.x?

VincentBostoen:
> Can you change the base back to 3.1.x?

Done, let me know if there is anything else.

Issue #1901, opened 2021-06-03
Add support for Apache MINA


spencergibb:
see https://github.com/spring-cloud/spring-cloud-config/issues/1881#issuecomment-843231353

SivaTharun:
@spencergibb can i work on this issue, i see that the above mentioned issues refer to replacing jsch library for apache MINA.

i see that jsch is being used for ssh session/file operations in the below files and their corresponding unit test classes.

- PropertyBasedSshSessionFactory.java
- PrivateKeyValidator.java
- PropertiesBasedSshTransportConfigCallback.java
- FileBasedSshTransportConfigCallback.java
- JGitEnvironmentRepository.java

PR #2108, opened 2022-06-15
Replaces abandoned JSch library with the forked one (com.github.mwiede.jsch) which is actively maintained. 

kvmw:
There are couple of issues (e.g. #2061) regarding the lack of support for new ssh-key algorithms by `JSch` (used by `JGit`). 

As my understanding the long-term plan is to replace `JSch` with `Apache MINA` (#1901). But in the short and mean-term term this PR can solve the issue by replacing the `JSch` library with a [forked one](https://github.com/mwiede/jsch) which promises a drop in replacement and 100% compatibility with the original `JSch` library. 

The PR enables following features:
- support for new ssh-key algorithms like `ed25519`
- support for new host-key algorithms like `ssh-ed25519`


kvmw:
This can be closed in favour of : https://github.com/spring-cloud/spring-cloud-config/pull/2113

PR #2113, opened 2022-06-21
Replaces Jsch with Apache MINA in config-server

kvmw:
-  Bumps jgit to 6.2.x
- FileBasedSshSessionFacotry sets the properties only for relevant host.
    In the existing `FileBasedSshSessionFactory` the `StrictHostKeyChecking` flag is decided based on top level 
   `strictHostKeyChecking` property under `git`.  This has changed in this pull-request  and the decision is made based on 
   `strictHostKeyChecking` for the specific hostname.  The main reason for this change was to make it behave like 
 `PropertyBasedSshSessionFactory`.  If this is not the desired behaviour, it can be reverted.

### DIFF
--- a/spring-cloud-config-dependencies/pom.xml
+++ b/spring-cloud-config-dependencies/pom.xml
@@ -15,7 +15,7 @@
 	<name>spring-cloud-config-dependencies</name>
 	<description>Spring Cloud Config Dependencies</description>
 	<properties>
-		<jgit.version>5.13.1.202206130422-r</jgit.version>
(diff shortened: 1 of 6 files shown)


spencergibb:
From a conversation with @ryanjbaxter, looks like this will be reworked for compatibility with the 3.1.x branch. Then a follow-up could change defaults and major versions in main.

ryanjbaxter:
@kvmw I merged https://github.com/spring-cloud/spring-cloud-config/pull/2122 into main as well.  We can update this PR to use the latest JGit changes and make other breaking changes

Issue #2123, opened 2022-07-18
Server doesn't handle redirected git repository anymore


ts-schu:
**Describe the bug**
spring-cloud-config v3.1.3:
spring-cloud-config-server configured to use a git repository throws TransportException with error code 302 Found when that repository gets redirected (e.g. moved to a diferent group/organization within the git server).

**Sample**
Setup the server similar to README.md (spring-cloud-config-server/src/main/resources/configserver.yml modified to contain:
```
info:
  component: Config Server

spring:
  application:
    name: configserver
  autoconfigure.exclude: org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration
  jmx:
    default_domain: cloud.config.server
  cloud:
    config:
      server:
        git:
          uri: http://gitea.minikube.local/test/repo.git
          username: user
          password: password

server:
  port: 8888
management:
  endpoints:
    web:
      base-path: /admin
```
) access to `README.md` in root of repo works fine (`curl localhost:8888/repo/README.md` -> 
```
$ curl localhost:8888/repo/README.md
{"name":"repo","profiles":["README.md"],"label":null,"version":"a2aebed6dd3ba26272feb2fb61c8b6965aad984a","state":null,"propertySources":[]}
```
).

Then the group/organization gets renamed (`test` to `test-redirected`) and the config-server gets restarted to eliminate caching:
the server now throws multiple exceptions
```
2022-07-18 12:15:18.081  WARN 13248 --- [nio-8888-exec-1] .c.s.e.MultipleJGitEnvironmentRepository : Error occured cloning to base directory.

org.eclipse.jgit.api.errors.TransportException: http://gitea.minikube.local/test/repo.git: 302 Found
        at org.eclipse.jgit.api.FetchCommand.call(FetchCommand.java:224) ~[org.eclipse.jgit-5.12.0.202106070339-r.jar:5.12.0.202106070339-r]
        at org.eclipse.jgit.api.CloneCommand.fetch(CloneCommand.java:303) ~[org.eclipse.jgit-5.12.0.202106070339-r.jar:5.12.0.202106070339-r]
        at org.eclipse.jgit.api.CloneCommand.call(CloneCommand.java:178) ~[org.eclipse.jgit-5.12.0.202106070339-r.jar:5.12.0.202106070339-r]
[...]

PR #2128, opened 2022-08-03
Fix access to redirected git repositories

ts-schu:
spring-cloud-config-server configured to use a git repository throws `TransportException with error code 302 Found` when that repository gets redirected (e.g. moved to a diferent group/organization).

Fixes [issue 2123](https://github.com/spring-cloud/spring-cloud-config/issues/2123)

Version : v3.1.3
Module : Spring-Cloud-Config-Server

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
@@ -70,7 +70,13 @@ public class HttpClientConfigurableHttpConnectionFactory implements Configurable
 	@Override
 	public HttpConnection create(URL url, Proxy proxy) throws IOException {
 		HttpClientBuilder builder = lookupHttpClientBuilder(url);
-		return new HttpClientConnection(url.toString(), null, builder != null ? builder.build() : null);
+		if (builder != null) {
+			return new HttpClientConnection(url.toString(), null, builder.build());
+		}
+		else {
+			/* No matching builder found: let jGit handle the creation of the HttpClient */
+			return new HttpClientConnection(url.toString());
+		}
 	}
 
 	private void addHttpClient(JGitEnvironmentProperties properties) throws GeneralSecurityException {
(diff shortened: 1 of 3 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
```diff
@@ -99,7 +99,7 @@ private HttpClientBuilder lookupHttpClientBuilder(final URL url) {
 
 		if (builderMap.isEmpty()) {
 			this.log.warn(String.format("No custom http config found for URL: %s", url));
-			return HttpClients.custom();
```
ryanjbaxter, line 102: Why this change?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/HttpClientSupport.java
```diff
@@ -65,6 +65,8 @@ public static HttpClientBuilder builder(HttpEnvironmentRepositoryProperties envi
 			httpClientBuilder.setDefaultCredentialsProvider(new SystemDefaultCredentialsProvider());
 		}
```
ryanjbaxter, line 67: A comment here indicating JGit is handling redirects would be good

ryanjbaxter (changes requested):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/HttpClientConfigurableHttpConnectionFactory.java
ts-schu: I changed the code to give jGit the oppertinity to use their own HttpClient for standard cases. It is not necessary but prevents incompatible configuration of HttpClient between jGit and SpringCloudConfig

ts-schu: I can drop that change (and the related changes to the unit test).

ts-schu: Good idea, but

- using HttpClientSupport needs either JGitEnvironmentProperties or MultipleJGitEnvironmentProperties and
- has to be handling GeneralSecurityException

IMO the effort is too much in comparison to letting jGit handle the creation of the HttpClient

ryanjbaxter: OK please ad a comment that we are letting JGit handle it

Issue #2130, opened 2022-08-08
Can't get the right config file on server by certain url pattern


woshikid:
I'm using org.springframework.cloud:spring-cloud-config-server:3.1.0

According to the doc https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#_quick_start
The HTTP service has resources in the following form:
```
/{application}/{profile}[/{label}]
/{application}-{profile}.yml
/{label}/{application}-{profile}.yml
/{application}-{profile}.properties
/{label}/{application}-{profile}.properties
```
For example:
```
curl localhost:8888/foo/development
curl localhost:8888/foo/development/master
curl localhost:8888/foo/development,db/master
curl localhost:8888/foo-development.yml
curl localhost:8888/foo-db.properties
curl localhost:8888/master/foo-db.properties
```

But when I use `curl localhost:8888/master/foo-db.properties` I got:
```
{"name":"master","profiles":["foo-db.properties"],"label":null,"version":"52ba55eecc7239f2a24f529eccf640befc34a6d9","state":null,"propertySources":[]}
```
That means the service take "master" as {application} instead of {label} and take "foo-db.properties" as {profile} not {application}-{profile}.properties

It seems that pattern `/{label}/{application}-{profile}.properties` have be overridden by pattern `/{application}/{profile}[/{label}]`

woshikid:
In spring-cloud-config-server:3.1.0 org.springframework.cloud.config.server.environment.EnvironmentController has `@GetMapping(path = "/{name}/{profiles:.*[^-].*}", produces = MediaType.APPLICATION_JSON_VALUE)` I think that's the regex caused the problem. I guess it want to exclude the pattern `{application}-{profile}.properties` but `.*[^-].*` still matches `"foo-db.properties"`. I think `[^-]*` or `[^\\.]*` may solve the problem.

woshikid:
It is fixed by later release, just need to simplify

PR #2132, opened 2022-08-08
Simplify regex in EnvironmentController

woshikid:
As described in #2130 , regex `.*[^-].*` still matches `"foo-db.properties"`. So we can use `[^-]*` or `[^\\.]*` to solve the problem.
According to #2083 , we should support dashes "-" in profile names. So we can use `[^\\.]*`.

I checked the history of this regex:
1. in spring-cloud-config-server:3.1.0 the regex is `.*[^-].*` but caused #2130
2. in spring-cloud-config-server:3.1.1 the regex is `[^-]+` but caused #2083
3. in spring-cloud-config-server:3.1.2 the regex is `(?!.*\\b(?:ya?ml|properties|json)\\b).*`. That's working, but maybe we should make it simpler and easier to understand.

So I suggest to use `[^\\.]*` to exclude profiles with file extensions.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
@@ -104,13 +104,13 @@ public class EnvironmentController {
 		this.acceptEmpty = acceptEmpty;
 	}
 
-	@GetMapping(path = "/{name}/{profiles:(?!.*\\b\\.(?:ya?ml|properties|json)\\b).*}",
+	@GetMapping(path = "/{name}/{profiles:[^\\.]*}",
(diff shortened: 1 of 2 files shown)


woshikid:
I found the existing test already covered the case that profile with file extensions, so I didn't add new one.
```java
@ParameterizedTest
@ValueSource(strings = { "yml", "yaml", "json", "properties" })
public void profileHavingAnExtension(String extensionKeyword) throws Exception {
    String profiles = "dev." + extensionKeyword;
    Environment dashEnvironment = new Environment("foo", profiles);
    dashEnvironment.add(new PropertySource("foo", new HashMap<>()));
    this.mvc.perform(MockMvcRequestBuilders.get("/foo/" + profiles))
            .andExpect(MockMvcResultMatchers.status().isNotFound());
    verifyNoInteractions(this.repository);
}
```

PR #2144, opened 2022-09-09
fix micrometer dependency changes

wind57:


wind57:
@ryanjbaxter this build fails because spring-boot has to fix it also... the issue is that `io.micrometer:micrometer-observation:1.10.0-SNAPSHOT` changed some location of the classes...

ryanjbaxter:
Yes the Boot team is actively working through it

ryanjbaxter (approved):

Issue #2133, opened 2022-08-08
Tracking Git branch instead of tag when branch ends with tag


ojecborec:
The `JGitEnvironmentRepository` has the method called `containsBranch` that is checking whether `label` is branch or not (line 743). Imaging we have

label = foo
branch = feature/foo
tag = foo

Given that it returns `true` because ref name ends with `/foo`.

Expected: Returns `false`.

ryanjbaxter:
See the section on labels here
https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#_git_backend

ojecborec:
I'm very well aware of following statement from documentation

> If the git branch or tag name contains a slash (/), then the label in the HTTP URL should instead be specified with the special string (_) (to avoid ambiguity with other URL paths).

But this is different. It has nothing to do with the HTTP URL.

ryanjbaxter:
I think the problem here is the ambiguity of label.  If you have a branch and a tag with the same name how do we know which one we should be using?  This problem is a subset of that problem where the branch name ends with the tag.  I am not sure what the best way to solve the problem is.  The only thing I can come up with is adding something to the label that identifies whether you are specifically talking about a branch or a tag.  Or we might be able to not use endsWith and force the label to be the complete branch name when specifying a label.

This code has been this way for 8 years and this is the first time we have seen this issue, so while its an issue it appears to be rare.

ojecborec:
> If you have a branch and a tag with the same name how do we know which one we should be using?

In this case yes. We could say that branch has higher priority and request would not fail. However this is not the case. Branch name is different to tag one. The problem is using `endsWith` method and `/` as a separator. One possible solution is to normalize ref name before performing full match, i.e.

```
refs/heads/main -> main
refs/remotes/origin/feature/foo -> feature/foo
```

Or maybe the `containsBranch` method can be replaced by calling `@Bean` annotated component allowing people to overwrite default implementation.

> This code has been this way for 8 years and this is the first time we have seen this issue, so while its an issue it appears to be rare.

Maybe people got used to it and found workarounds. This does not mean that Config Server should not support using `/` character to create a hierarchical name scheme.

ryanjbaxter:
>One possible solution is to normalize ref name before performing full match

[...]

PR #2147, opened 2022-09-15
Exact match of branch names in JGitEnvironmentRepository - Fixes gh-2133

woshikid:
As @ojecborec described in #2133 
> 1. Given
> branch=feature/foo
> tag=foo
>
> 2. Send following request to Config Server
> `/application/default/foo`
>
> 3. `JGitEnvironmentRepository.containsBranch` returns true because `refs/remotes/origin/feature/foo` ends with `/foo`
>
> 4. `checkout.call()`, line 468 throws `RefNotFoundException`. Ref origin/foo cannot be resolved. Which is true as this branch does not exist.

The method `containsBranch` in class `JGitEnvironmentRepository` do have bug dealing with branch name contains slashes (/).

I think it's ok to change the `endsWith` method into `equals` when matching branch names. And this fix the problem.


ryanjbaxter:
Can you add a test that covers the breaking case?

woshikid:
OK, I have added a test that checkout tag `foo` when exists branch `feature/foo`

ryanjbaxter (changes requested):
looks good, can you remove the hooks and info directory?

woshikid:
OK the git directory has been cleaned.

Issue #2161, opened 2022-09-24
suggestion on Exception Handling in MultipleJGitEnvironmentRepository


ndy2:
**Is your feature request related to a problem? Please describe.**
Hi! I'm newbie on spring config and frustrated by the 404 error on my config server
![image](https://user-images.githubusercontent.com/67302707/192106854-15ad1a42-4dd3-4bdd-bce3-b6b86a14e738.png)

the error log is like no such label master

I google the error and the answer is like add default-label @ application.properties
```properties
spring.cloud.config.server.git.default-label= main
```

but it did not solve the problem. 
Now I Found That default label is now main @ defulat 
and the tryMaster feature was added due to maintain compatibility
 [related commit](https://github.com/spring-cloud/spring-cloud-config/commit/38a09020e56940942680f48f34a0d2a7c94eb5b3)

---
above was my story of how i get to this suggestion.

I hope the method `MultipleJGitEnvironmentRepository.findOne` provide a little more kind log or exception handling
in blow lines

line 175
```java
catch (Exception e) {
	if (MultipleJGitEnvironmentProperties.MAIN_LABEL.equals(label) && isTryMasterBranch()) {
		candidate = getRepository(this, application, profile, MultipleJGitEnvironmentProperties.MASTER_LABEL);
		return findOneFromCandidate(candidate, application, profile,
				MultipleJGitEnvironmentProperties.MASTER_LABEL, includeOrigin);
	}
	throw e;
}
```

if the error is catched and then it try again on master branch

if another error is throw while try on master branch.
it just eat the original exception which is thrown by first try (main branch)  !!!

**Describe the solution you'd like**
I want to add simple log for original exception.

**Additional context**
the log i expected 
![image](https://user-images.githubusercontent.com/67302707/192107916-eb256d82-4682-48d3-a9bf-42f6d0d6a1ef.png)

i solve my problem and got to know actual exception by set
```properties
spring.cloud.config.server.git.try-master-branch: false
```

the log i got
![image](https://user-images.githubusercontent.com/67302707/192107850-53eb7ee0-4aed-4ba2-9953-534660b84d8b.png)

woshikid:
What about a suppressed exception?
```java
catch (Exception e) {
    if (MultipleJGitEnvironmentProperties.MAIN_LABEL.equals(label) && isTryMasterBranch()) {
        candidate = getRepository(this, application, profile, MultipleJGitEnvironmentProperties.MASTER_LABEL);
[...]

PR #2162, opened 2022-09-26
Update logging message on MultipleJGitEnvironmentRepository, JGitEnvironmentRepository

ndy2:
As decribe in #2161, it is more easy to debug by throw original exception.
It can be handled by adding original exception as suppressed to master branch exception in case both findOne try thorws exception.

Thanks for @woshikid to sharing solutions.


woshikid:
in `JGitEnvironmentRepository#getLocations` there is another Try Master Branch try catch. maybe we should do the same thing here.

ndy2:
I will check and add !

woshikid:
I debugged and found the actually Try Master Branch behavior happens in `JGitEnvironmentRepository#getLocations` in normal situation.

ndy2:
I found that there are actually three case where
`tryMasterBranch` flag in `JGitEnvironmentRepository` is used as flag for retry some logic

1.` JGitEnvironmentRepository#getLocations`
2. `JGitEnvironmentRepository#checkoutDefaultBranchWithRetry`
3.` MultipleJGitEnvironmentRepository#findOne`

two methods in JGitEnvironmentRepository handle original exception by logging.

Considering that, i think simple loggin on `MultipleJGitEnvironmentRepository#findOne` would be consistent solution for this issue.

So I add a log on `MultipleJGitEnvironmentRepository#findOne` refer to `JGitEnvironmentRepository`.

And fix a wrong log message in `JGitEnvironmentRepository`

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/MultipleJGitEnvironmentRepository.java
```diff
@@ -160,7 +160,6 @@ public Environment findOne(String application, String profile, String label, boo
 							this.logger.debug("Cannot load configuration from " + candidate.getUri() + ", cause: ("
 									+ e.getClass().getSimpleName() + ") " + e.getMessage(), e);
 						}
-						continue;
```
ryanjbaxter, line 163: Why is this being removed?

ryanjbaxter (changes requested):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/MultipleJGitEnvironmentRepository.java
ndy2: My IDE suggests
```
'continue' is unnecessary as the last statement in a loop 
```

so i removed it.
I find that there is one more such continue statement at `MultipleJGitEnvironmentRepository.getLocations`

May i remove 'continue;' from that also or retreive this one.

spencergibb: please limit changes to those specific to the issue.

ndy2:
After I restore the continue statement. 
BootstrapConfigServerIntegrationTests.contextLoads fails...

I don't know why it fails. Since the change i made is only to add some logging.

ryanjbaxter:
@ndy2 can you please pull the latest changes from 3.1.x into your branch?

PR #2166, opened 2022-09-28
Fixes NPE in HttpProxyDataFactory when is called with unknown host.

kvmw:
`PropertyBasedSshSessionFactory.HttpProxyDataFactory` throws NPE if it's called with an unknown host. 

It should handle those kind of hosts gracefully.


spencergibb:
I don't think an unknown host should fail gracefully. Can you explain why you think it should?

spencergibb:
Closing in favor of #2168

PR #2167, opened 2022-09-28
Supports private-keys with passphrase for config-server

kvmw:
Enables config-server to accept the private keys with passphrase


ryanjbaxter:
LGTM, should this also be in the 3.1.x branch?

kvmw:
@ryanjbaxter  yes. i was waiting for the reporter to confirm the case, before linking it to the issue. 

regarding 3.1.x branch:  it's nice to have it there too but SCS  in particular does not support the keys with passphrase.

PR #2168, opened 2022-09-28
Use both hostname and IP address to find ssh configuration in config-server

kvmw:


spencergibb:
I don't think an unknown host should fail gracefully. Can you explain why you think it should?

Closed #2166 as a duplicate, why open multiple PRs?

kvmw:
@spencergibb @ryanjbaxter 

After checking the issue in details with local builds for SCS tile I  found the root cause of the NPE. 

Basically if git configuration is done using `IP address` instead of `hostname`, config-server won't be able to find the git configs.  I've changed the code to look for both hostname and IP address for ssh calls. 

This new private method `findEnvironmentProperties` is the one that does the job. 

btw, I've used `127.0.0.1` and `localhost` in unit tests to verify the fix. but i'm not sure if in the pipeline or other developers machine `localhost` resolves to `127.0.0.1`. do you have any suggestion?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/PropertyBasedSshSessionFactory.java
```diff
@@ -193,6 +179,19 @@ protected Iterable<KeyPair> getDefaultKeys(File dir) {
...
+			Map<String, JGitEnvironmentProperties> sshKeysByHostname, InetSocketAddress socketAddress) {
+		if (sshKeysByHostname.containsKey(socketAddress.getHostString())) {
+			return sshKeysByHostname.get(socketAddress.getHostString());
+		}
+
+		if (socketAddress.getAddress() != null) {
+			return sshKeysByHostname.get(socketAddress.getAddress().getHostAddress());
+		}
+
+		return null;
```
ryanjbaxter, line 192: Should we return an empty map to avoid future NPEs?

kvmw: the return value is `JGitEnvironmentProperties`, not a map.  If we return a default instance of `JGitEnvironmentProperties` would cause other issues. the `null`  return value implies that there has been no configuration at all for this host/ip but a default `JGitEnvironmentProperties` does not convey the same message. 

I could use Optional here but since the method is private, I thought the use of Optional is overkill.

ryanjbaxter:
@kvmw i thought there was an issue opened that this PR resolved but I don't see it, am I just missing it, or am I imagining that and there wasn't a corresponding issue open?

spencergibb:
> I don't think an unknown host should fail gracefully.

Covering up a NPE isn't satisfactory to me. There should be an error that you couldn't reach the proxy shouldn't it?

kvmw:
> @kvmw i thought there was an issue opened that this PR resolved but I don't see it, am I just missing it, or am I imagining that and there wasn't a corresponding issue open?

@ryanjbaxter 
There is no issue reported for the PR. I think I was the first one faced this issue because SCS mirror service is using IP address instead of hostname for the configuration.  I can create one if is needed.

spencergibb:
No need for a new issue

kvmw:
> Maybe the happy middle ground here is if there isn't any proxy information in the app configuration then we log a warning saying that no information is found and they should check their configuration. I think the issue is we are concerned that things just continue as normal without any idea that something may be wrong.

@ryanjbaxter After checking the old commits (before Apache MINA migration) and also trying few scenarios, I came to conclusion that those methods won't be called with arbitrary hostnames or IP addresses.  So I removed the unnecessary null checks and also updated the PR title to avoid confusion.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/PropertyBasedSshSessionFactory.java
```diff
@@ -193,6 +173,18 @@ protected Iterable<KeyPair> getDefaultKeys(File dir) {
 		return new SingleKeyIdentityProvider(sshKeysByHostname);
 	}
 
+	private static JGitEnvironmentProperties findEnvironmentProperties(
+			Map<String, JGitEnvironmentProperties> sshKeysByHostname, InetSocketAddress socketAddress) {
+
+		JGitEnvironmentProperties sshProperties = sshKeysByHostname.get(socketAddress.getHostString());
+
+		if (sshProperties == null && socketAddress.getAddress() != null) {
+			sshProperties = sshKeysByHostname.get(socketAddress.getAddress().getHostAddress());
```
ryanjbaxter, line 182: What if this doesn't exist, won't this be null?

kvmw: For every given address,  the map (`sshKeysByHostname`) has either the hostname or the IP address.

The methods in the above classes are called by `JGit` when it wants to establish a connection to the hosts defined in `spring.cloud.config.server.git.uri` or `spring.cloud.config.server.git.repos[].uri`.   Any host in those `URI`s will be added to the Map. So I don't see any scenario where  above method returns null.

ryanjbaxter:
@kvmw when I merged this forward in the main branch this test is failing
https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/ssh/PropertyBasedSshSessionFactoryTest.java#L251

Looks like its because there isn't any ssh configuration for the host.  This test doesn't exist in the 3.1.x branch.  Maybe there is a null check missing?

kvmw:
@ryanjbaxter  `main` branch is slightly different from the `3.1.x`.  When Apache MINA was introduced there were some breaking changes which we kept in `main` but not in `3.1.x`.  

Do you want me to create a PR for the main branch too?

ryanjbaxter:
Sure although I might just take that and modify the merge to keep things simple as we are merging t 3.1.x into main.

kvmw:
> Sure although I might just take that and modify the merge to keep things simple as we are merging t 3.1.x into main.

In that case you may update that test like following to avoid NPE:
```java
	@Test
	public void defaultSshConfigIsSet() {
		JGitEnvironmentProperties sshProperties = new JGitEnvironmentProperties();
		sshProperties.setUri("ssh://gitlab.example.local:3322/somerepo.git");
		setupSessionFactory(sshProperties);

		SshConfigStore.HostConfig sshConfig = getDefaultSshHostConfig("gitlab.example.local", 123, "user.name");

		assertThat(sshConfig.getValue("HostName")).isEqualTo("gitlab.example.local");
		assertThat(sshConfig.getValue("Port")).isEqualTo("123");
		assertThat(sshConfig.getValue("User")).isEqualTo("user.name");
	}
```

The  `JGitEnvironmentProperties` needs at lest the URI.

ryanjbaxter:
Thanks that appears to get things working.  If you want to look at the latest commits from me on main just to make sure I didn't screw anything up that would be appreciated.

kvmw:
> Thanks that appears to get things working. If you want to look at the latest commits from me on main just to make sure I didn't screw anything up that would be appreciated.

@ryanjbaxter  there is only one unnecessary null check `HttpProxyDataFactory.get` which we had discussed above.  

Also another minor change was needed in `SshConfigStore.lookupDefault` to  set the default connection attempts to 1.  

I've created #2175 for above changes. feel free to apply the changes directly and close the PR, if it's easier.

PR #2171, opened 2022-09-30
Align Jdbc Repository's behavior with Git Repository

### LINKED ISSUE #2163: Jdbc Repository behaves differently from Git Repository
## issue with profile
When we request `/foo/dev` to config server
the Git Repository get config from following files:
* foo-dev.yml
* application-dev.yml
* foo.yml
* application.yml

the Jdbc Repository get config from following lines:
application | profile | label | key | value
-- | -- | -- | -- | --
foo | dev | master | ... | ...
foo | default | master | ... | ...
application | dev | master | ... | ...
application | default | master | ... | ...

When we request `/foo/default` to config server
the Git Repository get config from following files:
* foo-default.yml
* application-default.yml
* foo.yml
* application.yml

the Jdbc Repository get config from following lines:
application | profile | label | key | value
-- | -- | -- | -- | --
foo | default | master | ... | ...
application | default | master | ... | ...

**they treat profile `default` differently**
and Jdbc Repository don't support following config:
application | profile | label | key | value
-- | -- | -- | -- | --
foo | null | master | ... | ...
application | null | master | ... | ...

I thought they were the same as `foo.yml` and `application.yml` but they weren't.

## issue with label
the Git Repository first try label `main` then `master`.
the Jdbc Repository only try label `master`

Are they designed to be like this? If not, how should we fix this?

woshikid:
Close #2163 
The null value is so special in sql, so I had to add another sql to JdbcEnvironmentProperties


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JdbcEnvironmentRepository.java
```diff
@@ -128,6 +128,23 @@ public Environment findOne(String application, String profile, String label) {
 					}
 				}
 			}
+			// get properties without profile, equivalent to foo.yml, application.yml
```
ryanjbaxter, line 131: This code looks the same as the block above, can we extract this into a method and just call that method with the sql statement?

ryanjbaxter (changes requested):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JdbcEnvironmentRepository.java
woshikid: Done. 👌

PR #2176, opened 2022-10-08
Add default-label property to Jdbc Repository

### LINKED ISSUE #2165: Jdbc Repository don't try label main
the Git Repository first try label `main` then `master`.
the Jdbc Repository only try label `master`.

woshikid:
Close #2165 
add `defaultLabel` property to `JdbcEnvironmentProperties` as same as `NativeEnvironmentProperties`

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -891,6 +891,8 @@ The database needs to have a table called `PROPERTIES` with columns called `APPL
 All fields are of type String in Java, so you can make them `VARCHAR` of whatever length you need.
 Property values behave in the same way as they would if they came from Spring Boot properties files named `{application}-{profile}.properties`, including all the encryption and decryption, which will be applied as post-processing steps (that is, not in the repository implementation directly).
 
+NOTE: The default label used for JDBC is `master`.  You can change that by setting `spring.cloud.config.server.jdbc.defaultLabel`.
+
 ==== Redis Backend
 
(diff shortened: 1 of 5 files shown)


ryanjbaxter (changes requested):
Please add tests and documentation

woshikid:
sorry for the delay

ryanjbaxter (commented):
The default label should be `main` as it is with Git.  If there is no default label `main` then we should fall back and try `master` so we don't break anyone.

woshikid:
Yes, that's exactly what I wanted to do in the first place. And you also mentioned
> We try main then master because it was a default for Git repos. That doesn't mean much in terms of JDBC.

this sounds very reasonable, and I checked the design of `File System Backend`:
```java
private String defaultLabel = "master";
```
it uses `master` as default label and also don't have `tryMasterBranch` feature. So... I think maybe we should make JDBC Backend act as same as File System Backend. And this will not break anyone because the `lable` was hardcoded to `master` before.

woshikid:
I mean, the "try main then master feature" should exclusive to Git Backend

ryanjbaxter:
Maybe I was thinking that we could make that breaking change in our 2022 release, however since the release has already been released with `master` and in addition this PR is for 2021 we cannot break functionality, so now we have to try main and then fallback to master.

woshikid:
This PR will not break anything, the label used to be hardcoded to `master` and I just added a new variable `defaultLabel` with value `master`. No logic has been changed and everything will still run smoothly.

PR #2209, opened 2023-01-18
Support for ssh key with passphrase using Apache MINA

### LINKED ISSUE #2201: Passphrase is not used during ssh connection establishment
**Describe the bug**
SpringBoot version 3.0.0 
SpringCloud version 2022.0.0

According to docs https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#_authentication
I generated keypair adding keyphrase.
Then I specified this passphrase into application properties on branch `spring.cloud.config.server.git.passphrase`.
And this phrase was ignored during connection establishment to git server located at url like template `ssh://git@server:port/path/to/repo`

**Sample**
1. generate keypair with passphrase not empty using
`ssh-keygen -m PEM -t rsa -b 4096 -f ~/config_server_deploy_key.rsa`
2. add to ~/.ssh/config these lines
```
Host gitea.my.server
    HostName gitea.my.server
    IdentityFile ~/.ssh/config_server_deploy_key.rsa
    Port 2221
```
3. create initial config-server application with such properties
```
spring:
  cloud:
    config:
      server:
        git:
          uri: ssh://git@gitea.my.server:2221/path/to/repo.git
          clone-on-start: true
          ignore-local-ssh-settings: false
```
4. try to run simple test
```
@SpringBootTest
class ConfigServerApplicationTests {

    @Test
    void contextLoads() {
    }

}
```
5. these steps leads to login exception

---
My investigation of this issue.

Exception is thrown from `PassphraseCredentialsProvider.java`: 
```
  @Override
  public boolean get(URIish uri, CredentialItem... items) throws UnsupportedCredentialItem {
    for (final CredentialItem item : items) {
      if (item instanceof CredentialItem.StringType && item.getPromptText().startsWith(PROMPT)) {
        ((CredentialItem.StringType) item).setValue(this.passphrase);
        continue;
      }
      throw new UnsupportedCredentialItem(uri, item.getClass().getName() + ":" + item.getPromptText()); // HERE
    }
    return true;
  }
```

params in my case are:
1. constant PROMPT = Passphrase for
2. phrase from ssh: Key '/home/bvn13/.ssh/config_server_deploy_key.rsa' is encrypted. Enter the passphrase to decrypt it.
3. item instanceof CredentialItem.StringType is FALSE
4. item.getPromptText().startsWith(PROMPT) is FALSE

It leads me to idea that there is no CredentialProvider applicable to this kind of authorization (key + phrase).

So my questions are:
1. is this a bug?
2. what is `spring.cloud.config.server.git.passphrase` intended for?
3. how to apply custom CredentialProvider for such cases?

UPDATE:
Connection is established OK after I regenerate keypair lefting passphrase empty and refresh my key into Gitea server.

ryanjbaxter:
Fixes #2201

There are a couple issues with SSH keys using passphrases

* The `CredentialItem` type is now `Password` and not `StringType`
* The JSCH library would only here have 1 `CredentialItem`.  With Apache MINA there are now two, and the old code would throw an exception (for some reason) if the first one didn't match the type.  We now need to look through all of them to find the password prompt so I removed the code that would throw an exception.
* The prompt text is now different

@kvmw would appreciate it if you could also take a look at this

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/PassphraseCredentialsProvider.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/PassphraseCredentialsProvider.java
@@ -31,9 +31,9 @@ public class PassphraseCredentialsProvider extends CredentialsProvider {
 	/**
 	 * Prompt to skip iteration for.
 	 */
-	public static final String PROMPT = "Passphrase for";
+	public static final String PROMPT = "Passphrase";
 
-	private final String passphrase;
+	private final char[] passphrase;
 
 	/**
 	 * Initialize the provider with a the ssh passphrase.
@@ -41,7 +41,7 @@ public class PassphraseCredentialsProvider extends CredentialsProvider {
 	 */
 	public PassphraseCredentialsProvider(String passphrase) {
 		super();
-		this.passphrase = passphrase;
+		this.passphrase = passphrase.toCharArray();
 	}
 
 	/**
@@ -58,12 +58,13 @@ public class PassphraseCredentialsProvider extends CredentialsProvider {
 	@Override
 	public boolean supports(CredentialItem... items) {
 		for (final CredentialItem item : items) {
-			if (item instanceof CredentialItem.StringType && item.getPromptText().startsWith(PROMPT)) {
+			if (item instanceof CredentialItem.InformationalMessage) {
 				continue;
 			}
-			else {
-				return false;
+			if (item instanceof CredentialItem.Password && item.getPromptText().equals(PROMPT)) {
+				continue;
 			}
+			return false;
 		}
 		return true;
 	}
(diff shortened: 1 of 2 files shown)


spencergibb (approved):

kvmw:
@ryanjbaxter  judging based on `UsernamePasswordCredentialsProvider` implementation by JGit, i think, both `supports` and `get` methods should accept `CredentialItem.InformationalMessage` as well as `CredentialItem.Password` and `get` method should throw exception if any `CredentialItem` other than above types was in the list. 

I've drafted the #2210 based on your changes and what I understood from the `UsernamePasswordCredentialsProvider` implementation, just to explain my points.

ryanjbaxter:
@kvmw thanks!  I updated the PR based on your feedback

kvmw:
> @kvmw thanks! I updated the PR based on your feedback

Thanks! LGTM.

PR #2219, opened 2023-01-31
Fix AWS S3 ResourceController support

### LINKED ISSUE #2216: S3 and Git backends have different behaviors for file handling
**Describe the bug**
Spring Cloud Version: `2022.0.0`

We are comparing a Git and S3 backend to better understand the differences, specifically for serving text files. There's some different behavior between requesting files between the S3 and Git backends and was trying to understand why this is expected. 

Unless we're misunderstanding, it makes switching out the back-ends impact the clients. Only handling JSON files in S3 is fine for now, but we would have expected the application name and routes to be consistent between the two back-ends.

In the S3 implementation, it's treating the `{application}` as part of the object key prefix so the following requests work differently between the back-ends.   

Git:
`curl localhost:8080/foo/default/master/data1.json` (works)

S3:

`curl localhost:8080/foo/default/master/data1.json` (does not work)
`curl localhost:8080/foo/default/master/data1` (does not work)
`curl localhost:8080/data/default/master` (works)

Thanks for any guidance. 

**Sample**

Git Repo and Bucket have a single file called `data1.json`. In Git, it is in the `master` branch. In the S3 bucket, there is a `master` folder in bucket `bucket1`

S3 Config
```
awss3:
  bucket: bucket1
  region: us-east-1
  endpoint: http://localhost:9000
```

Git Config
```
git:
  uri: file:///Users/tgaines/dev/config-data
  default-label: master
  clone-on-start: true
```

ryanjbaxter:
There are several problems here:

* Fixes and improves documentation
* Fixes a bug where we were using the application as the bucket folder name when it should have been the label
* Uses the default label property if set and the query parameter useDefaultLabel is supplied
* Fixes the resource location check so it takes into account the encoded / in the path name of an S3 object

There is one use case that requires a fix in Spring Cloud AWS, and that is when there is no label supplied and there is no default label set.  There appears to be a. bug in the way the createRelative method creates the S3 object name in this case.  I have submitted a PR for this https://github.com/awspring/spring-cloud-aws/issues/648

Fixes #2216

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -1626,11 +1626,12 @@ server {
 [[spring-cloud-config-serving-plain-text-aws-s3]]
 === AWS S3
 
-To enable serving plain text for AWS s3, the Config Server application needs to include a dependency on Spring Cloud AWS.
+To enable serving plain text for AWS s3, the Config Server application needs to include a dependency on `spring-cloud-aws-context`.
 For details on how to set up that dependency, see the
-https://cloud.spring.io/spring-cloud-static/spring-cloud-aws/2.1.3.RELEASE/single/spring-cloud-aws.html#_spring_cloud_aws_maven_dependency_management[Spring Cloud AWS Reference Guide].
+https://docs.awspring.io/spring-cloud-aws/docs/2.4.3/reference/html/index.html#spring-cloud-aws-maven-dependency-management[Spring Cloud AWS Reference Guide].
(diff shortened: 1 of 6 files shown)


docs/src/main/asciidoc/spring-cloud-config.adoc
```diff
@@ -1626,11 +1626,12 @@ server {
 [[spring-cloud-config-serving-plain-text-aws-s3]]
 === AWS S3
 
-To enable serving plain text for AWS s3, the Config Server application needs to include a dependency on Spring Cloud AWS.
+To enable serving plain text for AWS s3, the Config Server application needs to include a dependency on `spring-cloud-aws-context`.
```
spencergibb, line 1629: Maybe add group id here?

spencergibb (approved):

Issue #2222, opened 2023-02-02
Misleading variable name in 3.1.x ``AwsS3EnvironmentRepository``


deeaitche:
The name of variable ``json`` at line 146-147 seems misleading.

https://github.com/spring-cloud/spring-cloud-config/blob/85314467fe6a85de1ce4ad1fbd7e8a763204ab60/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java#L144-L153

PR #2223, opened 2023-02-03
Fix variable name in AwsS3EnvironmentRepository.java for readability

deeaitche:
Issue #2222 

I have renamed the variable to make it more readable

The object is yaml, not json.


deeaitche:
@pivotal-cla This is an Obvious Fix”

PR #2228, opened 2023-02-08
Remove withReuse(true) in AwsS3IntegrationTests

eddumelendez:
Reusable containers is an experimental feature which needs additional
configuration in order to enable it. This is used in test and it
is not inteded to keep the container running.

See more https://www.testcontainers.org/features/reuse/


ryanjbaxter:
Thanks @eddumelendez!  I was in the code making some other changes and just removed the `.withReuse(true)` call.

PR #2243, opened 2023-03-22
Put ExceptionHandler into a ResourceControllerAdvice to allow overriding behavior

### LINKED ISSUE #2192: Custom exception handling for NoSuchResourceException is not working
**Describe the bug**
Spring cloud version: **2021.0.5**
I use custom exception handling in my spring-cloud-config-server application.
And the problem is that I could not overwrite exception handling for `NoSuchResourceException` that is defined in `org.springframework.cloud.config.server.resource.ResourceController` (lines 218-221), because it is being handled inside of the controller, which takes higher priority.
The question is how could I handle this exception in my controller advice.

**Sample**
The sample project could be found [here](https://github.com/isvyry/spring-cloud-config-exception-issue)

ryanjbaxter:
Fixes #2192

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/resource/ResourceController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/resource/ResourceController.java
@@ -29,17 +29,14 @@ import org.springframework.cloud.config.environment.Environment;
 import org.springframework.cloud.config.server.encryption.ResourceEncryptor;
 import org.springframework.cloud.config.server.environment.EnvironmentRepository;
 import org.springframework.core.io.Resource;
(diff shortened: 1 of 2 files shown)


spencergibb:
How does moving it to `ResourceControllerAdvice` allow overriding?

ryanjbaxter:
Any `@ExceptionHandler` within the `Controller` will be given priority if an exception is thrown within the `Controller`, it basically makes it impossible to change the way the exception is handled.

By moving it to `@ControllerAdvice` you can set order by using `@Ordered`.  I set ours to be lowest so someone else could set a higher priority in their own `@ControllerAdvice` to handle the exception themselves.

ryanjbaxter:
Thats because I just added it, you didn't miss it 🤣

Issue #2252, opened 2023-04-16
encrypt and decrypt endpoint lost data for encryptor retrievement


RouxAntoine:
**Version**

Spring cloud config server 4.0.2

**Explanation**

in this commit b2508eb97c6aac0284896d83edd50c38c83d1e87

encrypt and decrypt controller

https://github.com/spring-cloud/spring-cloud-config/commit/b2508eb97c6aac0284896d83edd50c38c83d1e87#diff-f18a83930fb437413479ed50adc18dedbf115b37e560d8481185a796c24444d8R107-L131

https://github.com/spring-cloud/spring-cloud-config/commit/b2508eb97c6aac0284896d83edd50c38c83d1e87#diff-f18a83930fb437413479ed50adc18dedbf115b37e560d8481185a796c24444d8R127-L150

stop to propagate data variable to `EnvironmentPrefixHelper.getEncryptorKeys()`

**Sample with custom TextEncryptorLocator**

```java
@Configuration(proxyBeanMethods = false)
public class Encryption {

    private final DefaultTextEncryptionAutoConfiguration defaultTextEncryptionAutoConfiguration;

    public Encryption() {
        defaultTextEncryptionAutoConfiguration = new DefaultTextEncryptionAutoConfiguration();
    }

    @Bean
    public TextEncryptorLocator textEncryptorLocator(MultiKeyProperties multiKeyProperties) {
        return keys -> {
            KeyProperties keyPropertiesForClient = multiKeyProperties.encrypt().get(keys.get("application"));
            return defaultTextEncryptionAutoConfiguration.defaultTextEncryptor(keyPropertiesForClient);
        };
    }

}
```

```
$ curl -s localhost:8090/encrypt --data-urlencode '{"key":"one","secret":"value"}Hello world 2'
```

the `keys` map contained nothing except `name` and `profiles` entry

<img width="1133" alt="image" src="https://user-images.githubusercontent.com/5455181/232349017-7e61b59c-1725-47f5-bf8e-a95dd7207d3e.png">

PR #2253, opened 2023-04-16
fix: transmit data from spring cloud config server controller to getEncryptor

RouxAntoine:
attention point to methods

https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java#L93

https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java#L84

which always have same bug but could not be properly fixed due to missing data in Get request

Fixes gh-2252

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
@@ -103,12 +103,12 @@ public class EncryptionController {
 	@PostMapping("/encrypt/{name}/{profiles}")
 	public String encrypt(@PathVariable String name, @PathVariable String profiles, @RequestBody String data,
 			@RequestHeader("Content-Type") MediaType type) {
-		TextEncryptor encryptor = getEncryptor(name, profiles, "");
-		validateEncryptionWeakness(encryptor);
 		String input = stripFormData(data, type, false);
+		TextEncryptor encryptor = getEncryptor(name, profiles, input);
+		validateEncryptionWeakness(encryptor);
 		Map<String, String> keys = helper.getEncryptorKeys(name, profiles, input);
 		String textToEncrypt = helper.stripPrefix(input);
-		String encrypted = helper.addPrefix(keys, encryptorLocator.locate(keys).encrypt(textToEncrypt));
+		String encrypted = helper.addPrefix(keys, encryptor.encrypt(textToEncrypt));
 		if (logger.isInfoEnabled()) {
 			logger.info("Encrypted data");
 		}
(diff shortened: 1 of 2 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
```diff
@@ -103,7 +103,7 @@ public String encrypt(@RequestBody String data, @RequestHeader("Content-Type") M
 	@PostMapping("/encrypt/{name}/{profiles}")
 	public String encrypt(@PathVariable String name, @PathVariable String profiles, @RequestBody String data,
 			@RequestHeader("Content-Type") MediaType type) {
-		TextEncryptor encryptor = getEncryptor(name, profiles, "");
+		TextEncryptor encryptor = getEncryptor(name, profiles, data);
```
ryanjbaxter, line 106: Shouldn't we be passing `input` instead?  Thats what the original code used

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
```diff
@@ -123,7 +123,7 @@ public String decrypt(@RequestBody String data, @RequestHeader("Content-Type") M
 	@PostMapping("/decrypt/{name}/{profiles}")
 	public String decrypt(@PathVariable String name, @PathVariable String profiles, @RequestBody String data,
 			@RequestHeader("Content-Type") MediaType type) {
-		TextEncryptor encryptor = getEncryptor(name, profiles, "");
+		TextEncryptor encryptor = getEncryptor(name, profiles, data);
```
ryanjbaxter, line 126: We don't even need this line, `encryptor` is not used in the `catch` block so we can just move the variable definition into the `try` block which already passes `data`.

ryanjbaxter (changes requested):
Could you please add tests?

ryanjbaxter:
Ping @RouxAntoine see feedback

RouxAntoine:
> Ping @RouxAntoine see feedback

hello, sorry for delay not so many time on my part I take a look

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
RouxAntoine: done in commit e7669e1b and use retrieved encryptor to encrypt message here https://github.com/spring-cloud/spring-cloud-config/pull/2253/files#diff-f18a83930fb437413479ed50adc18dedbf115b37e560d8481185a796c24444d8R111

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
RouxAntoine: done in commit e7669e1b

RouxAntoine:
I just added two assert in a existing test. To check the key in encryptorLocator

PR #2261, opened 2023-05-02
Include property sources without profile and include application property source

### LINKED ISSUE #1911: S3 Support - Sharing Configuration With All Applications
Hello,

I was reading the documentation section [Sharing Configuration With All Applications](https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#_sharing_configuration_with_all_applications), and there is no support for the S3 backend. Are you aware is there is a technical limitation to add such support? or if that would be something worth doing?

I was looking at the code in the [AwsS3EnvironmentRepository](https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java#L70), and with my limited knowledge of the project, I was thinking it could "just" be looking for the `application*` files as well to see if there was a shared properties already defined.

ryanjbaxter:
Fixes #1911

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
@@ -18,6 +18,9 @@ package org.springframework.cloud.config.server.environment;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Properties;
 
 import com.amazonaws.services.s3.AmazonS3;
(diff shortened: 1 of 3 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -68,45 +72,53 @@ public void setOrder(int order) {
 
 	@Override
 	public Environment findOne(String specifiedApplication, String specifiedProfiles, String specifiedLabel) {
-		final String application = StringUtils.isEmpty(specifiedApplication)
+		final String application = ObjectUtils.isEmpty(specifiedApplication)
```
spencergibb, line 75: Why the change from StringUtils to ObjectUtils?

spencergibb (approved):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
ryanjbaxter: It was deprecated so I figured I would just switch it while I was in the file

spencergibb: right

PR #2260, opened 2023-05-02
Support for profile specific config data loading

ryanjbaxter:
Fixes #1922

Compliments https://github.com/spring-cloud/spring-cloud-commons/pull/1228

PR #2269, opened 2023-05-21
Pass binder to ConfigServerInstanceProvider

ryanjbaxter:
Now that we the config data loader loads configuration in two phases, the first phase does not allow us to load configuration from the BootstrapContext because we have not finished loading configuration data yet.  Therefore it is useful to pass the Binder from the ResolverContext which contains the config data that has already been loaded in the first phase.  

Related to changes made in https://github.com/spring-cloud/spring-cloud-config/pull/2260

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerInstanceProvider.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerInstanceProvider.java
@@ -21,6 +21,8 @@ import java.util.List;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.boot.context.properties.bind.BindHandler;
+import org.springframework.boot.context.properties.bind.Binder;
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.retry.annotation.Retryable;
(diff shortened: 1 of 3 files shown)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerInstanceProvider.java
```diff
@@ -70,6 +88,10 @@ public interface Function {
 
 		List<ServiceInstance> apply(String serviceId);
 
+		default List<ServiceInstance> apply(String serviceId, Binder binder, BindHandler bindHandler, Log log) {
```
spencergibb, line 91: I'm a big confused by this. Why have this and the ones passed into the constructor?

ryanjbaxter: The ones in the constructor of `ConfigServerInstanceProvider` are then passed when it calls the `Function` above.

spencergibb: Why can't constructor binder etc work in all cases?

ryanjbaxter: They are the same.

They are passed in the constructor of `ConfigServerInstanceProvider` then used in `ConfigServerInstanceProvider.getConfigServerInstances` when it calls `Function.apply`.

spencergibb: Right, the other impls of function need them, not config client

PR #2278, opened 2023-06-08
Implement a better equals and hashcode method ConfigServerConfigDataResource

ryanjbaxter:
Spring Boot keeps a list of ConfigDataResources and will call contains on the list to determine whether it should load the resource.  Our hashcode and equals methods were using variables in ConfigServerConfigDataResource which had no bearing on the configuration that would be loaded, so this change simplifies this to make sure two ConfigServerConnfigDataResources would be considered equal if they would end up making the same request to the config server for configuration.

### DIFF
--- a/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientConfigDataLoaderTest.java
+++ b/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientConfigDataLoaderTest.java
@@ -46,7 +46,7 @@ import static org.mockito.Mockito.when;
  * spring.application.name will be set.
  *
  * At this point Boot has collected all active profiles so it will load all
- * spring.config.import statements again with active profiles. The subsequent 2 calls then
+ * spring.config.import statements again with active profiles. The subsequent call then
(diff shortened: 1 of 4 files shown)


spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientConfigDataLoaderTest.java
```diff
@@ -73,7 +73,7 @@ void context() {
 		verify(rest).exchange(eq("http://localhost:8888/{name}/{profile}"), eq(HttpMethod.GET),
 				ArgumentMatchers.any(HttpEntity.class), eq(Environment.class), eq("application"),
 				ArgumentMatchers.<String>any());
-		verify(rest, times(2)).exchange(eq("http://localhost:8888/{name}/{profile}"), eq(HttpMethod.GET),
+		verify(rest, times(1)).exchange(eq("http://localhost:8888/{name}/{profile}"), eq(HttpMethod.GET),
```
spencergibb, line 76: Awesome, we were testing for the bug 🤦🏻‍♂️

PR #2287, opened 2023-06-14
Configures SshSessionFactory only if there is at least one git uri with ssh scheme

### LINKED ISSUE #2284: java.lang.AssertionError when using HTTP/HTTPS url for the `git.uri` with `-enableassertions` flag enabled.
Both [FileBasedSshSessionFactory](https://github.com/spring-cloud/spring-cloud-config/blob/8ca88f30e5f9b2ae6af7e0a7947c9ebef639d2a6/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/FileBasedSshSessionFactory.java#L36) and [PropertyBasedSshSessionFactory](https://github.com/spring-cloud/spring-cloud-config/blob/8ca88f30e5f9b2ae6af7e0a7947c9ebef639d2a6/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/PropertyBasedSshSessionFactory.java#L70) are expecting at least one `ssh` URL in the git configuration. So a  HTTP/HTTPS URL for `git.uri`  can cause `java.lang.AssertionError`, if `-enableassertions` flag is true. 

The issue in often not visible because above flag is `false` by default. But in development or test environment, which the flag might be enabled, one can see the issue. 

**Sample**
Running any sample config-server application with above flag enabled can show the issue.
```
java -ea -jar ./build/libs/config-server-sample-0.0.1-SNAPSHOT.jar
```

**Possible Solution**
`FileBasedSshTransportConfigCallback` and `PropertyBasedSshTransportConfigCallback` should configure the `SshTransport` only if there is at least one SSH URL in the configuration.

current code:
```java
@Override
public void configure(Transport transport) {
  if (transport instanceof SshTransport) {
    ((SshTransport) transport).setSshSessionFactory(new FileBasedSshSessionFactory(
		new SshUriPropertyProcessor(this.sshUriProperties).getSshKeysByHostname()));
  }
}
```
proposed change:
```java
@Override
public void configure(Transport transport) {
  if (transport instanceof SshTransport) {
        var sshKeysByHostname = new SshUriPropertyProcessor(this.sshUriProperties).getSshKeysByHostname();
        if (!sshKeysByHostname.isEmpty()) {  // Only if the map is not empty
	  ((SshTransport) transport).setSshSessionFactory(new FileBasedSshSessionFactory(sshKeysByHostname));
	}
  }
}
```

kvmw:
closes #2284

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/FileBasedSshTransportConfigCallback.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/FileBasedSshTransportConfigCallback.java
@@ -16,10 +16,13 @@
 
 package org.springframework.cloud.config.server.ssh;
 
+import java.util.Map;
+
 import org.eclipse.jgit.api.TransportConfigCallback;
 import org.eclipse.jgit.transport.SshTransport;
 import org.eclipse.jgit.transport.Transport;
 
+import org.springframework.cloud.config.server.environment.JGitEnvironmentProperties;
 import org.springframework.cloud.config.server.environment.MultipleJGitEnvironmentProperties;
 
 /**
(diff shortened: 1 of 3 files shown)


spencergibb (changes requested):
Unfortunately, we can't delete or rename public classes and there isn't going to be a major release to do that soon.

kvmw:
> Unfortunately, we can't delete or rename public classes and there isn't going to be a major release to do that soon.

Reverted those changes.

ryanjbaxter (approved):
Is this only an issue in main, or is it also present in 3.1.x?

kvmw:
> Is this only an issue in main, or is it also present in 3.1.x?

3.1.x has the same issue.

kvmw:
> closes #2284

@ryanjbaxter 
#2288 is ready. but the code is slightly different.

ryanjbaxter:
I merged things forward, I think I did it correct, but you might want to double check me ;)

kvmw:
@ryanjbaxter  The merge looks fine. Just that @RunWith annotations are not needed in main branch. 

PR: https://github.com/spring-cloud/spring-cloud-config/pull/2290

Issue #1386, opened 2019-05-17
IllegalStateException: Cannot decrypt: key=cloud.config.password at 


ubalaji:
I have configserver deployed working, able to encrypt and decrypt passwords using url localhost:8888/encrypt and localhost:8888/decrypt. I encrypted password used in the config server security.user.password.

I added encryted securitykey to config client bootstrap properties. Now bootsttrap for config client as follows.

encrypt:
  key: caoConfigserverspring
application:
  name: xyz
cloud: 
  config: 
    uri: http://localhost:8888
    enabled: true
    fail-fast: true
    username: my_user
    password: '{cipher}1234567890hjdbcbvnhfg'
server:
  port: 8080

I tried password without single quote and application fails with parsing error. I tried with plain as password and the application works fine. When I use encrypted value, while deploying I see following error. What am I missing when using encrypted password?

2019-05-17 13:03:53 - Application run failed
java.lang.IllegalStateException: Cannot decrypt: key=cloud.config.password
        at org.springframework.cloud.bootstrap.encrypt.EnvironmentDecryptApplicationInitializer.doDecrypt(EnvironmentDecryptApplicationInitializer.java:238) ~[spring-cloud-context-2.1.0.RELEASE.jar!/:2.1.0.RELEASE]
        at org.springframework.cloud.bootstrap.encrypt.EnvironmentDecryptApplicationInitializer.decrypt(EnvironmentDecryptApplicationInitializer.java:169) ~[spring-cloud-context-2.1.0.RELEASE.jar!/:2.1.0.RELEASE]
        at org.springframework.cloud.bootstrap.encrypt.EnvironmentDecryptApplicationInitializer.initialize(EnvironmentDecryptApplicationInitializer.java:96) ~[spring-cloud-context-2.1.0.RELEASE.jar!/:2.1.0.RELEASE]
        at org.springframework.cloud.bootstrap.BootstrapApplicationListener$DelegatingEnvironmentDecryptApplicationInitializer.initialize(BootstrapApplicationListener.java:407) ~[spring-cloud-context-2.1.0.RELEASE.jar!/:2.1.0.RELEASE]
        at org.springframework.boot.SpringApplication.applyInitializers(SpringApplication.java:649) [spring-boot-2.1.4.RELEASE.jar!/:2.1.4.RELEASE]
[...]

PR #2334, opened 2023-09-30
Additional information in the log output when an InvalidCipherException occurs.

kworkbee:
When a client application is starting with a mismatch symmetric key (between client and server), the output log prints before InvalidCipherException is thrown due to un-decryptable configuration properties, which creates additional information with log message because the message alone lacks information to guess why the InvalidCipherException occurred.

Relates to #1386

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
@@ -136,7 +136,8 @@ public class EncryptionController {
 		}
 		catch (IllegalArgumentException | IllegalStateException e) {
 			if (logger.isErrorEnabled()) {
-				logger.error("Cannot decrypt key:" + name + ", value:" + data, e);
+				logger.error("Cannot decrypt key:" + name + ", value:" + data +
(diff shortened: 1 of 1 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/encryption/EncryptionController.java
```diff
@@ -136,7 +136,8 @@ public String decrypt(@PathVariable String name, @PathVariable String profiles,
 		}
 		catch (IllegalArgumentException | IllegalStateException e) {
 			if (logger.isErrorEnabled()) {
-				logger.error("Cannot decrypt key:" + name + ", value:" + data, e);
+				logger.error("Cannot decrypt key:" + name + ", value:" + data +
+					", Please verify if symmetric key is set correctly", e);
```
kworkbee, line 140: What Symmetric Key means here is the `encrypt.key` property.

ryanjbaxter: I would suggest your replace `symmetric key` with `encrypt.key`

ryanjbaxter (changes requested):

kworkbee:
@ryanjbaxter I fixed the message.

PR #2353, opened 2023-11-15
Add support for random exponential backoff policy

### LINKED ISSUE #2351: Reduce risk of thundering herd issues in retry logic

In an application using the cloud-config server we recently observed a situation, where the config server seemingly became overloaded due to a large amount of config client requests. We are still investigating the issue, but during the investigation we figured there is potential room for improvement in the retry logic of the config-client.

By default the config client uses a `ExponentialBackOffPolicy` with an initial interval of 1000, a multiplier of 1.1 and a maximum of 6 retries. This policy would lead to do retries after roughly 1000, 1100ms, 1210ms, 1331ms, 1464ms and 1610ms. This retry logic is fine for many scenarios, but can cause a [thundering herd problematic](https://en.wikipedia.org/wiki/Thundering_herd_problem) in others.

In our case (as it certainly might be very common), both client and server were running in Kubernetes, so while the client might have aborted after 6 attempts and stopped with an exception, the pod would be restarted and try again (with kubelet-controlled backoff between the restarts). We believe that this led to a situation where our system wasn't able to recover on its own, because the retry-behaviour of the client added salt to injury with respect to the overloaded config-server, since essentially on all retries each client would try it again at the same time.

The standard solution to a thundering herd problem is to add some random jitter to the backoff intervals. 

As spring-cloud-config uses spring-retry, a solution could be to use the `ExponentialRandomBackOffPolicy` (https://github.com/spring-projects/spring-retry/blob/main/src/main/java/org/springframework/retry/backoff/ExponentialRandomBackOffPolicy.java) instead of the `ExponentialBackOffPolicy` by default or make it configurable as an alternative to the current default.

**Describe alternatives you've considered**

As your documentation suggests that there is the possibility to implement a `RetryOperationsInterceptor`, this might be an alternative worth consideration by individual users like us. I haven't yet looked into it in detail, because it seems that having some random component for the delay in spring cloud directly could make a safer default.

ryanjbaxter:
Fixes #2351

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-config.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-config.adoc
@@ -1906,6 +1906,7 @@ First, you need to set `spring.cloud.config.fail-fast=true`.
 Then you need to add `spring-retry` and `spring-boot-starter-aop` to your classpath.
 The default behavior is to retry six times with an initial backoff interval of 1000ms and an exponential multiplier of 1.1 for subsequent backoffs.
 You can configure these properties (and others) by setting the `spring.cloud.config.retry.*` configuration properties.
+To use a random exponential backoff policy set `spring.cloud.config.retry.useRandomPolicy` to `true`.
 
 TIP: To take full control of the retry behavior and are using legacy bootstrap, add a `@Bean` of type `RetryOperationsInterceptor` with an ID of `configServerRetryInterceptor`.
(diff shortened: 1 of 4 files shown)


spencergibb (approved):

marnee01:
Note that I created issue https://github.com/spring-cloud/spring-cloud-config/issues/2378 to request additional documentation.

PR #2354, opened 2023-11-15
Do not combine profiles when spring.cloud.config.pofiles is set.

### LINKED ISSUE #2349: Version 4.0.4, unable to connect to Cloud Config if we have cloud config profile and spring environment profile.
Description:

We have a profile property to connect to appconfig.(spring.cloud.config.profile)
with spring-cloud-config-client-4.0.4.jar, in class ConfigServicePropertySourceLocator, there is  a new method which combines profile from property and environment with comma separated.
So, finally profile value while connecting to appconfig will be like below and failing app-config call.

HTTP GET https://XXXXXXXXXX/appconfig/properties/<ModuleID>/<PROFILE-XYX,ABC>/<LABEL>
This is failing.

Even though profile properties are overrider at line 122 ConfigClientProperties properties = this.defaultProperties.override(environment);

in the next line, profile value is getting updated by appending with all the profiles
properties.setProfile(String.join(",", combineProfiles(properties, environment)));

So in method getRemoteEnvironment, while preparing the URL, profile path param is goign as comma separated and there by causing issue.

ryanjbaxter:
Fixes #2349

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
@@ -50,7 +50,6 @@ import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.retry.annotation.Retryable;
 import org.springframework.util.Assert;
-import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.HttpServerErrorException;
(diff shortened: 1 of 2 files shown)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java
```diff
@@ -92,6 +92,9 @@ private List<String> combineProfiles(ConfigClientProperties properties,
 		if (!ObjectUtils.isEmpty(properties.getProfile())) {
 			combinedProfiles = Stream.of(properties.getProfile().split(",")).map(String::trim).filter(s -> !s.isEmpty())
 					.collect(Collectors.toList());
+			// If spring.cloud.config.profiles is set it should override all other
+			// profiles set other ways
+			return combinedProfiles;
```
spencergibb, line 97: This makes this method weird right? It's no longer combining profiles. It's returning either the config props profiles, or the active profiles, or default profiles

spencergibb:
Nothing else broke with the build, so maybe it is ok, but the terminology of the method and some variables don't make sense anymore.

ryanjbaxter:
@spencergibb let me know if this makes things clearer

spencergibb (approved):

PR #2357, opened 2023-11-25
Customize AwsSecretsManagerEnvironmentRepository.

ojecborec:


spencergibb (changes requested):
I'm afraid there can't be breaking changes. Can you describe exactly what you are trying to accomplish with this?

ojecborec:
The foundation for this PR is coming from real life experience. I’m reading config properties from 2 sources. One is Git which contains non sensitive data such as server port, context path etc. Every time I release to production I create a new tag, i.e. v1.0. That gives me ability to rollback my config. Then there are secrets, passwords etc. They’re stored in the AWS Secrets Manager.  These might change regularly. Rolling back to older version doesn’t work as this password might be invalid. So I don’t version secrets. When I make /{application}/{environment}/{version} request it would read secrets from AWS Secrets Manager only labeled as {version}. As these properties don’t exist I get empty response.

I know I can create my own repository but there are several drawbacks. 
1. AwsSecretsManagerEnvironmentRepository is class, not interface. 
2. Extending this class means calling super constructor that has 3 specific arguments which my class might not need. There’s no default constructor. 
3. In order to get instances of awsSmClient I need to create additional builder class/method and build it myself. And repository factory as well just like you do. 
4. I don’t want to copy/paste the whole repository class and amend only line, that is the AWS request. I would have to keep it in sync with origin in case anything changes.

This looks like too much hassle and risk if anything changes in the upstream for such a small request.

spencergibb:
Thanks for the detailed response. Changing from a class to an interface is not possible right now. We can add constructors, if that helps. If you are willing to rework, let me know.

spencergibb:
The build failures are checkstyle errors

ojecborec:
Can you have a look at this alternative solution https://github.com/spring-cloud/spring-cloud-config/pull/2358?

ryanjbaxter:
Should we close this PR then?

ojecborec:
Yes, I’m closing it.

PR #2358, opened 2023-11-26
Do not set staging label when fetching the secret values.

ojecborec:
Alternative to https://github.com/spring-cloud/spring-cloud-config/pull/2357.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentProperties.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentProperties.java
@@ -55,6 +55,11 @@ public class AwsSecretsManagerEnvironmentProperties implements EnvironmentReposi
 	private String defaultLabel;
 
 	/**
+	 * Do not set staging label when fetching the secret values.
+	 */
+	private boolean ignoreLabel;
+
+	/**
 	 * The order of the environment repository.
 	 */
 	private int order = Ordered.LOWEST_PRECEDENCE;
@@ -103,6 +108,14 @@ public class AwsSecretsManagerEnvironmentProperties implements EnvironmentReposi
 
 	public void setDefaultLabel(String defaultLabel) {
 		this.defaultLabel = defaultLabel;
+	}
+
+	public boolean isIgnoreLabel() {
+		return this.ignoreLabel;
+	}
+
+	public void setIgnoreLabel(boolean ignoreLabel) {
+		this.ignoreLabel = ignoreLabel;
 	}
 
 	public int getOrder() {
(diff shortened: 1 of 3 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentProperties.java
```diff
@@ -54,6 +54,11 @@ public class AwsSecretsManagerEnvironmentProperties implements EnvironmentReposi
 	 */
 	private String defaultLabel;
 
+	/**
+	 * Do not set staging label when fetching the secret values.
+	 */
+	private boolean ignoreLabel;
```
ryanjbaxter, line 60: The default value should be `false` in order to maintain the default behavior

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentRepository.java
```diff
@@ -86,6 +87,10 @@ public Environment findOne(String application, String profileList, String label)
 			label = defaultLabel;
 		}
 
+		if (ignoreLabel) {
```
ryanjbaxter, line 90: This can be combined with the `if` statement above

ryanjbaxter (commented):
Can you submit this against the 3.1.x branch

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentProperties.java
ojecborec: Primitive boolean is false by default.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentRepository.java
ojecborec: The above `if` statement is checking `label` provided by request and setting it to `defaultLabel` if empty. This statement set `label` to null regardless the value of `label` or `defaultLabel`.

ojecborec:
Looks like `AwsSecretsManagerEnvironmentRepository` in the `3.1.x` branch does not utilize label when reading of Sercets Manager at all.

```java
GetSecretValueRequest request = new GetSecretValueRequest().withSecretId(path);
```

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentRepository.java
ryanjbaxter: My thought was why set `label` if we are going to ignore the label anyways?

ojecborec: Understand. I've updated the code.

ojecborec:
Sure. PR's been created https://github.com/spring-cloud/spring-cloud-config/pull/2359.

PR #2375, opened 2024-01-18
Register a new class to resolve properties during bootstrap

ryanjbaxter:
This registers a new class which contains a `Binder` and `BindHandler` which comes from the Config Data profile resolution from Boot.

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.config.client;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
+import java.util.function.Supplier;
 
 import org.apache.commons.logging.Log;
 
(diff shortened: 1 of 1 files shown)


spencergibb (approved):

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
```diff
@@ -273,4 +277,25 @@ private class PropertyHolder {
...
+			this.binder = binder;
+			this.bindHandler = bindHandler;
+		}
+
+		public <T> T get(String key, Class<T> type, T defaultValue) {
+			return binder.bind(key, Bindable.of(type)).orElse(defaultValue);
+		}
+
+		public <T> T resolveConfigurationProperties(String prefix, Class<T> type, Supplier<T> supplier) {
+			return binder.bind(prefix, Bindable.of(type), bindHandler).orElseGet(supplier);
```
pantherdd, line 296:
```suggestion
		public <T> T resolveConfigurationProperties(String prefix, Class<T> type, Supplier<T> defaultValue) {
			return binder.bind(prefix, Bindable.of(type), bindHandler).orElseGet(defaultValue);
```

Issue #2401, opened 2024-04-07
Enable force refresh by forceRefresh query parameter regardless of refreshRate


opeco17:
**Is your feature request related to a problem? Please describe.**
[refreshRate](https://cloud.spring.io/spring-cloud-static/spring-cloud-config/2.0.2.RELEASE/single/spring-cloud-config.html#_git_refresh_rate) is useful to scale Cloud Config clients.
However, we want to make sure that some important clients can get the latest configurations regardless of refreshRate.

**Describe the solution you'd like**
Allow force refresh regardless of refreshRate by using forceRefresh query parameter.

**Describe alternatives you've considered**
We can temporarily update refreshRate to 0 while some important clients access Cloud Config
But the new refreshRate is effective for all Cloud Config clients and Config Server might receive huge traffic during that period.
Also updating refreshRate manually every time is tough for operators.

**Additional context**
PR: https://github.com/spring-cloud/spring-cloud-config/pull/2402

spencergibb:
Closing in favor of #2402

PR #2402, opened 2024-04-07
Enable force refresh by forceRefresh query parameter

opeco17:
Fixes https://github.com/spring-cloud/spring-cloud-config/issues/2401

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AbstractScmEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AbstractScmEnvironmentRepository.java
@@ -57,11 +57,17 @@ public abstract class AbstractScmEnvironmentRepository extends AbstractScmAccess
 
 	@Override
 	public synchronized Environment findOne(String application, String profile, String label, boolean includeOrigin) {
+		return findOne(application, profile, label, includeOrigin, false);
+	}
+
+	@Override
+	public synchronized Environment findOne(String application, String profile, String label, boolean includeOrigin,
+			boolean forceRefresh) {
 		NativeEnvironmentRepository delegate = new NativeEnvironmentRepository(getEnvironment(),
 				new NativeEnvironmentProperties(), this.observationRegistry);
-		Locations locations = getLocations(application, profile, label);
+		Locations locations = getLocations(application, profile, label, forceRefresh);
 		delegate.setSearchLocations(locations.getLocations());
-		Environment result = delegate.findOne(application, profile, "", includeOrigin);
+		Environment result = delegate.findOne(application, profile, "", includeOrigin, forceRefresh);
 		result.setVersion(locations.getVersion());
 		result.setLabel(label);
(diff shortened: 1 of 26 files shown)


ryanjbaxter:
This would be a good enhancement for our next major release where we can introduce some breaking changes.  We think it would be to explore introducing a "context object" so that next time we need to pass a parameter like this down the stack we aren't changing a bunch of method signatures to do so.

opeco17:
@ryanjbaxter 

Thank you for the review.
May I know which approach you prefer regarding context implementation?

**Approach A (Example)**

```java
public class RequestContext {
    private static final ThreadLocal<Map<String, String>> context = new ThreadLocal<>();

    public static void setContext(Map<String, String> value) {
        context.set(value);
    }

    public static Map<String, String> getContext() {
        return context.get();
    }

    public static void clear() {
        context.remove();
    }
}
```

```java
public class Service {
    public void doSomething() {
        String forceRefreshStr = RequestContext.get("forceRefresh");
        boolean forceRefresh = false;
        if (forceRefreshStr != null) {
            boolean forceRefres = Boolean.parseBoolean(forceRefreshStr);
        }
        // do something
    }
}
```

**Approach B (Example)**

```java
public class RequestContext {
    private boolean forceRefresh;

    public RequestContext(boolean forceRefresh) {
        this.forceRefresh = forceRefresh;
    };

    public boolean getForceRefresh() {
        return forceRefresh;
    }
}
```

```java
public class Service {
    public void doSomething(RequestContext ctx) {
        boolean forceRefresh = ctx.getForceRefresh();
        // do something
    }
}
```

ryanjbaxter:
Personally I would prefer approach B, but I would remove the constructor and just use setters and getters.

opeco17:
@ryanjbaxter 

I've implemented `RequestContext` to pass query parameters to downstream methods.
When we introduce new query parameters that need to be passed to downstream methods like `forceRefresh` next time, we can add them to `RequestContext`.

spencergibb (changes requested):
Request context should include more than just force refresh. As many as possible, if not all of them

opeco17:
@spencergibb 

Should we add both query parameters and path parameters to `RequestContext` or just query parameters?

ryanjbaxter:
I think we would want query and path parameters (name, profile, label, etc).  

And I agree it is a big changes, hence why it makes sense for a major release.  IMO I tink the request context change should be separated out into its own PR.

opeco17:
@ryanjbaxter

I agree with you

What about to add just query parameters to `RequestContext` in this PR and work on other parameters in another PR?
Or it's also reasonable to remove `RequestContext` from this PR completely and introduce `RequestContext` in another PR in my opinion

I can work on another PR to complete `RequestContext` as quickly as possible once the direction is decided

ryanjbaxter:
I would't remove `RequestContext` from this PR.  What I would suggest is we introduce `RequestContext` and move the existing parameters to it in another PR and then in this PR add `forceRefresh` to `RequestContext`.  We would need to merge the first PR before merging this PR.

PR #2403, opened 2024-04-09
Git shallow clone for better performance

### LINKED ISSUE #1544: Shallow clone for better performance
We regularly see response times from our config server of 15-25 seconds. We have on average ~30 requests with response times > 15s each day. These high response times happen when config server is fetching from one of our repos and there are updates in that repo. There are often updates since any time someone creates a new feature branch and makes a change to it, that is an update.

This repo has many, many files in it (for multiple applications). Some of the files have 1000s of lines. But the biggest culprit, I believe, is that it has a huge amount of history.

We are looking at potentially breaking it into multiple repos (and use the new composite repos feature for shared configs). That will help. However, the team wants to migrate the history over to the new repos, so now we will have multiple repos with huge history instead of just one.

We were hoping that we could improve performance by configuring the config server to do a shallow clone of the repos (e.g. `git clone --depth 1`) . There doesn't appear to be an option for that right now. 

**The Request:** I'd like to request you to enhance the config server to allow for optionally doing a shallow clone.

I realize this might not be feasible. I'm nearly a novice with git, and not sure if such a shallow clone would "play nicely" with the existing config server repo wrt to updates, etc. Also, it appears that the jgit library itself does not expose an option for doing a shallow sync (though I could have missed something).

If not feasible, or not something that might be implemented in the next couple of releases, is there an existing alternative option that might help improve our config server's response time? (Even if this feature does get into the queue, is there something I can do right now to improve performance?)

Note: We're currently at Greenwich.SR3.

pukkaone:
Do not download the entire commit history from the remote.  Execute the equivalent of `git clone --depth 1`

Fixes #1544

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
@@ -87,6 +87,7 @@ import static org.eclipse.jgit.transport.ReceiveCommand.Type.DELETE;
  * @author Ryan Lynch
  * @author Gareth Clay
  * @author ChaoDong Xi
+ * @author Chin Huang
  */
 public class JGitEnvironmentRepository extends AbstractScmEnvironmentRepository
 		implements EnvironmentRepository, SearchPathLocator, InitializingBean {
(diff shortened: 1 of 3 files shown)


spencergibb (changes requested):
Unfortunately we can't make a change like this that changes the default existing behavior for everyone in a minor.

spencergibb:
Maybe if you make it opt in with a configuration flag

pukkaone:
I added a configuration flag where the default is the existing behavior.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/JGitEnvironmentRepository.java
```diff
@@ -773,6 +777,9 @@ public Git getGitByOpen(File file) throws IOException {
 
 		public CloneCommand getCloneCommandByCloneRepository() {
 			CloneCommand command = Git.cloneRepository().setCloneSubmodules(cloneSubmodules);
+			if (cloneShallow) {
+				command.setDepth(1);
```
henri-tremblay, line 781: why not allowing to specify a depth instead?

Issue #2407, opened 2024-04-22
[Refactoring] Introduce RequestContext to easily pass new parameters to downstream methods


opeco17:
**Is your feature request related to a problem? Please describe.**
In the current implementation, it's necessary to change a bunch of method signatures when we introduce new HTTP parameters and pass them to downstream methods.

Let's say you introduce the new parameter `forceRefresh` like #2402.
In this case, you need to change the signature of `EnvironmentRepository` and `SearchPathLocator` and modify a lot of classes that implement those interfaces.

As Is
```java
public interface EnvironmentRepository {
	Environment findOne(String application, String profile, String label);
        Environment findOne(String application, String profile, String label, boolean includeOrigin);
}

public interface SearchPathLocator {
	Locations getLocations(String application, String profile, String label);
}
```

To Be
```java
public interface EnvironmentRepository {
	Environment findOne(String application, String profile, String label);
        Environment findOne(String application, String profile, String label, boolean includeOrigin);
        Environment findOne(String application, String profile, String label, boolean includeOrigin, boolean forceRefresh);
}

public interface SearchPathLocator {
	Locations getLocations(String application, String profile, String label);
        Locations getLocations(String application, String profile, String label, boolean forceRefresh);
}
```

**Describe the solution you'd like**
Introduce `RequestContext` and change the interface of `EnvironmentRepository` and `SearchPathLocator` like below.

```java
public interface EnvironmentRepository {
	Environment findOne(RequestContext ctx);
}

public interface SearchPathLocator {
	Locations getLocations(RequestContext ctx);
}
```

When you introduce new HTTP parameters, you can just change `RequestContext` and don't need to change a bunch of method signatures like the current implementation.

PR #2408, opened 2024-04-22
Refactoring/introduce request context

opeco17:
Fixes https://github.com/spring-cloud/spring-cloud-config/issues/2407


opeco17:
Hi, @ryanjbaxter @spencergibb 

Could you please review this PR?
Thank you!

ryanjbaxter:
I think this is the right idea for where we want to go.

The issue is as is it could only go in a major release because of the breaking API changes.  That is fine but after talking about it some more with the team we should really be deprecating the existing public APIs in the 2024 release (which will be a minor) and then remove them in 2025 release (which will likely be our next major).  Its more of a process thing than anything else, but this way we could also get #2402 in the 2024 release as well.

opeco17:
@ryanjbaxter 

Thank you for sharing.
It would be nice if #2402 becomes available in 2024 release.

May I know the next action that I should take?
Is some revise to this PR necessary?

ryanjbaxter:
Yes we would need to keep the existing public APIs intact and unchanged and mark them all as `@Deprecated` then introduce the new ones (as you already have in this PR) which use the `RequestContext` object.

PR #2428, opened 2024-06-02
Remove AWS SDK v1 Dependency since it is not needed anymore for testing

MatejNedic:
Hey, when checking issue regarding new version of Spring Cloud AWS I noticed AWS SDKV1 is used.
As of new version of Testcontainers and Localstack SDKv1 is not needed anymore so making dependency upgrade with removal.


ryanjbaxter:
Fixed this in https://github.com/spring-cloud/spring-cloud-config/commit/e464808fa395dd3242e9912c744991ca54d57fad

PR #2447, opened 2024-08-09
Handles multiple application names in CredhubEnvironmentRepository

kvmw:
The PR updates `CredhubEnvironmentRepository` to handle following cases:

1.  application name can be a comma separated string like `foo,bar,baz`. In this case, It should be spilted and each entry should be treated as app name. So the repository should search for `foo/profile/label`, `bar/profile/label`, `baz/profile/label` and finally `application/profile/label`. This was handled in other repositories like JGit or Valut but not in CredHub.

2. If the default application name (`application`)  is included in that comma separated string (like for example: `foo,application,bar`) it should be removed and added back to the end of the list. To make sure the default property source is added to the end of the list.  Again other repositories are taking care of this but not CredHub.

3. same for default profile (`default`). If it's included in the profile string (`profile1,default,profile2`), It should be removed and added back to the end of the list. Again other repositories are taking care of this but not CredHub.

Also, In current implementation, If there are no secrets at all in CredHub an empty `PropertySource` will be added to the environment ( `envrionment.propertySources` will have one `PropertySource` which has no properties) . But this PR skips that and `envrionment.propertySources` will be empty. If this is not desired, I can add that empty `ProperySource` back.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.config.server.environment;
 
 import java.util.Arrays;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.springframework.cloud.config.environment.Environment;
 import org.springframework.cloud.config.environment.PropertySource;
@@ -37,8 +38,6 @@ import static java.util.stream.Collectors.toMap;
  */
 public class CredhubEnvironmentRepository implements EnvironmentRepository, Ordered {
 
-	private CredHubOperations credHubOperations;
-
 	private static final String DEFAULT_PROFILE = "default";
 
 	private static final String DEFAULT_LABEL = "master";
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
Thanks!  Could you submit this against the 4.1.x branch and we can merge it forward?

@spencergibb what do you think about this change?
> Also, In current implementation, If there are no secrets at all in CredHub an empty PropertySource will be added to the environment ( envrionment.propertySources will have one PropertySource which has no properties) . But this PR skips that and envrionment.propertySources will be empty. If this is not desired, I can add that empty ProperySource back.

That could be considered breaking, so I am not sure we want to make that change in 4.1.x.  I guess we could do it in main.

kvmw:
> > Also, In current implementation, If there are no secrets at all in CredHub an empty PropertySource will be added to the environment ( envrionment.propertySources will have one PropertySource which has no properties) . But this PR skips that and envrionment.propertySources will be empty. If this is not desired, I can add that empty ProperySource back.
> 
> That could be considered breaking, so I am not sure we want to make that change in 4.1.x. I guess we could do it in main.

The second commit adds the main PropertySource, even if it is empty. The only caveat is the name of the PropertySource, in case application has more than one name.  For example, given `appOne,appTwo` as application name in the current code, PropertySource name would be `credhub-appOne,appTwo-myProfile-myLabel` but the new code changes it to `credhub-appOne-myProfile-myLabel`.

ryanjbaxter:
Why change the property source name?

kvmw:
> Why change the property source name?

PropertySource name has the following format: `credhub-<APPLICATION-NAME>-<PROFILE>-<LABEL>`.  
Until now, because of this bug which I am trying to fix, for a given application name like `appOne,appTwo` we would have only one PropertySource which was named `credhub-appOne,appTwo-default-master` . But with the new changes we will have 2 property sources with the following names: `credhub-appOne-default-master` and `credhub-appTwo-default-master`. Assuming they are both empty, we just need to keep the first one.

ryanjbaxter:
Is this how other EnvironmentRepositories handle this situation as well?

kvmw:
If you mean splitting the application name, yes.  In other repos application name is converted to an array and each individual item in array makes a separate property source. 

If you are referring to empty property sources, no. not all of them. For example, [Vault](https://github.com/spring-cloud/spring-cloud-config/blob/4654065718b8ba57204374ea69aced684752b526/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AbstractVaultEnvironmentRepository.java#L97) does not create empty property source.

kvmw:
> Yes I was referring to the name, I just want to be as consistent as possible.

JGit behaves the same. 

a sample config like this:
```yaml
spring:
  profiles:
    active: development,production
  application:
    name: cook,chef
```
will have the following property sources :
```
"configserver:https://github.com/cook-config/chef-production.properties"
"configserver:https://github.com/cook-config/cook-production.properties"
"configserver:https://github.com/cook-config/application-production.properties"
"configserver:https://github.com/cook-config/chef-development.properties"
"configserver:https://github.com/cook-config/cook-development.properties"
"configserver:https://github.com/cook-config/application-development.properties"
"configserver:https://github.com/cook-config/chef.properties"
"configserver:https://github.com/cook-config/cook.properties"
"configserver:https://github.com/cook-config/application.properties"
```

Issue #2449, opened 2024-08-10
Multiple labels with JDBC not working


ats1999:
My spring cloud config server is using JDBC as backend. My requirement is to use multiple labels to fetch config. 

When fetching data using `http://localhost:8888/test_app,test_app2/dev,prod/master` URL i am getting the following response 
```json
{
  "name": "test_app,test_app2",
  "profiles": [
    "dev,prod"
  ],
  "label": "master",
  "version": null,
  "state": null,
  "propertySources": [
    {
      "name": "test_app2-prod",
      "source": {
        "name_dev": "rahul_dev9999"
      }
    },
    {
      "name": "test_app-prod",
      "source": {
        "name": "rahul"
      }
    },
    {
      "name": "test_app2-dev",
      "source": {
        "name_dev": "rahul_dev9999"
      }
    },
    {
      "name": "test_app-dev",
      "source": {
        "name_dev": "rahul_dev"
      }
    }
  ]
}
```

But getting empty response after adding another label, `http://localhost:8888/test_app,test_app2/dev,prod/master,feature`

```json
{
  "name": "test_app,test_app2",
  "profiles": [
    "dev,prod"
  ],
  "label": "master,feature",
  "version": null,
  "state": null,
  "propertySources": []
}
```

Is this supported by spring cloud config? 

> It is required when working on feature branch, i want all configs to be pulled from master label while certain config from feature label.

ryanjbaxter:
I believe it should be supported.  Have you tried it using the native file system as a back end to see if that supports multiple labels?

ats1999:
Will check and let you know

ats1999:
Not working with file system backend as well 

Spring cloud config server application.properties
```text
spring.profiles.active=native
spring.cloud.config.server.native.searchLocations=/Users/ats1999/Desktop/projects/w3worker/config-service/config-server/src/main/resources/config
```

`/Users/ats1999/Desktop/projects/w3worker/config-service/config-server/src/main/resources/config/auth-dev-master.properties`
```text
auth=auth.dev.master
```
Upon querying http://localhost:8888/auth/dev/master, i am getting empty output. 

But once i removed master from `auth-dev-master.properties`, i am getting the desired properties with or without label in `http://localhost:8888/auth/dev/master`

Is this desired behaviour?

ryanjbaxter:
No I don't think so.  When I try this using a native backend I am getting a 500 back from the config server, is that not what you are seeing?

[...]

PR #2454, opened 2024-08-23
Support multiple labels in NativeEnvironmentRepository

ryanjbaxter:
Related to #2449

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/NativeEnvironmentRepository.java
@@ -210,9 +210,15 @@ public class NativeEnvironmentRepository implements EnvironmentRepository, Searc
 		if (this.addLabelLocations) {
 			for (String location : locations) {
 				if (StringUtils.hasText(label)) {
-					String labelled = location + label.trim() + "/";
-					if (isDirectory(labelled)) {
-						output.add(labelled);
+					String[] labels = { label };
+					if (labels[0].contains(",")) {
+						labels = labels[0].split(",");
+					}
+					for (String l : labels) {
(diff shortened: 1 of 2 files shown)


spencergibb:
Maybe for main?

kvmw:
@spencergibb 

Is the change in this PR is just for this particular repository or it supposed to be implemented by all other repositories?  I am thinking about JGitEnvironmentRepository and other similar repos. I don't think supporting multiple labels would be easy for JGitEnvironmentRepository
.

ryanjbaxter:
I am curious why you think it would not be easy for JGitEnvironmentRepository?

ryanjbaxter:
I guess I see why now since a label corresponds to a branch and we can only checkout and use a single branch at a time

kvmw:
yes, And It uses NativeEnvironmentRepository via [AbstractScmEnvironmentRepository](https://github.com/spring-cloud/spring-cloud-config/blob/d99423830a796b7c13c9d2d764f8c88fd2925095/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AbstractScmEnvironmentRepository.java#L59) internally. 

So I supposed we need to split the label and for each entry we have to run what is currently in `AbstractScmEnvironmentRepository::findOne`. 

Something like: 
```java

var labels = label.split(",");
for (String lbl : labels) {
                NativeEnvironmentRepository delegate = new NativeEnvironmentRepository(getEnvironment(),
				new NativeEnvironmentProperties(), this.observationRegistry);
		Locations locations = getLocations(application, profile, lbl);
		delegate.setSearchLocations(locations.getLocations());
		Environment result = delegate.findOne(application, profile, "", includeOrigin);
		result.setVersion(locations.getVersion());
		result.setLabel(lbl);
}
```

I am just worried about the impact of such a change. Maybe I am just being oversuspicious.

Issue #2439, opened 2024-07-10
Native support for RetryTemplateFactory as it uses reflections to access the field logger of RetryTemplate 


klopfdreh:
**Describe the bug**
We wanted to use Spring Cloud Config Client within a native image. During the startup of our application we saw a `NullPointerException`

```plain
java.lang.NullPointerException
at org.springframework.cloud.config.client.RetryTemplateFactory.create(RetryTemplateFactory.java:47)
at org.springframework.cloud.config.client.ConfigClientRetryBootstrapper.lambda$initialize$1(ConfigClientRetryBootstrapper.java:46)
at org.springframework.cloud.config.client.ConfigServerConfigDataLoader.load(ConfigServerConfigDataLoader.java:92)
at org.springframework.cloud.config.client.ConfigServerConfigDataLoader.load(ConfigServerConfigDataLoader.java:61)
at org.springframework.boot.context.config.ConfigDataLoaders.load(ConfigDataLoaders.java:96)
```

We had a look into `RetryTemplateFactory` and saw `field = ReflectionUtils.findField(RetryTemplate.class, "logger");` in line 31 and `field.set(retryTemplate, log);` in line 47. 

There is no reflect-config.json provided to describe this access. 

**Sample**
N/A

AlexGreg:
same for me..I solved it by adding to `reflect-config.json`

```
[
  {
    "name": "org.springframework.retry.support.RetryTemplate",
    "allDeclaredFields": true,
    "allDeclaredMethods": true,
    "allDeclaredConstructors": true
  },
  {
    "name": "org.springframework.cloud.config.client.RetryTemplateFactory",
    "allDeclaredMethods": true,
    "allDeclaredConstructors": true
  }
]
```

klopfdreh:
Hey @AlexGreg

thanks that you also commented this issue!

Exactly this needs to be provided by spring-cloud-config-client 👍

klopfdreh:
@ryanjbaxter - should we do it this way or should I provide a `RuntimeHintsRegistrar` and apply the hints that way?

klopfdreh:
@ryanjbaxter - I made a suggestion as this is a Spring Project I guess we are safe to use a RuntimeHintsRegistrar. See the PR for more information, please.

klopfdreh:
[...]

PR #2456, opened 2024-09-02
feat: native support for retry template factory

klopfdreh:
Fixes: https://github.com/spring-cloud/spring-cloud-config/issues/2439

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2013-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.config.client.aot;
+
+import org.springframework.aot.hint.MemberCategory;
+import org.springframework.aot.hint.RuntimeHints;
+import org.springframework.aot.hint.RuntimeHintsRegistrar;
+import org.springframework.aot.hint.TypeReference;
+import org.springframework.boot.context.config.ConfigDataLocation;
+import org.springframework.cloud.config.client.ConfigClientAutoConfiguration;
+import org.springframework.cloud.config.client.RetryTemplateFactory;
+import org.springframework.cloud.config.environment.Environment;
+import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.util.ClassUtils;
+
+/**
+ * The config client runtime hints to enable the client to be used for aot builds.
+ *
+ * @author Olga Maciaszek-Sharma
+ * @author Tobias Soloschenko
+ */
+class ConfigClientHints implements RuntimeHintsRegistrar {
+
+	@Override
+	public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
+		if (!ClassUtils.isPresent("org.springframework.cloud.config.client.ConfigServerConfigDataLoader",
+			classLoader)) {
+			return;
+		}
+		hints.reflection()
+			.registerType(TypeReference.of(ConfigClientAutoConfiguration.class),
+				hint -> hint.withMembers(
+					MemberCategory.INVOKE_DECLARED_CONSTRUCTORS)
+			).registerType(TypeReference.of(ConfigDataLocation.class),
+				hint -> hint.withMembers(
+					MemberCategory.INVOKE_DECLARED_METHODS)
+			).registerType(TypeReference.of("org.springframework.boot.context.config.ConfigDataProperties"),
+				hint -> hint.withMembers(
+					MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
+					MemberCategory.DECLARED_FIELDS,
+					MemberCategory.INTROSPECT_DECLARED_METHODS)
+			).registerType(TypeReference.of(Environment.class),
+				hint -> hint.withMembers(
+					MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
+					MemberCategory.INTROSPECT_DECLARED_METHODS,
+					MemberCategory.DECLARED_FIELDS)
+			).registerType(TypeReference.of(PropertySource.class),
+				hint -> hint.withMembers(
+					MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
+					MemberCategory.INTROSPECT_DECLARED_METHODS,
+					MemberCategory.DECLARED_FIELDS)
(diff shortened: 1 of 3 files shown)


klopfdreh:
I just saw that the previously mentioned `org.springframework.cloud.config.client.ConfigClientHints` is not present in the source code. Maybe it should be removed form aot.factories?

klopfdreh:
I used the package name `aot` like in spring batch: https://github.com/spring-projects/spring-batch/tree/main/spring-batch-core/src/main/java/org/springframework/batch/core/aot

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.retry.support.RetryTemplate;
+import org.springframework.util.ClassUtils;
+
+/**
+ * The config client runtime hints to enable the client to be used for aot builds.
+ *
+ * @author Olga Maciaszek-Sharma
+ * @author Tobias Soloschenko
+ */
+public class ConfigClientHints implements RuntimeHintsRegistrar {
```
OlgaMaciaszek, line 21: Please test if it has to be `public` and, if not, change to package access.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.aot.hint.MemberCategory;
+import org.springframework.aot.hint.RuntimeHints;
+import org.springframework.aot.hint.RuntimeHintsRegistrar;
+import org.springframework.aot.hint.TypeReference;
+import org.springframework.boot.context.config.ConfigDataLocation;
+import org.springframework.cloud.config.client.ConfigClientAutoConfiguration;
+import org.springframework.cloud.config.client.RetryTemplateFactory;
+import org.springframework.cloud.config.environment.Environment;
+import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.retry.support.RetryTemplate;
```
OlgaMaciaszek, line 12: This is actually from a different Spring project/repo, so optimally the hint should be added there; we can keep it here till it's done, but either add a separate PR there and link here or let me know and I'll get it done.

OlgaMaciaszek (changes requested):
Thanks @klopfdreh, looks good - but have added some minor comments. Also, please  submit your changes against `4.1.x` instead of `main`, so that we can get them into the `2024.0.x` release train as well.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
klopfdreh: Changed the access to package access

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
klopfdreh: I am going to create a PR at Spring Retry.

klopfdreh:
I force pushed the change so that they are done against `4.1.x`

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
klopfdreh: Just for my understanding. Is best practice to define the hint in the project which defines the class? I asked because to me it looks a bit odd, because Spring Retry does not use RetryTemplate with any kind of reflections, but Spring Cloud Config client does. That's the reason I added the hints here.

artembilan:
Hi, @OlgaMaciaszek !

Would you mind explaining me why do we need to change that `RetryTemplate.logger` property at all?
If there is a good reason it might be much better to have access to the property via setter exposed on the `RetryTemplate` instead of reflection.
Thanks

artembilan:
I see `ConfigServicePropertySourceLocator` propagates its own `static Log` without any external modification: https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java#L125
Same happens with a `ConfigServerConfigDataLocationResolver`: https://github.com/spring-cloud/spring-cloud-config/blob/main/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java#L69.
So, why do we need to propagate those non-configurable loggers?
Why regular one in `RetryTemplate` (and anywhere else) is not enough?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
OlgaMaciaszek: It's best for them to be available for any other project using that class.

OlgaMaciaszek:
Hi @artembilan, when it comes to Config, I mainly take care of any AOT/ native integrationn issues; @ryanjbaxter or @spencergibb may have more details regarding that implementation, so let's check with them on that.

ryanjbaxter:
IMO the change should be made here https://github.com/spring-projects/spring-retry/issues/470 if the issue lies in Spring Retry.

klopfdreh:
@ryanjbaxter there are two classes require reflections: RetryTemplateFactory which is located in Spring Cloud Config and RetryTemplate which is located in Spring Retry.

I could modify this PR to only cover the class located in Spring Cloud Config.

spencergibb:
The first one is in bootstrap before boot configures loggers, the second is for config data and its special deferred logger. I think both are attempts at same logging in an early phase of cloud or boot respectively

ryanjbaxter:
@klopfdreh agree, I think that only the classes in Spring Cloud Config should be configured here and configure the `RetryTemplate` in Spring Retry

artembilan:
Thanks, @ryanjbaxter and @spencergibb , for explanation!
So, the setter I suggest in Spring Retry would be OK?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2013-2022 the original author or authors.
```
OlgaMaciaszek, line 2: Please change the dates to `2013-2024`

OlgaMaciaszek (changes requested):
Thanks @klopfdreh. Looks good. Just a cosmetic change to add.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/aot/ConfigClientHints.java
klopfdreh: done

OlgaMaciaszek (approved):
LGTM.

klopfdreh:
Thanks a lot for all the help @OlgaMaciaszek / @ryanjbaxter / @spencergibb / @artembilan - we ported 11 of our task applications to Spring Boot Native and I made a lot of changes to various frameworks like AWS SDK V2 for Java / AWS CRT for Java / joda-time / github-api / spot bugs / Snappy Java / Micrometer Prometheus  to fix issues regarding native images. When the new version of Spring Cloud Data Flow is released we are starting with tests so it. It is so great that you helped us out here! 👍

OlgaMaciaszek:
Thanks @klopfdreh. Sounds great! Don't hesitate to reach out if there's any other issue.

PR #2519, opened 2024-09-06
Makes default-label in CredhubEnvironmentRepository configurable

kvmw:
the default-label in `credhub` backend should be configurable, similar to `git`, `jdbc` and other backends.


ryanjbaxter:
Can we add a note to the documentation about the property?

kvmw:
Sure, 
Just to confirm, the docs are the one in this repo under `/docs` folder.

kvmw:
The second commits updates docs to include default label configuration.

PR #2561, opened 2024-09-26
Initial implementation to support multiple labels in SVN and Git EnvironmentRepos

ryanjbaxter:


ryanjbaxter:
@kvmw here is an initial implementation for SCM repositories.  Let me know what you think

kvmw:
I've tried this locally and it is not returning back the property sources I was expecting.  

I believe the `findOne` method should be called for each individual repo, rather than once for the composite repo. Because locations have overlap in different branches, with the current approach, we are able to search the locations of last `git fetch`.

ryanjbaxter:
@kvmw you are right.  It works fine for SVN because it uses different directories for branches but git does not.  I removed the use of composite, let me know if that works as you expect

kvmw:
@ryanjbaxter  Yes, it is working as expected with my sample configs.

PR #2581, opened 2024-10-09
Default-label in CredhubEnviornmentRepository should be used only if label is not provided

kvmw:
Unlike default application name and profile, default label should only be used  if provided label is `null` or `empty`. 

I've added a second variation of `normalize` method  which accepts only comma-delimited string without default item.


kvmw:
@ryanjbaxter  This was a regression by #2568

ryanjbaxter (commented):
Makes sense, thanks!

PR #2613, opened 2024-10-30
Only ignore profile activation and inclusion if the property source is profile specific

ryanjbaxter:
If we set the flag on all property sources (like we do currently) then any profiles that should be activated via `spring.profiles.active` or `spring.profiles.include` coming from the config server will not be activated in the application.


spencergibb:
In practice, how does this affect the client app?

ryanjbaxter:
What ends up happening is that if any of the property sources that come from the config server contains `spring.profiles.active` or `spring.profiles.include` none of those profiles will be activated in the client.  So if you went to `/acturator/env` you wouldn't see any of those profiles active.  

This could result in 2 bugs and I am sure there would be more....

1.  Configuration classes or beans that would only be active if a certain profiles is active would not be configured.
2. If you had config data sources outside of the config server that should be activated with profiles being activated from the config server those would not be activated.

spencergibb:
So should this go only in main then for the behavior change, or does it fix bugs that should go in 4.1.x?

ryanjbaxter:
It would be a change in behavior but IMO its a change in behavior that is fixing a bug.  Maybe we should talk with the rest of the team to see what they think?

spencergibb:
yup, we can chat tomorrow

PR #2623, opened 2024-11-06
Fixes: ConfigClientProperties#uri can't be set correctly as the ConfigClientProperties in BootstrapContext will be replaced  

519674304:
Every time org.springframework.cloud.config.client.ConfigServerConfigDataLocationResolver#resolveProfileSpecific be invoked, it will create a new ConfigClientProperties and replace the ConfigClientProperties in the BootstrapContext, but the ConfigClientProperties in the ConfigServerInstanceMonitor will be only set in the first time invoke.
Though, org.springframework.cloud.config.client.ConfigServerConfigDataLocationResolver#loadProperties will replace the org.springframework.cloud.config.client.ConfigClientProperties#uri if spring.cloud.config.discovery.enabled equals true, but the uri is get from the ConfigClientProperties from BootstrapContext which was replaced.

### DIFF
--- a/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
+++ b/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
@@ -113,11 +113,10 @@ public class ConfigServerConfigDataLocationResolver
 		Binder binder = context.getBinder();
 		BindHandler bindHandler = getBindHandler(context);
 
-		ConfigClientProperties configClientProperties;
-		if (context.getBootstrapContext().isRegistered(ConfigClientProperties.class)) {
-			configClientProperties = binder
+		ConfigClientProperties configClientProperties = binder
 				.bind(ConfigClientProperties.PREFIX, Bindable.of(ConfigClientProperties.class), bindHandler)
 				.orElseGet(ConfigClientProperties::new);
+		if (context.getBootstrapContext().isRegistered(ConfigServerInstanceMonitor.class)) {
 			boolean discoveryEnabled = context.getBinder()
 				.bind(CONFIG_DISCOVERY_ENABLED, Bindable.of(Boolean.class), getBindHandler(context))
 				.orElse(false);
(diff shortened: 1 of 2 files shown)


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
```diff
@@ -126,20 +125,16 @@ protected PropertyHolder loadProperties(ConfigDataLocationResolverContext contex
...
 			// ConfigServerInstanceMonitor.refresh which will only
 			// be called the first time we fetch configuration.
+			// So here we should get the ConfigClientProperties from ConfigServerInstanceMonitor
 			if (discoveryEnabled) {
-				ConfigClientProperties bootstrapConfigClientProperties = context.getBootstrapContext()
-					.get(ConfigClientProperties.class);
-
+				ConfigServerInstanceMonitor configServerInstanceMonitor = context.getBootstrapContext()
+						.get(ConfigServerInstanceMonitor.class);
+				ConfigClientProperties bootstrapConfigClientProperties = configServerInstanceMonitor.getConfig();
```
ryanjbaxter, line 132: I assume the ConfigClientProperties here is the same as the one in the BootstrapContext?  The comment above indicates it should come from there.

ryanjbaxter:
This PR should be made against the 4.1.x branch.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
519674304: That's different, because once ConfigClientProperties first fetch from BootstrapContext then set in the new ConfigClientProperties#uri then the new ConfigClientProperties will replace the old ConfigClientProperties in BootstrapContext. Next time when you invoke RefreshEndpoint for refresh envirment. ConfigServerConfigDataLocationResolver#resolveProfileSpecific, you will fetch the ConfigClientProperties from BootstrapContext but this bean is not the ConfigClientProperties in ConfigServerInstanceMonitor so the ConfigClientProperties in BootstrapContext will never be affected by ConfigServerInstanceMonitor so ConfigClientProperties.uri is wrong when eureka find some instance of cloud-config-server is down and refresh the ConfigClientProperties.uri

519674304:
> This PR should be made against the 4.1.x branch.

Thanks, I will correct it.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
ryanjbaxter: I understand all of that, but I want to make sure the behavior before the uri is updated is maintained.

I want to confirm that `configServerInstanceMonitor.getConfig()` returns the same instance as `context.getBootstrapContext().get(ConfigClientProperties.class)` currently does.

519674304: image
image

519674304: > I understand all of that, but I want to make sure the behavior before the uri is updated is maintained.
> 
> I want to confirm that `configServerInstanceMonitor.getConfig()` returns the same instance as `context.getBootstrapContext().get(ConfigClientProperties.class)` currently does.

I think you need make some change in this line, the detail code I need read more code about Binder.
image

ryanjbaxter: Please use github to add comment and do not send images it makes it hard to keep track of everything....

So you are saying this PR likely needs more changes then?

519674304: because I see you want to keep the ConfigClientProperties in configServerInstanceMonitor and BootstrapContext is the same object, so I think ConfigServerConfigDataLocationResolver#resolveProfileSpecific shoud instead of create a new ConfigClientProperties but get ConfigClientProperties from BootstrapContext if it is exist.
I need more time to implement that.

ryanjbaxter:
Closing in favor of https://github.com/spring-cloud/spring-cloud-config/pull/2630

PR #1379, opened 2019-05-11
support multiple spring application names for Vault Fixes gh-1356

Haybu:
Fixes gh-1356

Issue #1356, opened 2019-04-18
Multiple spring application name for Vault


Viimmo:
My spring cloud config client has multiple spring.application.name to get differents property files with GIT but Vault is not working like GIT.

**application.properties of my server** :
```
spring.config.name=configserver
server.port=8188

spring.profiles.active=git,vault
 
spring.cloud.config.server.git.uri=https://.../config.git
spring.cloud.config.server.git.username=config
spring.cloud.config.server.git.password=*************
spring.cloud.config.server.git.clone-on-start=true
 
management.endpoints.web.exposure.include=*

spring.cloud.config.server.vault.port=8200
spring.cloud.config.server.vault.host=120.0.0.1
spring.cloud.config.server.vault.profileSeparator=/
```
**bootstrap.properties of my client** :
```
spring.application.name=application1,application2
spring.profiles.active=next
spring.cloud.config.uri=http://localhost:8188

spring.cloud.config.token=********
```

With this configuration, I am able to load with **GIT** my 2 property files application1-next.yml and application2-next.yml.

Sadly **Vault** is not working like I was expected : load values from the 2 paths
> application1/next/
> application2/next

It is looking for keys in the single path 
> application1,application2/next

Does anyone have a solution about this ? or maybe I miss something...

PR #2664, opened 2024-12-05
Support multiple Spring application names for AWS Secrets Manager

2012160085:
### Support multiple spring application names for AWS Secret Manager

Spring Cloud Config supports multiple application names in backends like Git and Vault, but this feature seems to be missing in AWS Secret Manager. This change enables AWS Secret Manager to handle comma-separated application names, aligning with these backends. 

This implementation is similar to PR [#1379](https://github.com/spring-cloud/spring-cloud-config/pull/1379), which added support for multiple applications in the Vault backend, and resolves the limitations discussed in [issue #1356](https://github.com/spring-cloud/spring-cloud-config/issues/1356).

#### What’s Changed
- Enabled processing of comma-separated application names for AWS Secret Manager, allowing it to handle multiple applications like Git and Vault backends.

#### Why This Matters
- Teams migrating from Git or Vault to AWS Secret Manager can now use the same configuration structure.

#### Testing
- Added test cases for:
  - Non-existing profile and label.
  - Default profile with an existing label.
  - Existing profile with null label.

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentRepository.java
@@ -112,24 +112,26 @@ public class AwsSecretsManagerEnvironmentRepository implements EnvironmentReposi
 
 		List<String> reversedProfiles = new ArrayList<>(Arrays.asList(profiles));
 		Collections.reverse(reversedProfiles);
+		if (!reversedProfiles.contains(defaultProfile)) {
+			reversedProfiles.add(defaultProfile);
+		}
+
+		List<String> applications = Arrays.asList(StringUtils.commaDelimitedListToStringArray(application));
+		Collections.reverse(applications);
+		if (!applications.contains(defaultApplication)) {
+			applications = new ArrayList<>(applications);
+			applications.add(defaultApplication);
+		}
+
 		for (String l : labels) {
 			for (String profile : reversedProfiles) {
-				addPropertySource(environment, application, profile, l);
-				if (!defaultApplication.equals(application)) {
-					addPropertySource(environment, defaultApplication, profile, l);
+				for (String app : applications) {
+					addPropertySource(environment, app, profile, l);
 				}
 			}
-			if (!Arrays.asList(profiles).contains(defaultProfile)) {
-				addPropertySource(environment, application, defaultProfile, l);
+			for (String app : applications) {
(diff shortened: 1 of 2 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentRepository.java
```diff
@@ -112,24 +112,26 @@ else if (StringUtils.isEmpty(label)) {
...
 			for (String profile : reversedProfiles) {
-				addPropertySource(environment, application, profile, l);
-				if (!defaultApplication.equals(application)) {
-					addPropertySource(environment, defaultApplication, profile, l);
+				for (String app : applications) {
+					addPropertySource(environment, app, profile, l);
 				}
 			}
-			if (!Arrays.asList(profiles).contains(defaultProfile)) {
-				addPropertySource(environment, application, defaultProfile, l);
```
ryanjbaxter, line 123: Are there existing tests that verify that changing this logic doesn't break anything?  I am nervous that we might be changing the order of certain property sources and breaking expectations.

2012160085: [This one](https://github.com/spring-cloud/spring-cloud-config/blob/f38589ba69c28205d405d2f3bc0f20dd6ac9a600/spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/AwsSecretsManagerEnvironmentRepositoryTests.java#L781C1-L825C3) appears to provide good coverage for the logic changes in AWS Secrets Manager, particularly regarding the ordering and inclusion of profiles.

It ensure that:
- Profiles specified as a comma-separated list (like `"prod,east"`) are handled in the correct order.
- `null` profiles are properly included.
- The default profile is correctly added, and its position relative to other profiles (including `null` profiles) is maintained.

Based on my review, this test, along with `testFindOneWithNullApplicationAndNonExistingProfileAndNullLabelWhenDefaultLabelIsSet` and `testFindOneWithExistingApplicationAndNonExistingProfileAndNoDefaultProfileForFooAndNullLabelWhenDefaultLabelIsSet`, seems to cover the existing logic quite well.

2012160085:
I added a comprehensive test case to thoroughly ensure that the logic for retrieving secrets remains consistent with the previous implementation.

The test covers scenarios with multi-profiles and multi-labels, including default profiles and null profiles, verifying the exact order of secrets in all cases.

PR #2679, opened 2024-12-13
Remove unused code branch In GitCredentialsProviderFactory

### LINKED ISSUE #2674: Possible unused branch in GitCredentialsProviderFactory
**Describe the bug**
It seems the branch starting on line 99 is never executed because whenever `hasText(username)` is true, the branch starting on line 93 is executed.

https://github.com/spring-cloud/spring-cloud-config/blob/e7a69671b3a9242e2412e030e8660af4a194f56c/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/GitCredentialsProviderFactory.java#L93-L106

ryanjbaxter:
Fixes #2674

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/GitCredentialsProviderFactory.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/GitCredentialsProviderFactory.java
@@ -88,16 +88,9 @@ public class GitCredentialsProviderFactory {
 			aws.setPassword(password);
 			provider = aws;
 		}
-		else if (hasText(username)) {
+		else if (hasText(username) && password != null) {
 			this.logger.debug("Constructing UsernamePasswordCredentialsProvider for URI " + uri);
 			provider = new UsernamePasswordCredentialsProvider(username, password.toCharArray());
-		}
-		else if (hasText(username) && !hasText(passphrase)) {
-			// useful for token based login gh-1602
-			// see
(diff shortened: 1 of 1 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/support/GitCredentialsProviderFactory.java
```diff
@@ -92,13 +92,6 @@ else if (hasText(username)) {
 			this.logger.debug("Constructing UsernamePasswordCredentialsProvider for URI " + uri);
 			provider = new UsernamePasswordCredentialsProvider(username, password.toCharArray());
 		}
-		else if (hasText(username) && !hasText(passphrase)) {
```
spencergibb, line 95: Maybe it shouldn't be removed and add `&& hasText(password)` to the branch above? That will fix a potential NPE there anyway.

ryanjbaxter: @spencergibb This logic is tasting if _passphrase_ has text not password.  I agree that we should add a null check for `password` above.  But we probably don't want to use a `hasText` check since having an empty string for password is valid (we have a test for that).

spencergibb: right, null check works

spencergibb (approved):

Issue #2735, opened 2025-02-03
Next URL is not attempted if not successful, but exception is not thrown


marnee01:
**Describe the bug**
See this code in ConfigServicePropertySourceLocator.getRemoteEnvironment method:

```
			if (response == null || response.getStatusCode() != HttpStatus.OK) {
				return null;
			}
```

We'd like to have this updated so that when this condition is met, it tries the next URL (if any).
We'd also like to have this log the status code at ERROR level when response is not null. And if response is null, log that fact at ERROR level.

**Sample**
It's difficult to provide a working example. We have a strange situation with our environment that is causing this condition to evaluate to true, but it is related to an Nginx issue, and I would not know how to replicate this in a standalone project.

ryanjbaxter:
Just so I am clear you woud like something like this?

```
if (response == null || response.getStatusCode() != HttpStatus.OK) {
				if(response == null) {
					logger.warn("The response from " + uri + " was null, trying the next uri if available");
				}
				if (response.getStatusCode() != HttpStatus.OK) {
					logger.warn("The response from " + uri + " had a status code of " + response.getStatusCode() ". Trying the next uri if available");
				}
				continue;
			}
```
I don't think we need to log it as an error, I think this should be a warning.

marnee01:
Basically, yes. Maybe it would be good to also have the same log messages that reports that it's trying the next URL, like it does in the catch blocks, for consistency.

As far as error vs warning, I'm a little unclear why there would be a null response or a non-OK response that doesn't generate one of the caught exceptions. I observed this happening, but due to the lack of logging, I am not sure what the response looked like. I couldn't reproduce it locally, so I couldn't step through the running code. If you think this is not an error condition, then a warning seems fine.

ryanjbaxter:
So when looking at making these changes, you actually wrote a test that breaks this assumption and verifies that when that status code is not 200 that it does not try the next uri
[shouldNotUseNextUriFor_404_And_CONNECTION_TIMEOUT_ONLY_Strategy](https://github.com/spring-cloud/spring-cloud-config/commit/7f83b1f547c9fbee0f50fb4d00eb29f97890ca22#diff-392ff2e5abfd8684353b48cc59d1a4022afcca74ea7f8d324b513c5bb9b90bdbR296-R299)

As far as an error or warning, I can see it go both ways.  In the case where there is another uri to try it feels like a warning, but if there is no other uri to try it seems like it could be an error.

[...]

PR #2798, opened 2025-03-21
Config Client: Fix logic to try all URLs

marnee01:
There was a miss when implementing the MultipleUriStrategy.ALWAYS. It was not trying the next URL if response was null, or if the HTTP Status Code was not OK, but no exception was thrown.

(This author does not know under what scenario null would be returned as a response, but the code was already checking for it, so I updated the logic to also try the next URL for null response.)

Closes gh-2735


marnee01:
@ryanjbaxter I had opened this against the latest branch previously, but had wanted to patch it back. You had instructed me to create this against 4.1.x and said to tag you, and that you would merge it forward. Thanks in advance.

PR #2810, opened 2025-04-01
Support server-side authentication for Vault in composite backends

### LINKED ISSUE #2821: Server-side authentication for Vault not supported in composite backends
**Describe the bug**

When there is Vault backend within a composite backend, then it cannot do server-side authentication. For example, it cannot do AppRole-based authentication.

Given following configuration for config-server:

```
spring:
  profiles:
    active: composite
  cloud:
    config:
      server:
        composite:
          - type: vault
            host: my-vault-server
            port: 5555
            authentication: approle
            roleId: 123
            secretId: 456
```

It won't be able to authenticate with Vault using its AppRole's role id and secret id. Instead it throws an error, saying that it expects the `X-Config-Token` header. This tells us that it's using a wrong config token provider.

The reason is that a default session manager bean is being created with the default Vault properties. However, a session manager is needed which uses the Vault properties from the Vault in the composite backend.

Authored w/ @kvmw

mamachanko:
By setting the SessionManager on the VaultTemplate, we can use authentication properties provided by composite backends.

Authored w/ @kvmw 

Fixes #2821

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultTemplateBuilder.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/vault/SpringVaultTemplateBuilder.java
@@ -21,6 +21,7 @@ import java.util.List;
 import org.springframework.cloud.config.server.environment.ConfigTokenProvider;
 import org.springframework.cloud.config.server.environment.VaultEnvironmentProperties;
 import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
Thanks!  Can you submit this against the 4.1.x branch?

mamachanko:
> Thanks! Can you submit this against the 4.1.x branch?

@ryanjbaxter of course! This PR is superseded by https://github.com/spring-cloud/spring-cloud-config/pull/2828.

PR #2855, opened 2025-05-15
Modifing the documentation about generating the jks file with keytool

kstojanovski-novatec:
- The format of the generated jks file is PKCS12 by checking it with 'keytool -list -v -keystore server.jks' and the application fails to start. Adding a parameter for creation of keys in JKS format solved the problem.

- Using bootstrap.yml does not have any impact on the application. Using application.properties gave the expected effect. Changing the description from bootstrap.yml with application.properties.


spencergibb:
Bootstrap is still supported, it's just not the default

PR #2911, opened 2025-06-16
Do not return profile-specific property sources

ryanjbaxter:
When a property source uses spring.profiles.import the config server would return property sources activated by this configuration property even if the config client did not request profile information.  This is due to using Spring Boots ConfigDataEnvironmentPostProcessor to load configuration from the file system (which is not only used for native but also for git and svn).  This change makes it so the config server will not activate profiles if the config client does not request any profile data.

See the corresponding change in Spring Boot https://github.com/spring-projects/spring-boot/issues/42932#event-18116616790

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -16,7 +16,7 @@ jobs:
     # if: github.repository_owner == 'spring-cloud'
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


ryanjbaxter:
See https://github.com/spring-cloud/spring-cloud-config/pull/3023

PR #2958, opened 2025-07-06
Support Git-style searchPaths with wildcards in AWS S3 buckets (#2812)

### LINKED ISSUE #2812: AwsS3EnvironmentRepository does not comply with Git
**Describe the bug**
Using version 3.1.9 due to restrictions, but looking in the code of AwsS3EnvironmentRepository I see the newest version has the same issue. Our configuration is sorted into folders for each application with the name of that given application. The implementation does not handle this like the Git version - tho I see that the newest can use folders, but with hardcoded "application" as the name of the file.

**Sample**
Works in Git, but not in AwsS3EnvironmentRepository :

app1/app1-profile.yml
app1/app1.yml

What looks to work in AwsS3EnvironmentRepository in 4.3.0-SNAPSHOT:

app1/application-profile.yml
app1/application.yml

tomy8964:
### Summary

Add full Git-style `searchPaths` support (placeholders + wildcards) to the AWS S3 backend so that users can migrate existing Git-based configurations without renaming to `application.*`.

* Implements literal, dot-wildcard (`.*`), single (`?`) and double (`**`) wildcard matching
* Retains the lookup order `.properties → .json → .yml → .yaml` when expanding literals or `.*` patterns
* Scans “directory” patterns for nested files
* Deduplicates identical keys across multiple patterns

This issue (https://github.com/spring-cloud/spring-cloud-config/issues/2812)
resolves #2812 

---

### Changes

1. **Core Logic**

   * Extended `getS3ConfigFileWithSearchPaths(...)` to treat `{application}` and `{profile}` placeholders identically to the Git backend
   * Added support for:

     * **Literal + auto-ext**: stops at first existing extension
     * **Dot-wildcard (`.*`)**: expands against supported extensions in priority order
     * **Single-character wildcard (`?`)**
     * **Multi-level wildcard (`**`)**
     * **Directory scan** for literal prefixes ending with `/`
   * Ensured **seenKeys** set to avoid duplicate property sources

2. **Tests**
   Added new JUnit 5 tests in `AwsS3EnvironmentRepositoryTests` to cover **all scenarios**:

   * `searchPaths_placeholderOnly_shouldResolveExactFile`
   * `searchPaths_wildcardOnly_shouldResolveAllProperties`
   * `searchPaths_placeholderAndWildcard_shouldResolveMatchingKeys`
   * `searchPaths_orderMatters_forPropertySourceOrder`
   * `searchPath_extensionPreserved` (dynamic tests for each extension)
   * `searchPaths_applicationAsDirectory_shouldStillHonorSearchPaths`
   * `multiDocumentYaml_withSearchPaths_shouldNotSplitDocuments`
   * `getLocations_returnsCorrect`
   * `searchPaths_deduplication_shouldOnlyAddOnce`
   * `searchPaths_singleCharacterWildcard_shouldMatchExactlyOneChar`
   * `searchPaths_withEmptyLabel_shouldUseDefaultLabel`
   * `searchPaths_multipleLabels_shouldApplyForEachLabelInReverseOrder`
   * `searchPaths_literalNotFound_shouldReturnEmpty`

   …plus the additional edge cases for literal-stop, dot-wildcard priority, and nested directory patterns.

3. **Example Usage**

```yaml
spring:
  cloud:
    config:
      server:
        awss3:
          bucket: my-config-bucket
          search-paths:
            - "{label}/{application}"          # literal + auto-ext
            - "{label}/{application}.*"        # dot-wildcard expansion
            - "{label}/common/*.json"          # JSON only in common/
            - "{label}/{application}.yml"      # explicit YML
```
---

### Backward Compatibility

This update does **not** break any existing AWS S3 config setups. The default lookup behavior is unchanged if no placeholders or wildcards are used in `searchPaths`.

### Migration

Existing users migrating from Git-backed config servers can now use their current `searchPaths` (including wildcards and placeholders) on S3 with no renaming or convention change required.

### Documentation

Documentation and usage examples for the enhanced `searchPaths` will be updated in the relevant documentation files after the merge.
If there are specific locations that require documentation updates, please let me know—I will be happy to update them.

### Performance and Cost

Using wildcards (such as `*` or `**`) in searchPaths may result in additional AWS S3 API calls (e.g., ListObjects), especially for large buckets or deeply nested directory patterns.
This could lead to increased latency and higher AWS costs.
Users should consider the structure and size of their buckets when designing searchPaths and monitor AWS S3 usage accordingly.

----


tomy8964:
Hi @ryanjbaxter!
I have updated the branch to resolve the merge conflicts with the latest main. Could you please take a look when you have some time? Thank you!

tomy8964:
Hi @ryanjbaxter! 

I have successfully addressed all the feedback from the Copilot review:
1. Fixed property source precedence and default application fallback in `findOne()`.
2. Restored the native Spring Cloud Config profile activation lifecycle (including negated profiles) by refactoring `addPropertySources()`.
3. Added path normalization (trimming leading slashes and duplicate `//`) and optimized literal extension probing to avoid redundant S3 calls.
4. Isolated the test state in `AwsS3EnvironmentRepositoryTests` by using a fresh `ConfigServerProperties` instance.

The checks are passing, and it's ready for your review. Could you please take another look when you have some time? Thank you!

tomy8964:
Hi @ryanjbaxter!

I have successfully addressed the second round of feedback from the Copilot review:

* **Generalized S3 Search Logic:** Refactored `getS3ConfigFileWithSearchPaths` by introducing a functional `Function` (`keyWrapper`) parameter, eliminating duplicate S3 object crawling and pagination logic between profile-specific and negated-profile cases.
* **Restored Negated Profiles:** Implemented `wrapKeyWithNegatedConfigFiles` to fully support multi-document YAML files with negated profile expressions under the `searchPaths` architecture.
* **Formatting & Factory Cleanup:** Streamlined the method chains, wrapped arguments to comply with project formatting conventions, and removed the redundant `setOrder` call in `AwsS3EnvironmentRepositoryFactory`.
* **Test Syntax Fixes:** Corrected minor YAML formatting issues (missing spaces after colons) within `AwsS3EnvironmentRepositoryTests`.

The checks are passing again, and the codebase is now much cleaner and more reusable. Could you please take another look when you have some time? Thank you!

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -67,16 +81,24 @@ public class AwsS3EnvironmentRepository implements EnvironmentRepository, Ordere
...
 
+	private final List<String> searchPaths;
+
 	public AwsS3EnvironmentRepository(S3Client s3Client, String bucketName, ConfigServerProperties server) {
 		this(s3Client, bucketName, false, server);
 	}
 
 	public AwsS3EnvironmentRepository(S3Client s3Client, String bucketName, boolean useApplicationAsDirectory,
 			ConfigServerProperties server) {
+		this(s3Client, bucketName, useApplicationAsDirectory, server, null);
```
ryanjbaxter: Could we pass an empty list here?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentProperties.java
```diff
@@ -48,6 +52,16 @@ public class AwsS3EnvironmentProperties implements EnvironmentRepositoryProperti
 
 	private int order = DEFAULT_ORDER;
 
+	private List<String> searchPaths = new ArrayList<>();
```
ryanjbaxter: This could be Collections.emptyList()

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -252,6 +279,185 @@ private List<S3ConfigFile> getS3ConfigFile(String application, String profile, S
...
+				.replace("{profile}", resolvedProfile)
+				.replace("{label}", resolvedLabel);
+
+			pattern = StringUtils.trimLeadingCharacter(pattern.replaceAll("/{2,}", "/"), '/');
+
+			if (!pathMatcher.isPattern(pattern)) {
+				boolean fileFound = false;
+				List<String> extensionsToProbe = (pattern.endsWith(".properties") || pattern.endsWith(".json")
+						|| pattern.endsWith(".yml") || pattern.endsWith(".yaml")) ? List.of("")
+								: List.of(".properties", ".json", ".yml", ".yaml");
```
ryanjbaxter: This could be private static final

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -252,6 +279,185 @@ private List<S3ConfigFile> getS3ConfigFile(String application, String profile, S
...
+					token = resp.nextContinuationToken();
+				}
+				while (token != null);
+
+				continue;
+			}
+
+			if (pattern.endsWith(".*")) {
+				String base = pattern.substring(0, pattern.length() - 2);
+				for (String ext : List.of(".properties", ".json", ".yml", ".yaml")) {
```
ryanjbaxter: extentions could be reused here

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -329,6 +535,32 @@ private S3ConfigFile getS3PropertiesOrJsonConfigFile(String application, String
...
+						Object onProfileValue = properties.get("spring.config.activate.on-profile");
+						if (onProfileValue == null) {
+							onProfileValue = properties.get("spring.config.activate.onProfile");
+						}
+						if (onProfileValue == null) {
+							return YamlProcessor.MatchStatus.NOT_FOUND;
+						}
+
+						String expression = onProfileValue.toString().trim();
+						if (!expression.contains("!") && !expression.contains("&") && !expression.contains("|")
```
ryanjbaxter: Couldnt you use `isSimpleProfileName`?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -473,8 +705,15 @@ class PropertyS3ConfigFile extends S3ConfigFile {
...
 	PropertyS3ConfigFile(String application, String profile, String label, String bucketName,
 			boolean useApplicationAsDirectory, S3Client s3Client) {
+		this(application, profile, label, bucketName, useApplicationAsDirectory, s3Client, true);
+	}
+
+	PropertyS3ConfigFile(String application, String profile, String label, String bucketName,
+			boolean useApplicationAsDirectory, S3Client s3Client, boolean callReadImmediately) {
 		super(application, profile, label, bucketName, useApplicationAsDirectory, s3Client);
-		this.properties = read();
+		if (callReadImmediately) {
```
ryanjbaxter: Is this necessary?  There are subclasses that now call read immediately, so could this implementation just not call read in the constructor?  And if so I would prefer to introduce a new class which has the new functionality (not call read immediately) and then leave these as is.

ryanjbaxter (changes requested):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryFactory.java
```diff
@@ -38,10 +41,8 @@ public AwsS3EnvironmentRepository build(AwsS3EnvironmentProperties environmentPr
 		configureClientBuilder(clientBuilder, environmentProperties.getRegion(), environmentProperties.getEndpoint());
 		final S3Client client = clientBuilder.build();
 
-		AwsS3EnvironmentRepository repository = new AwsS3EnvironmentRepository(client,
-				environmentProperties.getBucket(), environmentProperties.isUseDirectoryLayout(), server);
-		repository.setOrder(environmentProperties.getOrder());
```
ryanjbaxter, line 43: Why is setOrder no longer called?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -200,15 +223,19 @@ private List<S3ConfigFile> getNegatedProfileS3ConfigFileYaml(String application,
...
+				: getS3ConfigFileWithSearchPaths(app, profile, label,
+						key -> wrapKeyWithConfigFiles(key, app, profile, label));
 		addPropertySource(environment, s3ConfigFiles);
 	}
 
 	private void addNonProfileSpecificPropertySource(Environment environment, String app, String profile,
 			String label) {
-		List<S3ConfigFile> s3ConfigFiles = getS3ConfigFile(app, profile, label,
-				this::getNonProfileSpecificPropertiesOrJsonConfigFile, this::getNonProfileSpecificS3ConfigFileYaml);
+		List<S3ConfigFile> s3ConfigFiles = searchPaths.isEmpty() ? getS3ConfigFile(app, profile, label,
```
ryanjbaxter, line 247: When searchPaths is set why are nonProfileSpecificPropertiesOrJsonConfigFile not called?  Wouldn't this result in application.* files not being included?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -252,6 +279,185 @@ private List<S3ConfigFile> getS3ConfigFile(String application, String profile, S
...
+				String base = pattern.substring(0, pattern.length() - 2);
+				for (String ext : List.of(".properties", ".json", ".yml", ".yaml")) {
+					String key = base + ext;
+					if (!seenKeys.add(key)) {
+						continue;
+					}
+					try {
+						s3Client.headObject(HeadObjectRequest.builder().bucket(bucketName).key(key).build());
+						result.addAll(keyWrapper.apply(key));
+						break;
```
ryanjbaxter: Wouldn't this stop the for loop after the first file is found and we wouldn't check the rest of the extensions?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -252,6 +279,185 @@ private List<S3ConfigFile> getS3ConfigFile(String application, String profile, S
...
+					}
+					catch (S3Exception e) {
+						int status = e.statusCode();
+						if (status != 404 && status != 403) {
+							throw e;
```
ryanjbaxter: Might be worth logging this as INFO in case there is a legitimate configuration error occurring and not just that the file does not exist

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -252,6 +279,185 @@ private List<S3ConfigFile> getS3ConfigFile(String application, String profile, S
 
 	}
 
+	private List<S3ConfigFile> getS3ConfigFileWithSearchPaths(String application, String profile, String label,
```
ryanjbaxter, line 309: I would prefer this method be broken up into small more digestible chunks to make it easier to reason about and maintain/test

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -607,18 +847,99 @@ class JsonS3ConfigFile extends YamlS3ConfigFile {
...
+	}
+
 	@Override
 	protected List<String> getExtensions() {
 		return List.of("json");
 	}
 
 }
 
+class PropertyConfigFileFromKey extends PropertyS3ConfigFile {
```
ryanjbaxter: These three classes are nearly identical.  That combined with the callReadImmediately comment I had above makes it feel like these classes could be refactored

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -67,16 +81,24 @@ public class AwsS3EnvironmentRepository implements EnvironmentRepository, Ordere
...
 	public AwsS3EnvironmentRepository(S3Client s3Client, String bucketName, ConfigServerProperties server) {
 		this(s3Client, bucketName, false, server);
 	}
 
 	public AwsS3EnvironmentRepository(S3Client s3Client, String bucketName, boolean useApplicationAsDirectory,
 			ConfigServerProperties server) {
+		this(s3Client, bucketName, useApplicationAsDirectory, server, null);
+	}
+
+	public AwsS3EnvironmentRepository(S3Client s3Client, String bucketName, boolean useApplicationAsDirectory,
```
ryanjbaxter, line 100: It seems better at this point to just have a constructor that takes in AwsEnvironmentProperties

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryTests.java
```diff
@@ -762,6 +767,260 @@ public void getLocationsTest() {
 				"default", "defaultlabel", null, new String[] { "s3://test/defaultlabel" }));
 	}
 
+	// 1) Placeholder only
```
ryanjbaxter: We dont need the number comments on the tests

ryanjbaxter (changes requested):
Could you also add documentation?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we updated this constructor to pass Collections.emptyList() instead of null.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentProperties.java
tomy8964: Yes, we initialized searchPaths using Collections.emptyList() to optimize memory usage and guarantee immutability by default.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we extracted these extension lists into private static final constants (SUPPORTED_EXTENSIONS and EMPTY_EXTENSION) at the class level.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we replaced it to reuse the static SUPPORTED_EXTENSIONS constant.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we refactored isSimpleProfileName to be a package-private static helper method and reused it here to check for simple profile names.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: We refactored this entire part. Instead of adding a callReadImmediately flag to the base classes, we consolidated all three key-based config file classes (PropertyConfigFileFromKey, YamlConfigFileFromKey, JsonConfigFileFromKey) into a single S3ConfigFileFromKey class. The base classes remain unchanged (calling read() in the constructor as they originally did), and S3ConfigFileFromKey handles reading appropriately, allowing us to remove the callReadImmediately flag entirely.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryFactory.java
tomy8964: We introduced a new constructor in AwsS3EnvironmentRepository that accepts AwsS3EnvironmentProperties directly. Inside this constructor, the order is mapped and set directly from the properties (this.order = properties.getOrder()), making the repository initialization cleaner in the factory.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: When searchPaths is active, the wildcard directory scanning in addProfileSpecificPropertySource automatically discovers both profile-specific and non-profile-specific files (e.g. application.yml and application-profile.yml) matching the patterns. For each YAML file found, it loads both profile-specific and non-profile-specific documents. Skipping it here prevents duplicate S3 API calls and duplicate property sources in the environment.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, that was a mistake in the previous version. We removed the break statement so that all matching files with different extensions (e.g., both .properties and .yml at the same path) are scanned and loaded, aligning with Git/Native backend behavior.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we added an INFO level log statement (LOG.info("Error checking S3 object key: " + key, e);) before rethrowing the exception to make troubleshooting configuration errors easier.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we broke down getS3ConfigFileWithSearchPaths into smaller helper methods (probeLiteralPattern, scanDirectoryPattern, probeDotWildcardPattern, scanWildcardPattern) to improve readability and testability.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we completely refactored this by consolidating PropertyConfigFileFromKey, YamlConfigFileFromKey, and JsonConfigFileFromKey into a single S3ConfigFileFromKey class.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we added a new constructor AwsS3EnvironmentRepository(S3Client, AwsS3EnvironmentProperties, ConfigServerProperties) that delegates to the canonical constructor, simplifying the factory builder.

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryTests.java
tomy8964: Yes, we removed the numbered comments from the test cases.

tomy8964:
Hi @ryanjbaxter!
We have successfully addressed all of your feedback and suggestions. Here is a summary of the changes:
### Summary of Changes
1. **`searchPaths` Optimization & Safety**
   * Initialized `searchPaths` using `Collections.emptyList()` by default in both `AwsS3EnvironmentProperties` and `AwsS3EnvironmentRepository`.
   * Updated the repository constructor to guard against `null` and pass `Collections.emptyList()` instead.
   * Cleaned up the factory builder by introducing a constructor that accepts `AwsS3EnvironmentProperties` directly, removing the redundant `setOrder` call.
2. **Cleanups & Constant Extraction**
   * Extracted S3 file extension arrays into class-level `private static final` constants (`SUPPORTED_EXTENSIONS` and `EMPTY_EXTENSION`) and reused them.
   * Refactored `isSimpleProfileName` to be a package-private static helper method and reused it in `wrapKeyWithNegatedConfigFiles`.
   * Removed numbered comments in the test cases (`AwsS3EnvironmentRepositoryTests`).
3. **Subclass Refactoring**
   * Consolidated the redundant key-based S3ConfigFile subclasses (`PropertyConfigFileFromKey`, `YamlConfigFileFromKey`, and `JsonConfigFileFromKey`) into a single `S3ConfigFileFromKey` class, completely removing the `callReadImmediately` flag.
4. **S3 File Probe & Scanning Improvements**
   * Corrected a bug by removing the `break` statement in the literal file probing loop, ensuring all matching files with different extensions (e.g., both `.properties` and `.yml` at the same path) are properly loaded.
   * Added an `INFO` level log statement when checking S3 object keys before rethrowing unexpected `S3Exception`s to assist troubleshooting.
   * Refactored `getS3ConfigFileWithSearchPaths` into smaller, self-documenting helper methods (`probeLiteralPattern`, `scanDirectoryPattern`, `probeDotWildcardPattern`, `scanWildcardPattern`) to improve readability and maintainability.
5. **Documentation**
   * Added comprehensive documentation for the S3 `search-paths` feature in `aws-s3-backend.adoc`, explaining literal/directory paths, wildcards, placeholders, and providing a YAML configuration example.
All tests have been run and verified. Please let us know if there is anything else that needs to be addressed. Thank you! 🙏

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -244,12 +295,202 @@ private String[] parseProfiles(String profiles) {
...
 		}
-		return new ArrayList<>(yamlCreator.create(application, profile, label));
+		int slash = pattern.lastIndexOf('/', idx);
+		return (slash == -1 ? "" : pattern.substring(0, slash + 1));
+	}
 
+	private Optional<S3ConfigFile> createConfigFileFromKey(String key, String application, String profile,
+			String label) {
+		String ext = key.substring(key.lastIndexOf('.') + 1);
+		if ("properties".equalsIgnoreCase(ext)) {
```
ryanjbaxter: The return for these 3 conditions looks identical to me

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -67,27 +86,37 @@ public class AwsS3EnvironmentRepository implements EnvironmentRepository, Ordere
...
+		this(s3Client, properties.getBucket(), properties.isUseDirectoryLayout(), server, properties.getSearchPaths());
+		this.order = properties.getOrder();
 	}
 
 	@Override
 	public int getOrder() {
 		return this.order;
 	}
 
-	public void setOrder(int order) {
```
ryanjbaxter, line 87: We cannot remove public methods

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryTests.java
```diff
@@ -762,6 +767,280 @@ public void getLocationsTest() {
...
+
+	@Test
+	public void searchPaths_applicationAsDirectory_shouldStillHonorSearchPaths() {
+		List<String> paths = List.of("{label}/{application}/foo.*");
+		AwsS3EnvironmentRepository repo = new AwsS3EnvironmentRepository(s3Client, "bucket1", true, server, paths);
+		putFiles("v1/foo/foo.properties", "k=v\n");
+		putFiles("v1/foo/foo.json", jsonContent);
+
+		Environment env = repo.findOne("foo", "", "v1");
+		assertThat(env.getPropertySources()).hasSize(2);
```
ryanjbaxter, line 850: What about validating the 2nd property source

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -244,12 +295,202 @@ private String[] parseProfiles(String profiles) {
...
+		S3ConfigFileFromKey config = new S3ConfigFileFromKey(key, application, profile, label, bucketName,
+				useApplicationAsDirectory, s3Client, properties -> !YamlS3ConfigFile.onProfilePropertyExists(properties)
+						? YamlProcessor.MatchStatus.FOUND : YamlProcessor.MatchStatus.NOT_FOUND);
+		return List.of(config);
+	}
+
+	private String extractPrefix(String pattern) {
+		int idx = pattern.indexOf('*');
+		int q = pattern.indexOf('?');
+		if (q != -1 && (idx == -1 || q < idx)) {
```
ryanjbaxter: I think there is a bug here if the pattern begins with `*`.

If q == -1 and idx == 0 (wildcard at the very start), the else if is never entered because the outer if was false (q == -1), so idx stays 0 and the method falls through to the lastIndexOf logic rather than returning ""

ryanjbaxter (changes requested):

tomy8964:
Hi @ryanjbaxter!
We have successfully addressed the latest round of feedback:
- **Consolidated redundant checks**: Simplified `createConfigFileFromKey` by checking extension mapping directly against `SUPPORTED_EXTENSIONS`.
- **Restored public method**: Restored the `public void setOrder(int order)` method in `AwsS3EnvironmentRepository` to ensure backward compatibility.
- **Improved test assertions**: Added an assertion to validate the name of the second property source (`v1/foo/foo.json`) inside the directory layout search-paths test.
- **Refactored prefix extraction**: Redesigned `extractPrefix` to locate the first wildcard index of either `*` or `?` cleanly and safely, removing any potential edge-case bugs when the wildcard starts at index 0.
Checkstyle checks now pass cleanly locally with zero violations. Please take another look when you have some time. Thank you! 🙏

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we refactored `extractPrefix` to explicitly compute the first index of `*` or `?` first, cleanly returning an empty string if it's less than or equal to 0. This removes the complex `if-else` branch structure and resolves any potential ambiguity.

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepositoryTests.java
tomy8964: Yes, we added a new assertion to validate that the second property source (`v1/foo/foo.json`) is correctly resolved and present in the environment.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we restored the `public void setOrder(int order)` method back to `AwsS3EnvironmentRepository` to preserve backward compatibility for any external consumers.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, we consolidated these identical return branches into a single check using `SUPPORTED_EXTENSIONS.contains("." + ext.toLowerCase(Locale.ROOT))`.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
```diff
@@ -200,40 +234,61 @@ private List<S3ConfigFile> getNegatedProfileS3ConfigFileYaml(String application,
...
 
 	private void addNonProfileSpecificPropertySource(Environment environment, String app, String profile,
 			String label) {
-		List<S3ConfigFile> s3ConfigFiles = getS3ConfigFile(app, profile, label,
-				this::getNonProfileSpecificPropertiesOrJsonConfigFile, this::getNonProfileSpecificS3ConfigFileYaml);
-		addPropertySource(environment, s3ConfigFiles);
+		List<S3ConfigFile> s3ConfigFiles = searchPaths.isEmpty() ? getS3ConfigFile(app, profile, label,
+				this::getNonProfileSpecificPropertiesOrJsonConfigFile, this::getNonProfileSpecificS3ConfigFileYaml)
+				: Collections.emptyList();
+		if (s3ConfigFiles != null) {
```
ryanjbaxter: I dont think this null check is necessary

ryanjbaxter (approved):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/AwsS3EnvironmentRepository.java
tomy8964: Yes, since `getS3ConfigFile` always returns an instantiated `ArrayList` and `Collections.emptyList()` is also never null, `s3ConfigFiles` will never be null. We have removed the redundant null check and pushed the change with a DCO sign-off.

tomy8964:
Hi @ryanjbaxter!
I have removed the redundant null check in `addNonProfileSpecificPropertySource` and pushed the updated commit (fully signed off for DCO). 
Thank you so much for your approval and helpful guidance throughout this PR! 🙏

PR #3175, opened 2026-01-06
Fix invalid YAML Environment conversion when ?resolvePlaceholders=true

zackman0010:
When converting an environment to YAML and resolving placeholders, the produced YAML will sometimes be invalid. Specifically, this happens when a placeholder is replaced with a string that starts with `{`, usually `{cipher}`. `{` is a reserved character in YAML and needs to be quoted.

When SnakeYAML sees a string that starts with `{`, it will print it as a single-quoted string with all other strings remaining unquoted:
```yaml
parent:
  encrypted: '{cipher}1234567890abcdef'
  reference: ${parent.encrypted}
```
Spring will then take this produced String and resolve the properties in it, creating invalid YAML:
```yaml
parent:
  encrypted: '{cipher}1234567890abcdef'
  reference: {cipher}1234567890abcdef # Invalid YAML
```

This PR fixes the issue by running a second post-process step to wrap any values that start with a `{` in single-quotes, matching the rest of the YAML structure. The below screenshot shows an example of the added line of code running.
image

### DIFF
--- a/spring-cloud-config-sample/pom.xml
+++ b/spring-cloud-config-sample/pom.xml
@@ -83,6 +83,11 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-restclient</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.retry</groupId>
 			<artifactId>spring-retry</artifactId>
 		</dependency>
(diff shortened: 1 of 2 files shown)


zackman0010:
> Please add a test so we can guard against regressions

I've added a regression test, but I wasn't able to run it. For some reason, `org.springframework.boot.restclient` is saying it doesn't exist even when using a fresh clone. I'm not sure if something is wrong on my system or if there's a dependency missing. Do you have any ideas?

spencergibb:
Yes, it's a breaking change in boot, we will take care of it

zackman0010:
After merging in the latest commit from main, I was able to run my test and confirm it was successful.

Issue #2441, opened 2024-07-19
Automatically resolve File to base64 encoded value in configuration.


KrishnaST:
I want to server binary files as values through configuration from config server.  
e.g. The following configuration 

`keystoreFile='{file}`keystore.jks

Should be resolved to by reading content of mentioned file and converting to base64

`keystoreFile=YWJjZGV...

This feature request is based on similar underlying need [issues-490](https://github.com/spring-cloud/spring-cloud-config/issues/490)

johnycho:
Hi @ryanjbaxter @KrishnaST 

As discussed above, I’ve opened a PR to address the functionality described in this issue.
Here is the PR link: [Support resolving file content to Base64 via {file} prefix #3179](https://github.com/spring-cloud/spring-cloud-config/pull/3179)

I’d appreciate it if you could review when you have a moment. Thanks!

PR #3179, opened 2026-01-14
Support resolving file content to Base64 via {file} prefix

johnycho:
This commit introduces `FileResolvingEnvironmentRepository`, a decorator that intercepts property values starting with the `{file}` prefix. It reads the referenced file from the local file system, encodes the content to Base64, and replaces the original property value.

Key changes:
- Implemented `FileResolvingEnvironmentRepository` to wrap existing repositories.
- Ensured thread safety and immutability by creating new `PropertySource` instances instead of modifying existing maps.
- Added `FileResolvingEnvironmentRepositoryConfiguration` for auto-configuration.
- Added unit tests covering logic verification and immutable map handling.

Fixes gh-2441

### DIFF
--- a/docs/modules/ROOT/pages/server/environment-repository.adoc
+++ b/docs/modules/ROOT/pages/server/environment-repository.adoc
@@ -58,17 +58,33 @@ spring:
 
 WARNING: Enabling this feature allows the Config Server to read files from the local file system where the server is running. Ensure that the process has appropriate file system permissions and is running in a secure environment.
 
-When this feature is enabled, you can use the `{file}` prefix in your configuration values followed by the path to the file.
+==== Usage
+
+You can use the `{file}` prefix in your configuration values followed by the path to the file.
 The Config Server will read the file, encode its content to a Base64 string, and replace the property value.
 
-For example, in your backing repository (e.g., Git or Native), you can define a property like this:
+**1. Absolute Path**
+
+You can reference files on the Config Server's local file system using an absolute path:
 
 [source,yaml]
 ----
 server:
   ssl:
     key-store: {file}/etc/certs/keystore.jks
-    key-password: my-secret-password
 ----
 
-In this example, the Config Server reads `/etc/certs/keystore.jks`, encodes it, and returns the Base64 string as the value of `server.ssl.key-store`.
+**2. Relative Path (Repository-aware)**
+
+If you are using a repository that supports search paths (like Git, SVN, or Native), you can reference files **relative to the repository root** by starting the path with a dot (`.`):
+
+[source,yaml]
+----
+server:
+  ssl:
+    # Resolves 'certs/keystore.jks' located inside the Git repository
+    key-store: {file}./certs/keystore.jks
+----
+
+In this case, the Config Server will look for the file inside the cloned repository directory.
+If the repository does not support search paths (e.g., JDBC, Vault), relative paths will be ignored.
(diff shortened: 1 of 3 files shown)


ryanjbaxter:
Thanks for the PR.

This is something that we would have to consider merging into our next minor release as a new feature.  We won't begin that work until the July timeframe.

Please check the build failure and fix any errors.

I also think that this environment repository should be disabled by default since it has the potential to expose file content.

We would also need documentation on how this feature works.

johnycho:
Thanks for the review and the detailed feedback! @ryanjbaxter 

I completely understand the timeline regarding the next minor release in July.

I have just pushed the updates addressing your comments:
1. **Security**: The feature is now **disabled by default**. Users must explicitly enable it via `spring.cloud.config.server.file-resolving.enabled=true`.
2. **Tests**: Added `FileResolvingEnvironmentRepositoryConfigurationTests` to verify the auto-configuration logic and ensure the feature remains disabled by default.
3. **Documentation**: Added a section to `environment-repository.adoc` explaining usage and security warnings.
4. **Build**: Fixed the Checkstyle errors and verified the build passes locally.

Please let me know if there is anything else needed.

ryanjbaxter:
I am wondering if it would be useful to be able to reference another file in the environment repository.  For example the keystore could reside in the git repo and you might want to reference that file rather than a file in the filesystem of the config server.

johnycho:
Thank you for the feedback. @ryanjbaxter 

I see the point about referencing a separate Git repo, but in practice keeping artifacts in the same environment repo can be safer when configs and keystores must evolve together.

Splitting into another repo adds a version-coupling problem (which config revision matches which keystore revision), which can become another operational risk.

This PR is not trying to dictate repo structure, but to eliminate server-local file deployment as a dependency and improve reproducibility. Multi-repo support sounds like a good follow-up, but likely belongs in a separate issue/PR.

ryanjbaxter:
You are misunderstanding what I am suggesting I think...

Say you have this in `application.yml` within a git repo configured as an environment repsitory for the config server:

```
keystoreFile='{file}'/path/to/keystore.jks
```

My understanding of the current functionality is that `path/to/keystore.jks` is a path to a file residing on the config server.

What I think would be another common use case would be a path relative to `application.yml` within the git repo itself the config server is using.  For example:

```
keystoreFile='{file}' ./path/in/gitrepo/keystore.jks
```

johnycho:
Thank you for the clarification. @ryanjbaxter 

Just to confirm — would it be acceptable to extend the current {file} resolution so that it can also handle paths relative to the environment Git repository, rather than only files on the Config Server’s local filesystem?

For example, allowing something like
```
keystoreFile: "{file}./path/in/gitrepo/keystore.jks"
```
to resolve within the repository.

ryanjbaxter:
That is what I am suggesting.

One thing to keep in mind is that not all EnvironmetRepositories have files, for example JDBC

johnycho:
Thanks @ryanjbaxter for the suggestion!

I have updated the PR to support **relative paths** pointing to files within the configuration repository (e.g., Git).

### Key Changes:
1. **`SearchPathLocator` Implementation**: The `FileResolvingEnvironmentRepository` now implements `SearchPathLocator`. It delegates `getLocations()` calls to the underlying repository to find where the configuration files are stored locally.
2. **Relative Path Logic**: If a property value starts with `{file}./`, it now iterates through the locations provided by the repository to find the file.
   - Example: `key: {file}./certs/my-key.jks` will look for `certs/my-key.jks` inside the cloned Git repo.
3. **Graceful Fallback**: If the underlying repository does not support search paths (e.g., JDBC), or if the file is not found, the property remains unchanged and a warning log is recorded.
4. **Documentation**: Updated `environment-repository.adoc` to explain both absolute and relative path usage.
5. **Tests**: Added comprehensive tests covering relative path resolution and interface delegation.

I believe this covers the use case you mentioned!

PR #3221, opened 2026-04-26
Add ObservationRegistry support to ConfigClientRequestTemplateFactory

### LINKED ISSUE #2972: Configure observability for RestTemplate used in ConfigServerConfigDataLoader
When using Spring Cloud Config Client in a Spring Boot application, it's not possible to monitor the behavior of the integration with the Config Server because the `RestTemplate` instance used in `ConfigServerConfigDataLoader` is not instrumented with metrics nor observations using Micrometer.

I would like for the `RestTemplate` instance used in `ConfigServerConfigDataLoader` to be instrumented to export metrics and traces. At least, I would like a hook I can use to provide the instrumentation myself.

**Describe alternatives you've considered**
I tried customizing the `RestTemplate` instance used in `ConfigServerConfigDataLoader` as documented in https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#_providing_a_custom_resttemplate_using_config_data. However, it cannot be configured with the same `ObservationRestTemplateCustomizer` used for the general `RestTemplate` configured by Spring Boot because I can't get a hold of the `ObservationRegistry` bean which is configured by Spring Boot at a later point.

**Additional context**
Besides normal production observability needs, having this instrumentation would be beneficial to understand the bahavior of the Spring Cloud Config Client library, especially when it comes to retries.

LuccasAps:
## What this PR does

Closes #2972

The `RestTemplate` used in `ConfigServerConfigDataLoader` was not instrumented
with Micrometer observations, making it impossible to monitor the behavior of
the Config Server integration through metrics and traces.

### `ConfigClientRequestTemplateFactory`
- Added `ObservationRegistry` as an optional field
- Added a new constructor that accepts `ObservationRegistry`
- The existing constructor defaults to `ObservationRegistry.NOOP` to maintain backward compatibility
- When a non-NOOP registry is provided, applies `ObservationRestTemplateCustomizer` to the created `RestTemplate`

### `ConfigServerBootstrapper`
- Added `withObservationRegistry(ObservationRegistry)` hook so developers can provide an `ObservationRegistry` during the bootstrap phase

### `ConfigServerConfigDataLocationResolver`
- Updated `ConfigClientRequestTemplateFactory` registration to pass the `ObservationRegistry` from the `BootstrapContext` when available, falling back to `ObservationRegistry.NOOP`

## Usage

```java
SpringApplication app = new SpringApplication(MyApp.class);
app.addBootstrapRegistryInitializer(
    new ObservationConfigServerBootstrapper()
        .withObservationRegistry(myObservationRegistry)
);
app.run(args);
```

## Backward compatibility

No breaking changes. When no `ObservationRegistry` is provided,
`ObservationRegistry.NOOP` is used and behavior is identical to before.


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientRequestTemplateFactory.java
```diff
@@ -83,6 +97,11 @@ public RestTemplate create() {
 			template.setInterceptors(Arrays.asList(new GenericRequestHeaderInterceptor(headers)));
 		}
 
+		if (observationRegistry != null && observationRegistry != ObservationRegistry.NOOP) {
+			new ObservationRestTemplateCustomizer(observationRegistry, new DefaultClientRequestObservationConvention())
```
ryanjbaxter: Seems like we could create `ObservationRestTemplateCustomizer` in the constructor

ryanjbaxter (changes requested):
ConfigServerBootstrapper and ConfigClientRequestTemplateFactory will result in ClassNotFoundExceptions since micrometer is brought in via optional actuator dependencies.  We probably want to subclass both of these classes with the observation changes and only create them when `ClassUtils.isPresent("io.micrometer.observation.ObservationRegistry", null)`

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientRequestTemplateFactory.java
```diff
@@ -83,6 +97,11 @@ public RestTemplate create() {
 			template.setInterceptors(Arrays.asList(new GenericRequestHeaderInterceptor(headers)));
 		}
 
+		if (observationRegistry != null && observationRegistry != ObservationRegistry.NOOP) {
```
ryanjbaxter: `observationRegistry` would never be null

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientRequestTemplateFactoryTests.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Luccas Asaphe
+ *
+ */
+public class ConfigClientRequestTemplateFactoryTests {
+
+	@Test
+	void shouldInstrumentRestTemplateWhenObservationRegistryProvided() {
```
ryanjbaxter: Please make comments in English

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientRequestTemplateFactoryTests.java
```diff
@@ -0,0 +1,77 @@
...
+
+		// 3. verificar os interceptors
+		assertThat(restTemplate.getObservationRegistry()).isEqualTo(ObservationRegistry.NOOP);
+	}
+
+	@Test
+	void shouldNotInstrumentRestTemplateWhenObservationRegistryIsNoop() {
+		// 1. montar a factory
+		ConfigClientProperties properties = new ConfigClientProperties();
+		ObservationRegistry registry = ObservationRegistry.NOOP;
```
ryanjbaxter: This is never used

LuccasAps (commented):
Thanks for the review! I've addressed all the feedback:
- Removed the null check for observationRegistry
- Moved ObservationRestTemplateCustomizer instantiation to the constructor
- Translated test comments to English
- Used registry variable directly in the constructor instead of ObservationRegistry.NOOP
- Added integration test for withObservationRegistry

LuccasAps:
> ConfigServerBootstrapper and ConfigClientRequestTemplateFactory will result in ClassNotFoundExceptions since micrometer is brought in via optional actuator dependencies. We probably want to subclass both of these classes with the observation changes and only create them when `ClassUtils.isPresent("io.micrometer.observation.ObservationRegistry", null)`

Sorry! I had already applied the individual file changes before seeing this comment. 
After reading it, I refactored the implementation to use subclasses as suggested:

- Created `ObservationConfigClientRequestTemplateFactory` extending `ConfigClientRequestTemplateFactory` with all Micrometer-specific imports isolated
- Created `ObservationConfigServerBootstrapper` extending `ConfigServerBootstrapper` with the `withObservationRegistry` hook
- Used `ClassUtils.isPresent("io.micrometer.observation.ObservationRegistry", null)` in `ConfigServerConfigDataLocationResolver` to conditionally instantiate the observation subclass
- Removed all Micrometer imports from the base classes to prevent `ClassNotFoundException` when Actuator is not on the classpath

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
```diff
@@ -240,8 +240,13 @@ public List<ConfigServerConfigDataResource> resolveProfileSpecific(
 			.registerSingleton("configDataConfigClientProperties",
 					event.getBootstrapContext().get(ConfigClientProperties.class)));
 
-		bootstrapContext.registerIfAbsent(ConfigClientRequestTemplateFactory.class,
-				context -> new ConfigClientRequestTemplateFactory(log, context.get(ConfigClientProperties.class)));
+		bootstrapContext.registerIfAbsent(ConfigClientRequestTemplateFactory.class, context -> {
+			ConfigClientProperties props = context.get(ConfigClientProperties.class);
+			if (ClassUtils.isPresent("io.micrometer.observation.ObservationRegistry", null)) {
```
ryanjbaxter: Also need to check for `ObservationRestTemplateCustomizer`

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigServerBootstrapper.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.micrometer.observation.ObservationRegistry;
+
+import org.springframework.boot.bootstrap.BootstrapRegistry;
+
+public class ObservationConfigServerBootstrapper extends ConfigServerBootstrapper {
+
+	private ObservationRegistry observationRegistry;
+
+	static ObservationConfigServerBootstrapper create() {
```
ryanjbaxter: Is there a reason why this is package private but the constructor is public?

ryanjbaxter (changes requested):

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigServerBootstrapper.java
LuccasAps: Just following the superclass pattern, happy to change if needed!

ryanjbaxter: Yes makes sense to be consistent with the visibility (in both places)

LuccasAps (commented):
Added a check for `ObservationRestTemplateCustomizer` as well. Thanks for catching that!

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigServerBootstrapper.java
ryanjbaxter: I don't se this addressed

ryanjbaxter (changes requested):
I think it would be worth while adding a test that uses `@ClassPathExclusions` to test when micrometer and restclient are not on the classpath.  You can see an example of such a test here https://github.com/spring-cloud/spring-cloud-commons/blob/3a1c2dfc5cf73fcabcf53dc0ba6c2da215454de6/spring-cloud-commons/src/test/java/org/springframework/cloud/client/discovery/ManagementServerPortUtilsTests.java#L30

LuccasAps:
> micrometer

After investigating the dependency tree, I found that `micrometer-observation` 
(which contains `ObservationRegistry`) is actually a direct compile dependency of 
`spring-web`, not of the Actuator:

spring-web -> micrometer-observation -> ObservationRegistry

This means `ObservationRegistry` will always be present on the classpath, making the 
first `ClassUtils.isPresent` check redundant. The class that is truly optional is 
`ObservationRestTemplateCustomizer` from `spring-boot-restclient`.

For now I kept both checks and both exclusions in `@ClassPathExclusions`. Should we 
simplify to only check/exclude `spring-boot-restclient`? Happy to update if you think 
that's the right approach.

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigServerConfigDataWithoutMicrometerTests.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * @author Luccas Asaphe
+ *
+ */
+@ClassPathExclusions({ "spring-boot-starter-actuator-*.jar", "spring-boot-restclient-*.jar" })
+public class ConfigServerConfigDataWithoutMicrometerTests {
+
+	@Test
+	void contextStartsWithoutMicrometer() {
```
ryanjbaxter, line 42: Would be good to verify the correct beans are created

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigServerConfigDataWithoutMicrometerTests.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.cloud.test.ClassPathExclusions;
+import org.springframework.context.ConfigurableApplicationContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Luccas Asaphe
+ *
+ */
+@ClassPathExclusions({ "spring-boot-starter-actuator-*.jar", "spring-boot-restclient-*.jar" })
```
ryanjbaxter: Excluding actuator is not needed since it is brought in by web

ryanjbaxter:
> For now I kept both checks and both exclusions in @ClassPathExclusions. Should we
simplify to only check/exclude spring-boot-restclient? Happy to update if you think
that's the right approach.

Looks like there is just one check, and that is fine:

https://github.com/spring-cloud/spring-cloud-config/pull/3221/changes#diff-942b499b60cc628a4dfe9bab68d5fa1dcb65016ba3f23042ff697a2bb172a104R245-R246

LuccasAps (commented):
Updated the test to verify that `ConfigClientRequestTemplateFactory` (base class) is created when `spring-boot-restclient` is not on the classpath, and removed the `spring-boot-starter-actuator` exclusion since it is not needed — only `spring-boot-restclient` needs to be excluded as it contains `ObservationRestTemplateCustomizer`.

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigServerConfigDataWithoutMicrometerTests.java
```diff
@@ -0,0 +1,64 @@
...
+
+	@Test
+	void contextStartsWithoutMicrometer() {
+		try (ConfigurableApplicationContext context = new SpringApplicationBuilder(TestConfig.class)
+			.web(WebApplicationType.NONE)
+			.addBootstrapRegistryInitializer(registry -> registry.addCloseListener(event -> {
+				BootstrapContext bootstrapContext = event.getBootstrapContext();
+				ConfigurableListableBeanFactory beanFactory = event.getApplicationContext().getBeanFactory();
+
+				ConfigClientRequestTemplateFactory templateFactory = bootstrapContext
```
ryanjbaxter: This is not exactly testing the ClassIsPresent logic if you are registering it

LuccasAps (commented):
Done! Refactored the test to verify the ClassIsPresent logic by checking that the RestTemplate has no ObservationRegistry configured when spring-boot-restclient is absent.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerBootstrapper.java
```diff
@@ -35,7 +35,7 @@ public class ConfigServerBootstrapper implements BootstrapRegistryInitializer {
 
 	private LoaderInterceptor loaderInterceptor;
 
-	static ConfigServerBootstrapper create() {
+	public static ConfigServerBootstrapper create() {
```
jonatan-ivanov: Why does this need to be public?
It seems the only new place where this is used is `ObservationConfigServerBootstrapper` which is in the same package and it might be package-private as well as `ConfigServerBootstrapper`.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
```diff
@@ -240,8 +240,14 @@ public List<ConfigServerConfigDataResource> resolveProfileSpecific(
 			.registerSingleton("configDataConfigClientProperties",
 					event.getBootstrapContext().get(ConfigClientProperties.class)));
 
-		bootstrapContext.registerIfAbsent(ConfigClientRequestTemplateFactory.class,
-				context -> new ConfigClientRequestTemplateFactory(log, context.get(ConfigClientProperties.class)));
+		bootstrapContext.registerIfAbsent(ConfigClientRequestTemplateFactory.class, context -> {
+			ConfigClientProperties props = context.get(ConfigClientProperties.class);
+			if (ClassUtils
+				.isPresent("org.springframework.boot.restclient.observation.ObservationRestTemplateCustomizer", null)) {
```
jonatan-ivanov: At this point we are too early in the lifecycle here to check the presence of the bean instead of the class, right?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigClientRequestTemplateFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class ObservationConfigClientRequestTemplateFactory extends ConfigClientRequestTemplateFactory {
+
+	private final ObservationRestTemplateCustomizer observationRestTemplateCustomizer;
+
+	public ObservationConfigClientRequestTemplateFactory(Log log, ConfigClientProperties properties,
+			ObservationRegistry observationRegistry) {
+		super(log, properties);
+		this.observationRestTemplateCustomizer = observationRegistry != ObservationRegistry.NOOP
+				? new ObservationRestTemplateCustomizer(observationRegistry,
```
jonatan-ivanov: Can we/should we reuse the bean that Boot creates (see [`RestTemplateObservationAutoConfiguration`](https://github.com/spring-projects/spring-boot/blob/607e4bbe65e00bc401b3c33e9c1a615a9f1888aa/module/spring-boot-restclient/src/main/java/org/springframework/boot/restclient/autoconfigure/RestTemplateObservationAutoConfiguration.java#L48))?

I mean if the user disables RestTemplateObservationAutoConfiguration, should the observation config for this `RestTemplate` disabled too?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigClientRequestTemplateFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.web.client.RestTemplate;
+
+public class ObservationConfigClientRequestTemplateFactory extends ConfigClientRequestTemplateFactory {
+
+	private final ObservationRestTemplateCustomizer observationRestTemplateCustomizer;
+
+	public ObservationConfigClientRequestTemplateFactory(Log log, ConfigClientProperties properties,
+			ObservationRegistry observationRegistry) {
+		super(log, properties);
+		this.observationRestTemplateCustomizer = observationRegistry != ObservationRegistry.NOOP
```
jonatan-ivanov: Please use `!observationRegistry.isNoop()` instead (it's more flexible).

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigClientRequestTemplateFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+import io.micrometer.observation.ObservationRegistry;
+import org.apache.commons.logging.Log;
+
+import org.springframework.boot.bootstrap.BootstrapContext;
+import org.springframework.boot.restclient.observation.ObservationRestTemplateCustomizer;
+import org.springframework.http.client.observation.DefaultClientRequestObservationConvention;
+import org.springframework.web.client.RestTemplate;
+
+public class ObservationConfigClientRequestTemplateFactory extends ConfigClientRequestTemplateFactory {
```
jonatan-ivanov: Does this need to be public (as well as its ctor)?
I'm thinking if `ConfigClientRequestTemplateFactory` being public is enough?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigClientRequestTemplateFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+	private final ObservationRestTemplateCustomizer observationRestTemplateCustomizer;
+
+	public ObservationConfigClientRequestTemplateFactory(Log log, ConfigClientProperties properties,
+			ObservationRegistry observationRegistry) {
+		super(log, properties);
+		this.observationRestTemplateCustomizer = observationRegistry != ObservationRegistry.NOOP
+				? new ObservationRestTemplateCustomizer(observationRegistry,
+						new DefaultClientRequestObservationConvention())
```
jonatan-ivanov: This is one reason why we should not create a new `ObservationRestTemplateCustomizer`, Boot does something different: it lets users to customize the convention. Even if we duplicate exactly what Boot does, how can we keep this in-sync?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigServerBootstrapper.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.config.client;
+
+import io.micrometer.observation.ObservationRegistry;
+
+import org.springframework.boot.bootstrap.BootstrapRegistry;
+
+public class ObservationConfigServerBootstrapper extends ConfigServerBootstrapper {
```
jonatan-ivanov: Does this need to be public (and its ctor)?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigServerBootstrapper.java
```diff
@@ -0,0 +1,44 @@
...
+
+	private ObservationRegistry observationRegistry;
+
+	public static ObservationConfigServerBootstrapper create() {
```
jonatan-ivanov: Why is this needed (we already have a ctor with the same empty params)?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigServerBootstrapper.java
```diff
@@ -0,0 +1,44 @@
...
+		return new ObservationConfigServerBootstrapper();
+	}
+
+	public ObservationConfigServerBootstrapper withObservationRegistry(ObservationRegistry observationRegistry) {
```
jonatan-ivanov: Is it possible to enable this by default without the user need to do this?
(from your PR description):
```java
SpringApplication app = new SpringApplication(MyApp.class);
app.addBootstrapRegistryInitializer(
    new ConfigServerBootstrapper()
        .withObservationRegistry(myObservationRegistry)
);
app.run(args);
```

Also, I have two questions in connection with this:
- Didn't you want to write: `new ObservationConfigServerBootstrapper()` instead of `new ConfigServerBootstrapper()`?
- Where exactly should users get the `myObservationRegistry`? (Given that it is created by Boot later and we don't want users to end-up with multiple registries.)

spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigClientRequestTemplateFactoryTests.java
```diff
@@ -0,0 +1,77 @@
...
+		ConfigClientProperties properties = new ConfigClientProperties();
+		ObservationRegistry registry = ObservationRegistry.create();
+		ObservationConfigClientRequestTemplateFactory factory = new ObservationConfigClientRequestTemplateFactory(
+				LogFactory.getLog(getClass()), properties, registry);
+
+		// 2. create the RestTemplate
+		RestTemplate restTemplate = factory.create();
+
+		// 3. verify the observation registry
+		assertThat(restTemplate.getObservationRegistry()).isEqualTo(registry);
```
jonatan-ivanov: It would be much better to verify if `RestTemplate` is indeed instrumented by:
- Using `ObservationRegistry`
- Calling `RestTemplate`
- Verifying if Observations are created (see `micrometer-observation-test`)

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigClientRequestTemplateFactory.java
LuccasAps: That's a good point. Using a hardcoded DefaultClientRequestObservationConvention prevents 
users from customizing the convention and creates a sync problem with what Boot does.

We thought about an approach to solve both issues — the convention and the registry sync:
using addCloseListener from the BootstrapContext to re-instrument the RestTemplate when 
the ApplicationContext is ready, at which point both the ObservationRegistry (with handlers 
already registered by ObservationRegistryPostProcessor) and the user's custom 
ClientRequestObservationConvention bean would be available.

The limitation of this approach is that observations from the initial bootstrap call 
(the first fetch from Config Server) would not be captured, since the ApplicationContext 
does not exist yet at that point. Observations from refreshes would be captured correctly.

Given that the issue mentions retries as a key use case:
1. Is it acceptable to miss observations from the initial bootstrap call?
2. Is re-instrumenting via addCloseListener the right approach here?
3. Are there any flaws in this reasoning or anything we might have misunderstood?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigClientRequestTemplateFactory.java
LuccasAps: If we use the addCloseListener approach, this would be partially handled — if the user 
disables RestTemplateObservationAutoConfiguration, the ObservationRestTemplateCustomizer 
bean would not exist in the ApplicationContext and the RestTemplate would not be 
re-instrumented after bootstrap.

However, the RestTemplate would still be instrumented during the bootstrap phase since 
the ObservationRegistry is created there before the ApplicationContext exists.

To fully respect the user's intent, we would need to check during bootstrap whether the 
user has disabled the observation support. One possible approach would be to read a 
property (e.g. spring.cloud.config.client.observation.enabled) via the Binder during 
bootstrap, since it is available at that point.

Does this approach make sense, or is there a better way to handle this?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
LuccasAps: Yes, that's correct. At this point we are in the bootstrap phase, so the ApplicationContext 
does not exist yet and we cannot check for bean presence. Checking for class presence via 
ClassUtils.isPresent is the only viable option here. 
Is there anything we are missing?

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ObservationConfigServerBootstrapper.java
LuccasAps: You are right, the PR description should use new ObservationConfigServerBootstrapper() 
instead of new ConfigServerBootstrapper() — that was a mistake in the description.

Regarding enabling this by default without the user needing to configure it manually — 
we are analyzing an approach using addCloseListener to re-instrument the RestTemplate 
when the ApplicationContext is ready, which would make this automatic. We will share 
more details in the other comment.

Regarding where users should get the ObservationRegistry — this is exactly the problem 
we are trying to solve. The ObservationRegistry is created by Boot in the ApplicationContext, 
which does not exist yet during bootstrap. The addCloseListener approach would solve this 
by accessing the Boot-created ObservationRegistry directly from the ApplicationContext, 
avoiding multiple registries. Does this approach make sense, or is there anything we are missing? We will share more details in the other comment.

jonatan-ivanov: >Regarding where users should get the ObservationRegistry — this is exactly the problem we are trying to solve.

You mean this PR should solve it or you have an idea to solve this? (I'm asking because I think the current changes PR do not solve the issue.)

>The addCloseListener approach would solve this
by accessing the Boot-created ObservationRegistry directly from the ApplicationContext,
avoiding multiple registries. Does this approach make sense, or is there anything we are missing?

I don't know, can you tell me more about the the addCloseListener approach or can you share some code I can look at?

LuccasAps: The idea is to use addCloseListener from the BootstrapContext to re-instrument the RestTemplate 
when the ApplicationContext is ready, at which point the Boot-created ObservationRegistry 
(with handlers already registered) and the user's custom ClientRequestObservationConvention 
bean would be available.

Here is a rough idea of how this would work:

````java
bootstrapContext.addCloseListener(event -> {
    ApplicationContext applicationContext = event.getApplicationContext();
    if (applicationContext.containsBean(ObservationRestTemplateCustomizer.class.getName())) {
        ObservationRestTemplateCustomizer customizer = applicationContext
            .getBean(ObservationRestTemplateCustomizer.class);
        RestTemplate restTemplate = event.getBootstrapContext().get(RestTemplate.class);
        customizer.customize(restTemplate);
    }
});
````

The limitation of this approach is that the initial bootstrap call to the Config Server 
would not be captured since the ApplicationContext does not exist yet at that point. 
Observations from refreshes would be captured correctly.

Does this approach make sense? Is there a better way to handle this?

jonatan-ivanov: >The limitation of this approach is that the initial bootstrap call to the Config Server would not be captured since the ApplicationContext does not exist yet at that point.

I guess that's ok, maybe we can add more logs (in a different issue/PR) if needed to help people troubleshooting things.

>Does this approach make sense? Is there a better way to handle this?

I think it makes sense. Or let me say I can imagine a situation where this would work but the last time I needed to interact with the bootstrap context was many years ago. :) Have you tried this?

@ryanjbaxter What do you think?

LuccasAps:
Refactored the observation support to use a simpler and more automatic approach:

Instead of requiring users to manually configure the ObservationRegistry, the RestTemplate 
used by the Config Client is now automatically instrumented after the ApplicationContext 
is refreshed, using the ObservationRestTemplateCustomizer bean created by Spring Boot.

This approach:
- Requires no user configuration
- Reuses the same ObservationRegistry created by Boot, avoiding multiple registries
- Respects user customizations of ClientRequestObservationConvention
- Respects the user's decision to disable RestTemplateObservationAutoConfiguration
- Avoids ClassNotFoundException when spring-boot-restclient is not on the classpath

Known limitation: the initial bootstrap call to the Config Server is not captured since 
the ApplicationContext does not exist yet at that point. Observations from refreshes are 
captured correctly. This was confirmed as acceptable by the reviewer.

PR #3244, opened 2026-06-09
docs: add observability page to nav and fix heading level

### LINKED ISSUE #2272: Observability Reference In Docs Results In Unresolved
Looks like the reference in `_observability.doc` is referencing the wrong path.  When building the release train docs we get an uresolved error.

https://github.com/spring-cloud/spring-cloud-config/commit/69f11e8608f56c287049e1a6e41abd57b6f91198

won-seoop:
## Problem

The `observability.adoc` page exists in the Antora docs but is not referenced in `nav.adoc`, making it unreachable from the documentation navigation and causing unresolved reference errors when building the release train docs.

Additionally, the page uses a level-2 heading (`== Observability metadata`) instead of a level-1 heading (`= Observability`), which is invalid for a standalone Antora page.

## Changes

- Add `* xref:observability.adoc[]` to `nav.adoc` so the page appears in the sidebar
- Promote the heading from level 2 (`==`) to level 1 (`=`) and update title to match other Spring Cloud projects (e.g. Spring Cloud Stream)

Fixes #2272

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -38,3 +38,4 @@
 ** xref:server/push-notifications-and-bus.adoc[]
 ** xref:server/aot-and-native-image-support.adoc[]
 * xref:client.adoc[]
+* xref:observability.adoc[]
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
Another duplicate

PR #3246, opened 2026-06-09
docs: document Git backend unavailability behaviour and stale-cache fallback

### LINKED ISSUE #572: Add docs for Config Server behavior when backend is not accessible
The current Spring Cloud Config docs describe different configuration parameters for setting up a backend such as a Git repo. It would be helpful to describe expected behavior at runtime when the backend that was setup is not accessible for any number of reasons (network, crash, misconfiguration). Some of the situations to describe could be:

* What happens when Config Server starts up but is not able to access backend (Git)?
* What happens when backend is not accessible after Config Server already started successfully and pulled configuration values?
* What happens if backend becomes unavailable during process of pulling configuration values?
* How is refresh affected by backend becoming unaccessible?

I'm sure that there are other situations to be described. Please let me know if there are any questions. Thank you.

won-seoop:
## What

Closes #572.

Adds a new **"Behavior When the Git Backend Is Unavailable"** section to the Git Backend reference page.

## Changes

A summary table covering four distinct scenarios:

| Scenario | Behaviour |
|---|---|
| No prior local clone + Git unreachable at first request | Clone fails → HTTP 500 |
| Git becomes unreachable **after** at least one successful clone/fetch | Fetch fails silently (WARN log), server falls back to stale local clone → HTTP 200 with potentially outdated values |
| `force-pull: true` + Git unreachable | Reset applied then fetch fails → no stale fallback → HTTP 500 |
| `cloneOnStart: true` + Git unreachable at startup | Server fails to start |

Also includes a NOTE advising operators to monitor WARN log output and enable the Health Indicator, because the stale-cache fallback is completely transparent to clients.

## Motivation

The existing docs explain `cloneOnStart`, `force-pull`, and `refreshRate` in isolation but do not describe what happens to clients when the remote Git server becomes unreachable at various points in the lifecycle. In particular the "stale fallback" path — where the server silently serves an older cached configuration — was completely undocumented and has confused many users.

### DIFF
--- a/docs/modules/ROOT/pages/server/environment-repository/git-backend.adoc
+++ b/docs/modules/ROOT/pages/server/environment-repository/git-backend.adoc
@@ -475,6 +475,46 @@ value of this property is specified in seconds.  By default the value is 0, mean
 the config server will fetch updated configuration from the Git repo every time it
 is requested. If the value is a negative number the refresh will not occur.
 
+[[git-backend-unavailability]]
+== Behavior When the Git Backend Is Unavailable
+
+Spring Cloud Config Server clones the remote Git repository to a local working directory (the _basedir_ cache) when configuration is first requested (or at startup when `cloneOnStart: true`).
+This local clone is used as a read-only cache so that the Config Server can continue serving configuration even if the remote Git backend becomes temporarily unreachable.
+
+The following table summarises the behaviour under various failure scenarios:
+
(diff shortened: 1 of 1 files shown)


ryanjbaxter:
Another duplicate PR
https://github.com/spring-cloud/spring-cloud-config/pull/3251

PR #3248, opened 2026-06-09
Document @RefreshScope + @ControllerAdvice double-invocation issue (GH-1452)

### LINKED ISSUE #1452: @RefreshScope and component-scanned @ControllerAdvice beans.
This is in 2.1.3.RELEASE, and Spring Boot 2.1.7.

Beans that are annotated with both @RefreshScope and @ControllerAdvice, and which are component-scanned (as opposed to explicit configuration) have their @ModelAttribute methods executed twice per incoming request, as opposed to once.  One of the executions is via the proxy, the other is direct.

Simple example:
```java
package test;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;

@SpringBootApplication
public class TestMain
{
    public static void main(final String[] args)
    {
        SpringApplication.run(TestMain.class, args);
    }
}
```
----

```java
package test;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.beans.factory.annotation.Value;
import org.springframework.cloud.context.config.annotation.RefreshScope;
import org.springframework.web.bind.annotation.ControllerAdvice;
import org.springframework.web.bind.annotation.ModelAttribute;

import javax.servlet.http.HttpServletRequest;

@RefreshScope
@ControllerAdvice
public class TestAdvice
{
    private static final Logger logger = LoggerFactory.getLogger(TestAdvice.class);

    @Value("${test}")
    private String testConfigValue;
    
    @ModelAttribute
    public void adviceMethod(final HttpServletRequest httpServletRequest)
    {
        logger.info("advice method executed with " + testConfigValue);
    }
}
```
----

```java
package test;

import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;

@RestController
public class TestController
{
    @GetMapping(value = "/test")
    public String test()
    {
        return "test";
    }
}
```
---

Requests to the controller log as so:

2019-08-20 14:05:07.237  INFO 5156 --- [nio-8080-exec-5] test.TestAdvice                          : advice method executed with testvalue
2019-08-20 14:05:07.237  INFO 5156 --- [nio-8080-exec-5] test.TestAdvice                          : advice method executed with testvalue

Remove the @RefreshScope, and they log just once.

won-seoop:
## Summary

Closes GH-1452

Applying `@RefreshScope` directly to a `@ControllerAdvice` bean causes Spring MVC to discover both the CGLIB proxy (from the refresh scope) and the underlying target class. As a result, `@ModelAttribute` methods are invoked twice per request.

The maintainers already recommended the workaround in the issue thread: extract refreshable values to a separate `@ConfigurationProperties` bean with `@RefreshScope` and inject it into the advice. However, this was never documented, leaving users to discover it on their own.

## Changes

Added a new `[[refresh-scope-known-limitations]]` section to `client.adoc` (placed before the AOT section at the end of the page):
- Explains the root cause: CGLIB proxy + component-scan double-registration
- Provides a concrete before/after code example showing the workaround
- Uses callout annotations to highlight which bean gets `@RefreshScope` and which doesn't

## Test plan
- [ ] Verify AsciiDoc renders correctly
- [ ] Confirm new section appears in the navigation/ToC

### DIFF
--- a/docs/modules/ROOT/pages/client.adoc
+++ b/docs/modules/ROOT/pages/client.adoc
@@ -363,6 +363,44 @@ String name = "World";
 
 The preceding code would sets the value of the `name` variable to `appAsecret`.
 
+[[refresh-scope-known-limitations]]
+== Refresh Scope Known Limitations
+
+=== @RefreshScope and @ControllerAdvice
+
+Applying `@RefreshScope` directly to a `@ControllerAdvice` bean that is discovered via component scanning (for example, annotated with `@ControllerAdvice` in a `@SpringBootApplication` package) causes Spring MVC to register the bean twice: once for the CGLIB proxy created by the refresh scope and once for the underlying target class. This results in `@ModelAttribute` methods (and other advice methods) being invoked twice per request.
+
+This is a known interaction between Spring Cloud's refresh scope proxying and Spring MVC's component scan for `ControllerAdvice` beans.
+
+*Workaround:* Extract the refreshable configuration values into a separate `@ConfigurationProperties` or `@Component` bean annotated with `@RefreshScope`, and inject that bean into the `@ControllerAdvice`. The `@ControllerAdvice` itself should *not* carry `@RefreshScope`:
+
+[source,java]
+----
+@ConfigurationProperties(prefix = "myapp")
+@RefreshScope                                  // <1>
+public class MyAppConfig {
+    private String secret;
+    // getter/setter
+}
+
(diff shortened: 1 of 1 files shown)


ryanjbaxter:
Please do not spam us with multiple PRs for the same issue
https://github.com/spring-cloud/spring-cloud-commons/pull/1696

PR #3278, opened 2026-08-22
Add Google Cloud Parameter Manager environment repository

### LINKED ISSUE #2900: Add EnvironmentRepository For Google's Parameter Manager
See https://cloud.google.com/secret-manager/parameter-manager/docs/overview

This seems like a more generalized version of Secret Manager, which we already support.

yashdotdev13:
Fixes #2900

## Summary

Adds support for using Google Cloud Parameter Manager as an environment repository.

This change:

- Adds the Google Cloud Parameter Manager dependency.
- Adds configuration properties for project ID, location, parameter labels, and repository order.
- Adds `GoogleParameterManagerEnvironmentRepository`.
- Adds `GoogleParameterManagerEnvironmentRepositoryFactory`.
- Registers the repository when `ParameterManagerClient` is available.
- Loads parameters based on application and profile labels.
- Adds unit test coverage for parameter retrieval and rendering.

## Testing

```text
./mvnw.cmd -pl spring-cloud-config-server -Dtest=GoogleParameterManagerEnvironmentRepositoryTests test

### DIFF
--- a/docs/modules/ROOT/pages/server/environment-repository/gcp-parameter-manager-backend.adoc
+++ b/docs/modules/ROOT/pages/server/environment-repository/gcp-parameter-manager-backend.adoc
@@ -1,0 +1,97 @@
+[[gcp-parameter-manager-backend]]
+= Google Cloud Parameter Manager Backend
+
+Spring Cloud Config Server can use link:https://cloud.google.com/parameter-manager[Google Cloud Parameter Manager] as an `EnvironmentRepository`.
+Parameters are exposed as `PropertySource` entries whose values are obtained from the latest parameter versions.
+
+== Enabling the backend
+
+The repository is registered when all of the following hold:
+
+* The Spring profile `parameter-manager` is active.
+* `com.google.cloud.parametermanager.v1.ParameterManagerClient` is on the classpath.
+
+[source,yaml]
+----
+spring:
+  profiles:
+    active: parameter-manager
+  cloud:
+    config:
+      server:
+        gcp-parameter-manager:
+          project-id: my-project
+          location: global
+          order: 0
+          application-label: application
+          profile-label: profile
+          default-label: main
+----
+
+Configuration properties are bound under `spring.cloud.config.server.gcp-parameter-manager`:
+
+|===
+| Property | Default | Description
+
+| `order`
+| Same as other environment repositories
+| Order in a composite repository (`org.springframework.core.Ordered`).
+
+| `application-label`
+| `application`
+| Parameter label key used to match the Config Server `\{application}` name (see <<gcp-parameter-manager-mapping>>).
+
+| `profile-label`
+| `profile`
+| Parameter label key used to match the Config Server profile (see <<gcp-parameter-manager-mapping>>).
+
+| `location`
+| `global`
+| Google Cloud Parameter Manager location.
+
+| `default-label`
+| `main`
+| Label returned by the Config Server when the client does not provide a label.
+
+| `project-id`
+| _empty_
+| Google Cloud project ID containing the parameters. This property must be configured.
+
+|===
+
+[[gcp-parameter-manager-mapping]]
+== Mapping applications and profiles
+
+The repository lists the parameters in the configured Google Cloud project and location once for each Config Client request and filters the result according to the requested application and profiles.
+
+Parameters are matched using their labels:
+
+* The label configured by `application-label` must match the requested `\{application}` name.
+* The label configured by `profile-label` must match the requested profile.
+* Matching is case-insensitive.
+* If an application label is missing, it defaults to `application`.
+* If a profile label is missing, it defaults to `profile`.
+
+When a Config Client requests an application other than `application`, the `application` application is also included first. This allows shared configuration to be provided for all applications.
+
+The configured Config Server default profile is also included before additional requested profiles when necessary.
+
+For every matching parameter, the repository reads the latest version and adds its rendered payload as a property value. The property name is derived from the parameter name.
+
+For example, a parameter with the name:
+
+----
+projects/my-project/locations/global/parameters/my-property
+----
+
+is exposed as:
+
+----
+my-property
+----
+
+== Authentication
+
+The Parameter Manager client is created using the Google Cloud client library's default authentication mechanism. Configure Google Cloud credentials using link:https://cloud.google.com/docs/authentication/application-default-credentials[Application Default Credentials].
+
+This can include credentials provided through environment configuration, local development credentials, or Google Cloud runtime service accounts.
--- a/docs/modules/ROOT/pages/server/environment-repository/gcp-parameter-manager-backend.adoc
+++ b/docs/modules/ROOT/pages/server/environment-repository/gcp-parameter-manager-backend.adoc
@@ -9,7 +9,7 @@ Parameters are exposed as `PropertySource` entries whose values are obtained fro
 The repository is registered when all of the following hold:
 
 * The Spring profile `parameter-manager` is active.
-* `com.google.cloud.parametermanager.v1.ParameterManagerClient` is on the classpath.
+* The `com.google.cloud:google-cloud-parameter-manager` dependency is on the classpath.
 
 [source,yaml]
 ----
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
@@ -26,13 +26,16 @@ import com.google.cloud.parametermanager.v1.RenderParameterVersionResponse;
 
 import org.springframework.cloud.config.environment.Environment;
 import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.cloud.config.server.config.ConfigServerProperties;
 import org.springframework.core.Ordered;
 import org.springframework.util.StringUtils;
 
 /**
  * @author Yash Chauhan
  */
-public class GoogleParameterManagerEnvironmentRepository implements EnvironmentRepository, Ordered {
+public class GoogleParameterManagerEnvironmentRepository implements EnvironmentRepository, Ordered, AutoCloseable {
+
+	private final ConfigServerProperties configServerProperties;
 
 	private final ParameterManagerClient parameterManagerClient;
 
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+
+		this.parameterManagerClient = parameterManagerClient;
+		this.properties = properties;
+		this.order = properties.getOrder();
+	}
+
+	@Override
+	public Environment findOne(String application, String profile, String label) {
+		if (!StringUtils.hasText(label)) {
+			label = "master";
```
ryanjbaxter: To be consistent this should be `main`

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
```diff
@@ -0,0 +1,122 @@
...
+			profile = "default";
+		}
+
+		if (!profile.startsWith("default")) {
+			profile = "default," + profile;
+		}
+
+		String[] profiles = StringUtils.trimArrayElements(StringUtils.commaDelimitedListToStringArray(profile));
+
+		Environment result = new Environment(application, profile, label, null, null);
```
ryanjbaxter: is `application` is not equal to "apploication" then "application" should be included by default

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
```diff
@@ -0,0 +1,122 @@
...
+import org.springframework.cloud.config.environment.PropertySource;
+import org.springframework.core.Ordered;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Yash Chauhan
+ */
+public class GoogleParameterManagerEnvironmentRepository implements EnvironmentRepository, Ordered {
+
+	private final ParameterManagerClient parameterManagerClient;
```
ryanjbaxter, line 40: This needs to be closed when no longer used

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
ryanjbaxter: This is also usually configurable

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	@Override
+	public Environment findOne(String application, String profile, String label) {
+		if (!StringUtils.hasText(label)) {
+			label = "master";
+		}
+
+		if (!StringUtils.hasText(profile)) {
+			profile = "default";
```
ryanjbaxter: This is usually configurable

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
```diff
@@ -0,0 +1,122 @@
...
+		Map<String, String> result = new HashMap<>();
+
+		String projectId = properties.getProjectId();
+		if (!StringUtils.hasText(projectId)) {
+			throw new IllegalStateException("Google Cloud project ID must be configured");
+		}
+
+		LocationName locationName = LocationName.of(projectId, properties.getLocation());
+
+		for (Parameter parameter : client.listParameters(locationName).iterateAll()) {
```
ryanjbaxter: Why call this for each parameter?  Seems wasteful and potentially expensive.  I think we can just call
`lient.listParameters(locationName)` once and then filter that result for each parameter

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepositoryTests.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class GoogleParameterManagerEnvironmentRepositoryTests {
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testFindOne() {
```
ryanjbaxter, line 41: Please add more tests for additional usages

ryanjbaxter (changes requested):
This will need documentation as well

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
yashdotdev13: Addressed in the subsequent changes. The default has been updated to main for consistency, and the value is configurable.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
yashdotdev13: Addressed in the subsequent changes. The application profile is now included by default when the requested application name is different from application.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
yashdotdev13: Addressed in the subsequent changes. The ParameterManagerClient is now closed when it is no longer needed.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
yashdotdev13: Good point. I’ve made the default profile configurable rather than hardcoding it. Thanks!

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
yashdotdev13: Thanks! Good point. I've updated the implementation to call client.listParameters(locationName) once and filter the returned parameters instead of making the call repeatedly. This should avoid unnecessary API calls and reduce potential cost.

spring-cloud-config-server/src/test/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepositoryTests.java
yashdotdev13: Added additional test coverage for the relevant usage scenarios.

yashdotdev13:
Thanks! I've added documentation for the Google Cloud Parameter Manager backend, including configuration properties, application and profile mapping, and authentication.

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
```diff
@@ -0,0 +1,150 @@
...
+			label = this.properties.getDefaultLabel();
+		}
+
+		String defaultProfile = this.configServerProperties.getDefaultProfile();
+
+		if (!StringUtils.hasText(profile)) {
+			profile = defaultProfile;
+		}
+
+		if (!profile.startsWith(defaultProfile)) {
```
ryanjbaxter: startsWith is going to problematic, for example if profile is `development` and defaultProfile is `dev` this would match when it shouldn't

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
```diff
@@ -0,0 +1,150 @@
...
+			}
+		}
+		catch (Exception ex) {
+			throw new IllegalStateException("Could not access Google Cloud Parameter Manager", ex);
+		}
+
+		return result;
+	}
+
+	private Map<String, String> getParameters(Iterable<Parameter> parameterList, String application, String profile,
```
ryanjbaxter, line 125: An exception here, for example when calling `/versions/latest`, will cause the rest of the application + profile combinations to not be fetched

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
```diff
@@ -0,0 +1,150 @@
...
+			String projectId) {
+
+		Map<String, String> result = new HashMap<>();
+
+		for (Parameter parameter : parameterList) {
```
ryanjbaxter, line 130: It looks like every iteration of this loops will result in a network call according to the docs of `AbstractPage.iterateAll()`.  To acoid this we can replace like 91 with 
```
List<Parameter> parameterList = new ArrayList<>();
this.parameterManagerClient.listParameters(locationName).iterateAll().forEach(parameterList::add);
```

ryanjbaxter (changes requested):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
```diff
@@ -0,0 +1,162 @@
...
+			label = this.properties.getDefaultLabel();
+		}
+
+		String defaultProfile = this.configServerProperties.getDefaultProfile();
+
+		if (!StringUtils.hasText(profile)) {
+			profile = defaultProfile;
+		}
+
+		if (!profile.equals(defaultProfile)) {
```
ryanjbaxter: This fixes the prior bug but it introduces a new bug when there is more than 1 profile (comma separated).  The right fix would be something like 
`Arrays.asList(profile).contains(defaultProfile))`

ryanjbaxter (changes requested):

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/GoogleParameterManagerEnvironmentRepository.java
yashdotdev13: Fixed, thanks. I now split the comma-separated profiles first and check whether the default profile is already present before adding it. The tests pass successfully.

docs/modules/ROOT/pages/server/environment-repository/gcp-parameter-manager-backend.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+Spring Cloud Config Server can use link:https://cloud.google.com/parameter-manager[Google Cloud Parameter Manager] as an `EnvironmentRepository`.
+Parameters are exposed as `PropertySource` entries whose values are obtained from the latest parameter versions.
+
+== Enabling the backend
+
+The repository is registered when all of the following hold:
+
+* The Spring profile `parameter-manager` is active.
+* `com.google.cloud.parametermanager.v1.ParameterManagerClient` is on the classpath.
```
ryanjbaxter: Instead of the class, please specify what the dependency is that they need to add to the classpath

yashdotdev13: Updated to reference the google-cloud-parameter-manager dependency instead of the Java class. Thanks!

docs/modules/ROOT/pages/server/environment-repository/gcp-parameter-manager-backend.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+Spring Cloud Config Server can use link:https://cloud.google.com/parameter-manager[Google Cloud Parameter Manager] as an `EnvironmentRepository`.
+Parameters are exposed as `PropertySource` entries whose values are obtained from the latest parameter versions.
+
+== Enabling the backend
+
+The repository is registered when all of the following hold:
+
+* The Spring profile `parameter-manager` is active.
+* The `google-cloud-parameter-manager` dependency is on the classpath.
```
ryanjbaxter: I would write this as `com.google.cloud:google-cloud-parameter-manager`

PR #3279, opened 2026-08-29
Add HTTP property path notifier

### LINKED ISSUE #2152: Support sending events to clients with HTTP instead of AMPQ / kafka with Spring Cloud Config Monitor
I want to send events from spring-cloud-config with `spring-cloud-config-monitor`, through HTTP, instead of AMPQ / Kafka.

I'd like the server to send HTTP requests to all of the clients instead of publishing a message, which requires fit infrastructure.

yashdotdev13:
Closes #2152

## Summary

Adds support for notifying applications through HTTP when a configuration change is detected.

The existing `PropertyPathEndpoint` now supports multiple `PropertyPathNotifier` implementations, allowing the existing Spring Cloud Bus notifier and the new HTTP notifier to coexist.

### Changes

- Added `PropertyPathNotifier` abstraction.
- Added `BusPropertyPathNotifier` for Spring Cloud Bus notifications.
- Added `HttpPropertyPathNotifier` for HTTP-based notifications.
- Added conditional auto-configuration for the HTTP notifier via:
  `spring.cloud.config.server.monitor.http.enabled=true`.
- Updated `PropertyPathEndpoint` to notify all configured notifiers.
- Preserved the existing single-notifier constructors for compatibility.
- Added tests for the Bus and HTTP notifier implementations.
- Added auto-configuration coverage for the HTTP notifier.

## Testing

- Ran the complete `spring-cloud-config-monitor` test suite.
- 117 tests passed with 0 failures and 0 errors.
- Checkstyle passed with 0 violations.

### DIFF
--- a/docs/modules/ROOT/pages/server/push-notifications-and-bus.adoc
+++ b/docs/modules/ROOT/pages/server/push-notifications-and-bus.adoc
@@ -29,6 +29,7 @@ When Spring Cloud Bus is not available, the HTTP notifier is used automatically.
 When Spring Cloud Bus is available and enabled, the Bus notifier takes precedence.
 
 The HTTP notifier discovers application instances through a `DiscoveryClient` and sends a `POST` request to the refresh endpoint of each affected instance.
+A `DiscoveryClient` implementation must be available on the classpath for HTTP notifications to be enabled.
 
 By default, the refresh endpoint is:
 
--- a/docs/modules/ROOT/pages/server/push-notifications-and-bus.adoc
+++ b/docs/modules/ROOT/pages/server/push-notifications-and-bus.adoc
@@ -20,6 +20,44 @@ NOTE: The `RefreshRemoteApplicationEvent` is transmitted only if the `spring-clo
 
 NOTE: The default configuration also detects filesystem changes in local git repositories. In that case, the webhook is not used. However, as soon as you edit a config file, a refresh is broadcast.
 
+[[http-property-path-notifier]]
+== HTTP Property Path Notifications
+
+The Config Server can notify applications directly over HTTP instead of using Spring Cloud Bus.
+When Spring Cloud Bus is not available, the HTTP notifier is used automatically.
+
+When Spring Cloud Bus is available and enabled, the Bus notifier takes precedence.
+
+The HTTP notifier discovers application instances through a `DiscoveryClient` and sends a `POST` request to the refresh endpoint of each affected instance.
+
+By default, the refresh endpoint is:
+
+[source,text]
+----
+/actuator/refresh
+----
+
+The refresh endpoint can be customized for individual services with the following property:
+
+[source,yaml]
+----
+spring:
+  cloud:
+    config:
+      server:
+        monitor:
+          http:
+            endpoints:
+              orders: /management/refresh
+              inventory: /actuator/custom-refresh
+----
+
+You can also provide the refresh endpoint as `refresh-endpoint` metadata on a `ServiceInstance`. The service-specific configuration takes precedence over instance metadata.
+
+When neither a configured endpoint nor instance metadata is available, `/actuator/refresh` is used.
+
+NOTE: The target applications must expose the configured refresh endpoint for HTTP notifications to succeed.
+
 [[webhook-secret-validation]]
 == Webhook Secret Validation
 
--- a/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
+++ b/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
@@ -22,7 +22,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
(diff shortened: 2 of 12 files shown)


spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/HttpPropertyPathNotifier.java
```diff
@@ -0,0 +1,50 @@
...
+	public HttpPropertyPathNotifier(DiscoveryClient discoveryClient, RestClient restClient) {
+		this.discoveryClient = discoveryClient;
+		this.restClient = restClient;
+	}
+
+	@Override
+	public void notifyApplications(Set<String> services) {
+		for (String service : services) {
+			for (ServiceInstance instance : this.discoveryClient.getInstances(service)) {
+				this.restClient.post().uri(instance.getUri() + "/actuator/refresh").retrieve().toBodilessEntity();
```
ryanjbaxter: Actuator might reside at a different endpoint, it could be customized.  We should make it configurable.

We could have a properties map which maps service to endpoint, or we could have it placed in the metadata in the service instance.  Maybe supporting both options would be best.

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
```diff
@@ -72,36 +75,38 @@ protected static class BusPropertyPathConfiguration {
...
 
-		// TODO: With the current implementation bus can't be disabled
 		@Bean
-		@ConditionalOnMissingBean(BusProperties.class)
-		public PropertyPathEndpoint noBusBeanPropertyPathEndpoint(
-				@Value("${spring.cloud.bus.id:application}") String id,
-				MonitorConfigurationProperties monitorProperties) {
-			return new PropertyPathEndpoint(new CompositePropertyPathNotificationExtractor(this.extractors), id,
-					monitorProperties.getMaxDashes());
+		@ConditionalOnProperty(prefix = "spring.cloud.config.server.monitor.http", name = "enabled",
```
ryanjbaxter: This should just be automatically enabled if bus is not on the classpath.  And if bus is on the classpath and it is set to enabled it should be prioritized over bus

ryanjbaxter (changes requested):
We will need documentation as well

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/HttpPropertyPathNotifier.java
yashdotdev13: Addressed. The HTTP notifier now supports configurable endpoints through monitor properties, with service instance metadata as an additional option. It falls back to /actuator/refresh when neither is configured.

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
yashdotdev13: Addressed. HTTP notifications are now enabled automatically when Bus is unavailable, while Bus is prioritized when it is available and enabled. Setting spring.cloud.bus.enabled=false enables the HTTP notifier.

docs/modules/ROOT/pages/server/push-notifications-and-bus.adoc
```diff
@@ -20,6 +20,44 @@ NOTE: The `RefreshRemoteApplicationEvent` is transmitted only if the `spring-clo
...
 
+[[http-property-path-notifier]]
+== HTTP Property Path Notifications
+
+The Config Server can notify applications directly over HTTP instead of using Spring Cloud Bus.
+When Spring Cloud Bus is not available, the HTTP notifier is used automatically.
+
+When Spring Cloud Bus is available and enabled, the Bus notifier takes precedence.
+
+The HTTP notifier discovers application instances through a `DiscoveryClient` and sends a `POST` request to the refresh endpoint of each affected instance.
```
ryanjbaxter, line 31: It would be good to call out that a `DiscoveryClient` implementation must be on the classpath

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
```diff
@@ -65,43 +67,54 @@ private static FilterRegistrationBean<WebhookValidatorFilter> createRegistration
...
-	protected static class NoBusPropertyPathConfiguration {
+	protected static class PropertyPathEndpointConfiguration {
+
+		@Autowired(required = false)
+		private List<PropertyPathNotificationExtractor> extractors;
 
 		@Bean
-		public PropertyPathEndpoint noBusPropertyPathEndpoint(@Value("${spring.cloud.bus.id:application}") String id,
-				@Autowired(required = false) List<PropertyPathNotificationExtractor> extractors,
+		public PropertyPathEndpoint propertyPathEndpoint(List<PropertyPathNotifier> notifiers,
```
ryanjbaxter, line 117: I think we can get into a situation where there is no PropertyPathNotifier on the classpath when bus is disabled and no discovery client is on the classapth.  
We probably want to verify at least one of those is present and fail otherwise

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/HttpPropertyPathNotifier.java
```diff
@@ -0,0 +1,73 @@
...
+		this.discoveryClient = discoveryClient;
+		this.restClient = restClient;
+		this.monitorProperties = monitorProperties;
+	}
+
+	@Override
+	public void notifyApplications(Set<String> services) {
+		for (String service : services) {
+			for (ServiceInstance instance : this.discoveryClient.getInstances(service)) {
+				this.restClient.post()
```
ryanjbaxter: If there is an exception making one of these requests the rest won't happen, we should catch and log the exception but continue with the rest of the instances

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/HttpPropertyPathNotifier.java
```diff
@@ -0,0 +1,73 @@
...
+					.uri(instance.getUri() + getEndpoint(service, instance))
```
ryanjbaxter: We should use `UriComponentsBuilder`

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
ryanjbaxter: We should add a test for this as well

ryanjbaxter (changes requested):

ryanjbaxter:
Also I think the bus dependncy should be made optional
https://github.com/spring-cloud/spring-cloud-config/blob/7da841ad8a1c066c8c1e5a590f94421e68247cf3/spring-cloud-config-monitor/pom.xml#L35-L38

docs/modules/ROOT/pages/server/push-notifications-and-bus.adoc
yashdotdev13: Addressed. Updated the documentation to clarify that a DiscoveryClient implementation must be available for HTTP notifications to be enabled.

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/HttpPropertyPathNotifier.java
yashdotdev13: Addressed. Replaced string-based URI concatenation with UriComponentsBuilder when constructing the target refresh URI.

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/HttpPropertyPathNotifier.java
yashdotdev13: Addressed. Exceptions from individual HTTP notifications are now caught and logged, allowing notification processing to continue for the remaining instances.

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/EnvironmentMonitorAutoConfiguration.java
yashdotdev13: Addressed. Added coverage for the scenario where no PropertyPathNotifier is available.

PR #3282, opened 2026-09-02
Include the default application in the Redis backend

### LINKED ISSUE #3134: Redis Environment repository not respecting/returning default "application" values
**Describe the bug**
Git based config server returns properties for the app, it's profiles, as well as defaults contained in "application.properties"
Redis based config server returns ONLY the app files, and not the values stored in the key "application"

**Sample**
Stand up a config server backed by redis 
curl localhost:8080/app/profile1,profile2 | jq

review the propertySources returned, it will only be searching for "app","app-profile1","app-profile2", and NOT "application","application-profile1", and "application-profile2"

HDPark95:
`RedisEnvironmentRepository` only reads the hashes named after the requested application (`app`, `app-<profile>`), so properties under the shared `application` key are never returned. The other backends include them: jdbc, mongodb, credhub, native and aws-s3.

This adds the default application keys, ordered the way jdbc and credhub do, so the requested application still wins over the shared defaults: `app-p2`, `application-p2`, `app-p1`, `application-p1`, `app`, `application`. When the requested application is `application` the key list is unchanged.

New test in `RedisEnvironmentRepositoryIntegrationTests`: on 5.0.x it fails with `["redis:myapp-prod", "redis:myapp"]` and passes with the change.

Fixes gh-3134

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/RedisEnvironmentRepository.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/RedisEnvironmentRepository.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.config.server.environment;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 
@@ -33,6 +34,8 @@ import org.springframework.util.StringUtils;
  * @author KNV Srinivas
  */
 public class RedisEnvironmentRepository implements EnvironmentRepository, Ordered {
+
+	private static final String DEFAULT_APPLICATION = "application";
 
 	private final StringRedisTemplate redis;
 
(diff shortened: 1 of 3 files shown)


spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/RedisEnvironmentRepository.java
```diff
@@ -59,10 +62,13 @@ public Environment findOne(String application, String profile, String label) {
 	}
 
 	private List<String> addKeys(String application, List<String> profiles) {
-		List<String> keys = new ArrayList<>();
-		keys.add(application);
+		List<String> applications = new ArrayList<>(
```
ryanjbaxter: I am pretty sure application can be `null` and can be comma delimated, both of which are not handled here 
https://github.com/spring-cloud/spring-cloud-config/blob/99f4902c22df43dff72a186db05f92c08020a5f8/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/CredhubEnvironmentRepository.java#L97-L112

ryanjbaxter (changes requested):
Can you submit this against the 5.0.x branch?

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/RedisEnvironmentRepository.java
HDPark95: Handled in 1411856, following `CredhubEnvironmentRepository#normalize`: `application` goes through `commaDelimitedListToStringArray`, trimmed, filtered on `hasText` and de-duplicated, with `application` prepended. So `null` resolves to the shared hash only, and `app1,app2` reads every listed hash with later names taking precedence, the same order the Credhub backend produces. Two integration tests cover both cases (`null` -> `[application-prod, application]`; `"app1, app2,app1"` -> `[app2-prod, app1-prod, application-prod, app2, app1, application]`). Reverting the source change fails both.

HDPark95:
@ryanjbaxter the PR is targeting `5.0.x` already (base branch `5.0.x`, one commit on top). Let me know if you'd rather have it against a different branch. The null / comma-delimited handling is in 1411856, see the inline reply.

PR #3289, opened 2026-09-13
Restore empty YAML arrays in JSON and YAML config responses

### LINKED ISSUE #2746: Spring Cloud Config returns an empty string instead of an empty array
**Describe the bug**
After upgrading to the latest version of Spring Cloud Config, I encountered an issue where the cloud-config service returns an empty string ("") instead of an empty array ([]).

`<spring-cloud.version>2024.0.0</spring-cloud.version>`

**Sample**
Example:

```
config.yaml

network:
  urls: []
``` 

I expect to receive the following object:

```
{
  "network":{
    "urls":[]
  }
}
``` 

However, I receive this instead:

```
{
  "network":{
    "urls":""
  }
}
``` 

I noticed a similar issue in the past, which might be related to this problem as well.
https://github.com/spring-cloud/spring-cloud-config/issues/1572

Could you please investigate the cause of this issue as soon as possible?
Thanks in advance!

arimu1:
## Summary

- When config is loaded from YAML, Spring's flattened property representation stores empty collections as an empty string (`network.urls=`).
- Rebuilding nested JSON/YAML in `EnvironmentController` treated those values as literal empty strings, so `network.urls: []` was served as `"urls": ""`.
- After merging property sources, restore standalone empty-string entries (with no indexed `[n]` keys for the same prefix) back to empty lists before building nested JSON/YAML.

Fixes gh-2746

## Test plan

- [x] `./mvnw -pl spring-cloud-config-server test -Dtest=EnvironmentControllerTests,NativeEnvironmentRepositoryTests`
- [x] Added fixture `classpath:/test/empty-array/application.yml` with `network.urls: []`
- [x] Unit tests: `emptyArrayInJson`, `emptyArrayInYaml`
- [x] Integration-style test: `emptyYamlArrayInJsonOutput`

## Notes

Properties format cannot distinguish an intentional empty string scalar from an empty collection (both flatten to `key=`). This change follows the same convention Spring uses when flattening YAML (empty collection → empty string) and reverses it for structured JSON/YAML endpoints (and the shared flatten path used by `.properties`).

### DIFF
--- a/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
+++ b/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/environment/EnvironmentController.java
@@ -20,10 +20,12 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 
 import jakarta.servlet.http.HttpServletResponse;
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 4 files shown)


ryanjbaxter:
IDK if I like the tradeoff of legitimate empty srtrings being converted to empty arrays.  This also seems to impact properties files not just JSON/YAML

@spencergibb thoughts?

spencergibb:
If we can guarantee that legit empty strings don't get translated to empty arrays, I'm ok with that.

ryanjbaxter:
Yeah I am not sure we can guarantee that, at least not with the way it is currently implemented

PR #3272, opened 2026-08-13
Add Azure DevOps workload identity support

abhiramaab:
Closes #3270

This adds workload identity authentication for Git repositories hosted on Azure DevOps (dev.azure.com), using the proposed client-id and managed-identity-enabled configuration properties.

Implementation includes:
- Azure Identity dependency and conditional configuration
- Azure managed identity token acquisition for Azure DevOps
- Bearer token injection into JGit HTTP transport
- Git property binding for client-id and managed-identity-enabled
- Transport callback selection with existing SSH fallback behavior
- Unit and configuration tests

Validation:
- Full spring-cloud-config-server test suite: 952 tests, 0 failures, 0 errors, 9 skipped

PR #3294, opened 2026-09-17
Fix SSH configuration for nested repositories

### LINKED ISSUE #2105: Config Server use the wrong SSH URI key when the additional repos are of the same hostname
**Describe the bug**

If user have several repos, and each repo has its own ssh credential, Config Server will fail. Please check the below test result.

**Sample**

Test materials:
Private repo A in my account
Private repo B in my account
Private repo C in other's account
Account key A in my account
Account key B in my account
Private repo C in other's account
Deploy key A in repo A
Deploy key B in repo B
Case A: repo A with account key A as main repo, works Case B: repo B with account key B as main repo, works Case C: repo C with account key C as main repo, works Case D: repo A with deploy key A as main repo, works Case E: repo B with deploy key B as main repo, works Case F: repo A with account key A as main repo, repo B with account key B as additional repo, works Case G: repo A with deploy key A as main repo, repo B with deploy key B as additional repo, NO works Case H: repo A with account key A as main repo, repo C with account key C as additional repo, NO works

It seems that Config Server can't work with multiple repos. The reason case F works is because the two account keys belong to me and either of them can be access to the two repos.

I think the problem is caused by in `extractNestedProperties`
[spring-cloud-config/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/SshUriPropertyProcessor.java](https://github.com/spring-cloud/spring-cloud-config/blob/d8f6341f336e8da78eaab424ca3a3e82431f83db/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/SshUriPropertyProcessor.java#L59)

```
 private Map<String, JGitEnvironmentProperties> extractNestedProperties( 
```

It uses hostname as sshUriPropertyMap's key so in this case, two different github repos will be overridden to use one ssh configuration.

yashdotdev13:
## Overview

Fixes #2105

When multiple Git repositories use the same SSH hostname, nested repositories could inherit the parent repository's `TransportConfigCallback`. This could cause the nested repository to use the parent's SSH configuration.

This change creates a transport configuration callback for each nested repository using its own `JGitEnvironmentProperties`.

## Changes

- Build a `TransportConfigCallback` for each nested repository.
- Generalize SSH transport configuration to work with `JGitEnvironmentProperties`.
- Update `SshUriPropertyProcessor` to support both parent and nested repository properties.
- Preserve Git credentials provider configuration for nested repositories.
- Add regression coverage verifying that parent and nested repositories receive their respective transport callbacks.

## Testing

- `SshUriPropertyProcessorTest` — 6/6 passed
- `MultipleJGitEnvironmentRepositoryFactoryTests` — 2/2 passed
- Nested SSH integration tests — passed for both Static and List configurations.
- Checkstyle — 0 violations


ryanjbaxter:
Thanks!  This is somewhat related to another PR that is currently in progress, #3272.  I would like to get #3272 merged first.  That PR is close to being ready, it just needs docs.

yashdotdev13:
Thanks for the clarification! That makes sense. I'll wait for #3272 to be merged first and then update this PR if any changes are needed.

ryanjbaxter:
@yashdotdev13 #3272 has been merged

spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/ssh/PropertiesBasedSshTransportConfigCallback.java
```diff
@@ -50,7 +52,10 @@ public PropertiesBasedSshTransportConfigCallback(MultipleJGitEnvironmentProperti
 	}
 
 	public MultipleJGitEnvironmentProperties getSshUriProperties() {
-		return this.sshUriProperties;
+		if (this.sshUriProperties instanceof MultipleJGitEnvironmentProperties multipleProperties) {
+			return multipleProperties;
+		}
+		throw new IllegalStateException("SSH URI properties are not multiple repository properties");
```
ryanjbaxter: This now throws an exception if its not `MultipleJGitEnvironmentProperties`.  It seems possible that it could be `PatternMatchingJGitEnvironmentProperties` now, should it still throw an exception?  I dont see this tested anywhere either.

ryanjbaxter (changes requested):

Issue #1885, opened 2021-05-23
Refresh events are dispatched even when unrelated git file is changed


ChiragMoradiya:
**Describe the bug**
- Given config-server is configured with the `spring.cloud.config.server.git.searchPaths` to `spring-config`
- And cloud-bus is also configured on the config-server
- And gitlab webhook is setup to notify the git repository changes on `/monitor` path of the config-server
- When I change an unrelated file in the git repository (e.g. `/docker-compose/docker-compose.yml`)
- Then config-server dispatches a `RefreshRemoteApplicationEvent` for the destinations: `docker-compose:*:**` and `docker-compose:*:**` [ISSUE]

**Expected Result**
As the git repository file updated was out of the configured `searchPaths`, No RefreshRemoteApplicationEvent` should be dispatched.

**Suggested Fix**
[`PropertyPathEndpoint.notifyByPath`](https://github.com/spring-cloud/spring-cloud-config/blob/be708efc18b2d503f9f27b3d96cbc668f938aaaa/spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/PropertyPathEndpoint.java#L75) should be updated.  It should filter out the paths retrieved from the `PropertyPathNotifiationExtractor` which aren't matching the configured `searchPaths`.

spencergibb:
Thanks for the suggested fix, but I doubt we'd be able to match search paths. Maybe there could be a list of regexes to ignore.

PR #3296, opened 2026-09-17
Add ignored paths for config monitor

yashdotdev13:
## Summary

- Add `spring.cloud.config.server.monitor.ignored-paths` for excluding repository paths from refresh notifications.
- Filter ignored paths before resolving service names.
- Add test coverage for ignored and non-ignored paths.
- Document the new monitor property.

This is the `5.0.x` compatible version of the change from #1885.


spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/MonitorConfigurationProperties.java
```diff
@@ -112,6 +121,36 @@ public void setMaxDashes(int maxDashes) {
...
+
+	public Http getHttp() {
+		return http;
+	}
+
+	public void setHttp(Http http) {
+		this.http = http;
+	}
+
+	public static class Http {
```
ryanjbaxter: What is this for?

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/PropertyPathEndpoint.java
```diff
@@ -66,17 +65,28 @@ public class PropertyPathEndpoint implements ApplicationEventPublisherAware {
...
 	public PropertyPathEndpoint(PropertyPathNotificationExtractor extractor, String busId, int maxDashes) {
+		this(extractor, busId, maxDashes, Collections.emptyList());
+	}
+
+	public PropertyPathEndpoint(PropertyPathNotificationExtractor extractor, String busId, int maxDashes,
+			List<String> ignoredPaths) {
 		this.extractor = extractor;
 		this.busId = busId;
 		this.maxDashes = maxDashes;
+		this.ignoredPaths = ignoredPaths.stream().map(Pattern::compile).toList();
```
ryanjbaxter: We should add a null guard here before calling .stream()

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/PropertyPathEndpoint.java
```diff
@@ -93,8 +103,11 @@ public Set<String> notifyByPath(@RequestHeader HttpHeaders headers, @RequestBody
 			Set<String> services = new LinkedHashSet<>();
 
 			for (String path : notification.getPaths()) {
```
ryanjbaxter: Looking at this code, I think we should add a limit to the number of paths we are processing here since that comes externally.  Adding a property to like DEFAULT_MAX_PATHS similar to DEFAULT_MAX_DASHES with a similar warn logic would be a good idea.

ryanjbaxter (changes requested):

spring-cloud-config-monitor/src/main/java/org/springframework/cloud/config/monitor/MonitorConfigurationProperties.java
yashdotdev13: Http is pre-existing configuration and isn't part of this change. My change only adds the ignoredPaths property and its getter/setter to MonitorConfigurationProperties.

ryanjbaxter:
I think you merged main into your branch on accident

yashdotdev13:
Thanks Ryan. I did have some unrelated history on the branch previously. I cleaned up the branch and force-updated the existing PR so it now contains only the #1885 change on top of 5.0.x. The current PR has a single commit (4ae95683) with the six relevant files.

yashdotdev13:
The branch has now been cleaned up and the updated commit is passing both DCO and CI checks. Please take another look when you get a chance.

ryanjbaxter (approved):
Looks good but we will have to wait to merge this until after the 24th when we do our release

yashdotdev13:
Thanks,  sounds good. I’ll wait until after the release.

PR #3298, opened 2026-09-18
Port application prefix validation to configdata

### LINKED ISSUE #1938: Port `application-` prefix failure analyzer to configdata
See  #1588

yashdotdev13:
## Description

Port the existing `application-` prefixed application name validation to the Config Data loading path.

Previously, this validation was handled by the legacy `PropertySourceLocator` path, but Config Data did not perform the same validation.

This change:
- Validates the resolved `spring.application.name` / `spring.cloud.config.name`.
- Throws `InvalidApplicationNameException` when the resolved name starts with `application-`.
- Preserves case-insensitive validation.
- Adds tests covering application name, config name override, and case-insensitive values.

## Tests

- `mvnw -pl spring-cloud-config-client test`
- 190 tests, 0 failures, 7 skipped

Fixes #1938


spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
```diff
@@ -192,7 +193,9 @@ protected PropertyHolder loadProperties(ConfigDataLocationResolverContext contex
 			}
 			configClientProperties.setUri(uri);
 		}
-
+		if (StringUtils.startsWithIgnoreCase(configClientProperties.getName(), "application-")) {
+			throw new InvalidApplicationNameException(configClientProperties.getName());
```
ryanjbaxter: The legacy path only throws InvalidApplicationNameException when configClientProperties.isFailFast() is true otherwise it just logs a warning and skips loading remote config, letting the app continue starting. The new code throws unconditionally.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
```diff
@@ -192,7 +193,9 @@ protected PropertyHolder loadProperties(ConfigDataLocationResolverContext contex
 			}
 			configClientProperties.setUri(uri);
 		}
-
+		if (StringUtils.startsWithIgnoreCase(configClientProperties.getName(), "application-")) {
```
ryanjbaxter: This logic in now duplicated and should use a shared helper

ryanjbaxter (changes requested):
Add a test that covers the failFast=false case as well

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
```diff
@@ -235,6 +232,17 @@ public List<ConfigServerConfigDataResource> resolveProfileSpecific(
 		PropertyHolder propertyHolder = loadProperties(resolverContext, uris);
 		ConfigClientProperties properties = propertyHolder.properties;
 
+		if (ConfigClientProperties.isInvalidApplicationName(properties.getName())) {
```
ryanjbaxter: I guess this whole code block can be lifted out into a helper method.  It should not be ConfigClientProperties.  It seems like a new utility class might be needed, something called `ApplicationNameValidator`

ryanjbaxter (changes requested):

PR #3301, opened 2026-09-19
Expose profile metadata in Config Server property sources

### LINKED ISSUE #1874: Add profile to config server response data
Currently, the response from config server contains the profiles used to make the call.
```
{
  "name": "foo",
  "profiles": [
    "development"
  ]
  ...
```

As part of ongoing ConfigData changes, it's useful to know the **specific** profile associated with the property source if applicable.  See #1795. This matches profile against the name, but if available in the response, no string matching would be needed.

See https://github.com/spring-projects/spring-boot/issues/25940#issuecomment-820647528

yashdotdev13:
## Description

Resolves #1874

This change exposes the profile associated with a `StandardConfigDataResource` in the Config Server `PropertySource` response.

Previously, `ConfigServerConfigDataLoader` determined whether a property source was profile-specific primarily by matching the profile against the property source name. This can be unreliable when the profile is not part of the property source name.

### Changes

- Add `profile` metadata to the Config Server `PropertySource` DTO.
- Extract the profile from `StandardConfigDataResource` in `NativeEnvironmentRepository`.
- Propagate the profile metadata to the Config Server response.
- Update `ConfigServerConfigDataLoader` to use the explicit profile metadata when available.
- Retain the existing property-source name matching as a fallback for compatibility with Config Server versions that do not provide profile metadata.
- Add a regression test covering a profile-specific property source whose name does not contain the profile.

### Tests

- `ConfigDataProfileOnlyTests` — 2 tests passed
- `ConfigServerConfigDataLoaderTests` — 35 tests passed, 3 skipped


ryanjbaxter:
Given the critical code path this change relies on I am inclined to wait until a major release to make this change.

yashdotdev13:
Thanks Ryan, understood.

Given that this changes the profile handling in a critical Config Data path, I'm fine with waiting for a major release. I'll leave the PR as-is for now and can update it if there are any changes needed for the major release.

PR #3303, opened 2026-09-22
Migrate Config Client to Spring Framework native retry

### LINKED ISSUE #3261: Migrate from Spring Retry to Spring Framework's native retry features
In applications using Spring Boot 4/Spring Framework 7, other libraries in the Spring portfolio switched to using the built-in retry features that superseded the old Spring Retry project (Batch, Integration, AMQP, Kafka, Pulsar...). When using Spring Cloud Config Client, it's necessary to add the extra Spring Retry dependency (which is [maintenance only](https://github.com/spring-projects/spring-retry)).

It would be great if Spring Cloud Config could migrate from the deprecated Spring Retry library to Spring Framework's built-in `RetryTemplate` infrastructure.

**Additional context**
https://spring.io/blog/2025/09/09/core-spring-resilience-features

Emin0v:
### Motivation

Spring Framework now provides native retry support, while Config Client still depends on Spring Retry for retry behavior.

This change migrates Config Client to Spring Framework's retry APIs and removes the Spring Retry dependency.

### Changes

* Migrate Config Data loading and discovery retry to Spring Framework retry.
* Migrate the legacy bootstrap retry path from Spring Retry annotations and `RetryOperationsInterceptor` to Spring AOP backed by native retry, keeping the `configServerRetryInterceptor` bean name as the override point.
* Preserve `max-attempts` total-execution semantics.
* Preserve existing top-level failure behavior.
* Add `spring.cloud.config.retry.enabled` as an explicit opt-out.
* Deprecate `spring.cloud.config.retry.use-random-policy` while preserving its behavior.
* Remove Spring Retry from Config Client dependencies.

### Compatibility

* `max-attempts` continues to mean the total number of executions, including the initial invocation.
* Existing `spring.cloud.config.retry.*` property names are unchanged.
* `spring.cloud.config.retry.enabled=false` disables retry, so the config server is contacted once.
* `use-random-policy` remains functional but is deprecated.

### Tests

`./mvnw -pl spring-cloud-config-client verify`

211 tests, 0 failures, 0 errors, 7 skipped

`./mvnw -pl spring-cloud-config-sample -am -Dtest=ApplicationFailFastTests,ConfigDataRetryIntegrationTests test`

3 tests, 0 failures

The Config Client dependency tree no longer contains Spring Retry.

Fixes gh-3261


Emin0v:
Marking this ready for review. I’m aware of #3304, which addresses the same issue with a different implementation approach.

There are a few compatibility differences that may be worth comparing before either PR moves forward:

- preserving the previous retry exception classification so Error is not retried;
- preserving the existing upward-randomized useRandomPolicy behavior rather than approximating it with symmetric jitter;
- removing the Spring Retry dependency from the Config Client module and updating the related documentation.

I’m happy to coordinate or consolidate based on maintainer preference.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServiceBootstrapConfiguration.java
```diff
@@ -62,25 +66,54 @@ public ConfigServicePropertySourceLocator configServicePropertySource(ConfigClie
 	}
 
 	@ConditionalOnProperty(ConfigClientProperties.PREFIX + ".fail-fast")
-	@ConditionalOnClass({ Retryable.class, Aspect.class, AopAutoConfiguration.class })
+	@ConditionalOnProperty(name = RetryProperties.PREFIX + ".enabled", matchIfMissing = true)
 	@Configuration(proxyBeanMethods = false)
-	@EnableRetry(proxyTargetClass = true)
 	@Import(AopAutoConfiguration.class)
```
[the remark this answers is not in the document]
Emin0v: Fixed in d361d1d. The affected legacy retry beans now explicitly preserve target-class proxying, with regression coverage for spring.aop.proxy-target-class=false.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/RetryProperties.java
```diff
@@ -30,6 +31,12 @@ public class RetryProperties {
 	 */
 	public static final String PREFIX = "spring.cloud.config.retry";
 
+	/**
+	 * Whether to retry failed requests to the Config Server. Only has an effect when
+	 * spring.cloud.config.fail-fast is enabled.
+	 */
+	boolean enabled = true;
```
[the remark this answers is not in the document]
Emin0v: Fixed in d361d1d. The generated configuration-properties reference now includes spring.cloud.config.retry.enabled with its default value and description.

docs/modules/ROOT/pages/client.adoc
```diff
@@ -90,17 +90,23 @@ NOTE: To get similar functionality using `spring.config.import`, simply omit the
 
 If you expect that the config server may occasionally be unavailable when your application starts, you can make it keep trying after a failure.
 First, you need to set `spring.cloud.config.fail-fast=true`.
-Then you need to add `spring-retry` and `spring-boot-starter-aop` to your classpath.
+Retry is built on the retry support of Spring Framework, so no extra dependency is required.
```
ryanjbaxter: probably worth linking out to the docs here

docs/modules/ROOT/pages/client.adoc
```diff
@@ -90,17 +90,23 @@ NOTE: To get similar functionality using `spring.config.import`, simply omit the
...
 The default behavior is to retry six times with an initial backoff interval of 1000ms and an exponential multiplier of 1.1 for subsequent backoffs.
+`max-attempts` counts the initial invocation, so the default of `6` means one invocation followed by at most five retries.
 You can configure these properties (and others) by setting the `spring.cloud.config.retry.*` configuration properties.
-To use a random exponential backoff policy set `spring.cloud.config.retry.useRandomPolicy` to `true`.
+To turn retry off while keeping `fail-fast` enabled, set `spring.cloud.config.retry.enabled=false`; the client then contacts the config server exactly once.
```
ryanjbaxter: To maintain the opt-in behavior for retry functionality we had before I think it better that the default for `enabled` be `false` and have users continue to be implicit about wanting this functionality.  This keeps retry out of the picture for users who never had it before rather than being surprised that it was enabled by default.

spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServerConfigDataLocationResolver.java
```diff
@@ -177,6 +177,9 @@ protected PropertyHolder loadProperties(ConfigDataLocationResolverContext contex
 					map.from(() -> properties.getProperty("fail-fast"))
 						.as(Boolean::valueOf)
 						.to(configClientProperties::setFailFast);
+					map.from(() -> properties.getProperty("enabled"))
```
ryanjbaxter: Why are we adding this here?  This was not something we supported in the prior implementation

ryanjbaxter (changes requested):

